PR #33, opened 2020-09-29
chore: update sourcepointId to name

ioanna0:
## What does this change?

Sourcepoint Id is moved to `consent-management-platform` and name is used instead to use inside the `getConsentFor` 
https://github.com/guardian/consent-management-platform/pull/276

## Why?
Will work with new cmp release

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -5,7 +5,6 @@ export type TagAtrribute = {
 
 export type GetThirdPartyTag = (arg0: {
 	shouldRun: boolean;
-	facebookAccountId?: string;
 }) => ThirdPartyTag;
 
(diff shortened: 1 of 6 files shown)


sndrs (approved):

src/types.ts
```diff
@@ -16,7 +13,7 @@ export type ThirdPartyTag = {
 	loaded?: boolean;
 	onLoad?: () => void;
 	shouldRun: boolean;
-	sourcepointId?: string;
+	name?: string;
```
mxdvl, line 16: Would it be worth importing the `enum` type from `@guardian/consent-management-platform`?

PR #38, opened 2020-10-05
Bump deps

sndrs:


mxdvl:
How comes you’re updating the `yarn.lock` but not `package.json`?

sndrs:
> How comes you’re updating the yarn.lock but not package.json?

the package only define the acceptable range, whereas the lockfile defines the actual version to install. these are just updates within the range specified in package – effectively just what dependabot does in one go

PR #40, opened 2020-10-07
feat(3PT): Add inizio

mxdvl:
## What does this change?

Add Inizio third-party tag & adds types on the window.

- Use [DefinitelyTyped / `@types/doubleclick-gpt`][googletag] for `window.googletag`
- Use `undefined` for `sourcepointId` to highlight an error if it is set instead of `name`

[googletag]: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/doubleclick-gpt/index.d.ts

## Why?

Move it out of `frontend`.


mxdvl:
@sndrs what’s the take on merge commits? Are they `chore`?

src/types.ts
```diff
@@ -13,7 +13,16 @@ export type ThirdPartyTag = {
 	loaded?: boolean;
 	onLoad?: () => void;
 	shouldRun: boolean;
+	sourcepointId?: undefined;
```
ioanna0, line 16: this field should no longer be needed? as all occurrences of sourcepointIds inside `frontend` have been moved into  the `consent-management-platform`

src/types.ts
```diff
@@ -13,7 +13,16 @@ export type ThirdPartyTag = {
...
 	name?: string;
 	url?: string;
 	useImage?: boolean;
 };
+
+declare global {
+	interface Window {
+		_brandmetrics?: { cmd: string; val: Record<string, unknown> }[];
+		// eslint-disable-next-line no-undef
+		googletag: googletag.Googletag;
```
ioanna0, line 26: nice that you found  the exact Googletag type 👍 .

ioanna0:
@sndrs @mxdvl 
what happens when 1 PR has multiple different types of commits for semantic releases? I can see 5 different types of commits in this PR, curious how will that work?

sndrs:
> what’s the take on merge commits such as 72c9993—are they chore?

yeah chore should be fine i'd think. i've been rebasing locally to avoid this and so far i've not encountered any grim conflict chains

>what happens when 1 PR has multiple different types of commits for semantic releases? I can see 5 different types of commits in this PR, curious how will that work?

that's fine - semantic release will examine each commit and decide what it needs to do. you can [see it actually examining commits in the build log](https://github.com/guardian/commercial-core/runs/1181969855?check_suite_focus=true#step:6:39)

src/third-party-tags/inizio.ts
```diff
@@ -0,0 +1,34 @@
...
+				window.googletag.cmd.push(() => {
+					window.googletag.pubads().setTargeting('inizio', 't');
+				});
+			}
+			console.log(`surveyAvailable: ${survey.measurementId}`);
+		}
+	};
+	// eslint-disable-next-line no-underscore-dangle
+	window._brandmetrics = window._brandmetrics || [];
+	// eslint-disable-next-line no-underscore-dangle
```
sndrs, line 19: wonder if we should just disable this rule for this codebase? feels like it's going to crop up a lot and it's not really catching anything useful?

src/types.ts
```diff
@@ -13,7 +13,16 @@ export type ThirdPartyTag = {
...
+	sourcepointId?: undefined;
 	name?: string;
 	url?: string;
 	useImage?: boolean;
 };
+
+declare global {
+	interface Window {
+		_brandmetrics?: { cmd: string; val: Record<string, unknown> }[];
+		// eslint-disable-next-line no-undef
```
sndrs, line 25: what is this erroring on? we should be wary of disabling rules

src/third-party-tags/inizio.ts
```diff
@@ -0,0 +1,34 @@
...
+			if (window && window.googletag) {
+				window.googletag.cmd.push(() => {
+					window.googletag.pubads().setTargeting('inizio', 't');
+				});
+			}
+			console.log(`surveyAvailable: ${survey.measurementId}`);
+		}
+	};
+	// eslint-disable-next-line no-underscore-dangle
+	window._brandmetrics = window._brandmetrics || [];
```
sndrs, line 18: if you wanna be hipster 🕺 
```suggestion
	window._brandmetrics ||= [];
```

mxdvl: If I **commit suggestion,** will it make commitizen unhappy?

sndrs: ooh dont know – wanna try it?

src/types.ts
mxdvl: ESLint is unhappy because `googletag` is not defined… If we make it happy by import it from the declaration file type as below, TS is then unhappy because you cannot import from a declaration file.

```ts
import googletag from '@types/doubleclick-gpt';
```

src/third-party-tags/inizio.ts
mxdvl: Defining the type seems to have fixed it anyway. Removed in ed7e573b6b8605ca0743eeaae6492d51f081b5ec

src/types.ts
mxdvl: Following a suggestion from @SiAdcock, declaring it as a `readonly` global in `.eslintrc` did the trick:

```json
{
	"globals": { "googletag": "readonly" },
	# … rest of config
}
```

mxdvl:
I had to manually rebase and `git push --force-with-lease` to allow **Rebase and merge**.

src/third-party-tags/inizio.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+
+describe('index', () => {
+	it('should use the feature switch option', () => {
+		const inizioInstance = inizio({ shouldRun: true });
+		expect(inizioInstance).toStrictEqual({
+			shouldRun: true,
+			url:
+				'//cdn.brandmetrics.com/survey/script/e96d04c832084488a841a06b49b8fb2d.js',
+			name: 'inizio',
+			onLoad: inizioInstance.onLoad,
```
sndrs, line 11: is this testing anything?

mxdvl: Not really…

sndrs: ah i guess it's needed for the `toStrictEqual`?

mxdvl: Actually, it makes more sense to use `toMatchObject` in this context, as in #52.

ioanna0:
@mxdvl think that was left behind

ioanna0 (approved):
👍 looks good. i think we don't need sourcepointId in the ThirdPartyTag but otherwise can 🚢

src/types.ts
mxdvl: Good point!

```suggestion
```

mxdvl:
This needs some tests added, which caused a coverage failure in #134 when we removed Lotame in #138.

PR #42, opened 2020-10-08
Require eslint disable explanation

sndrs:
## What does this change?

- lints for explanations if you `eslint-disable`

## Why?

info for the next viewer


jest.config.js
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable @typescript-eslint/no-var-requires */
+/* eslint-disable @typescript-eslint/no-var-requires -- this really is a node file, it's fine */
```
mxdvl, line 1:
```suggestion
/* eslint-disable @typescript-eslint/no-var-requires -- this really is a node file, it’s fine */
```

mxdvl (commented):
Good idea!

PR #48, opened 2020-10-12
build(package.json): linting now runs pre-commit instead of pre-push

ashishpuliyel:
## What does this change?
The hook for linting now runs pre-commit instead of pre-push

## Why?
Linter runs on staged files, but was setup to run on pre-push which is odd. Now it runs pre-commit

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,8 +23,8 @@
 		"compile:common": "tsc --project ./tsconfig.build.json --outDir ./dist/cjs --module CommonJS",
 		"compile:esm": "tsc --project ./tsconfig.build.json --outDir ./dist/esm",
 		"lint": "eslint .",
-		"prepush:lint": "lint-staged",
-		"prepush:test": "jest --verbose --runInBand --onlyChanged",
+		"precommit:lint": "lint-staged",
+		"precommit:test": "jest --verbose --runInBand --onlyChanged",
 		"test": "jest",
 		"tsc": "tsc --noEmit",
 		"validate": "npm-run-all tsc lint test build"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -23,16 +23,16 @@
 		"compile:common": "tsc --project ./tsconfig.build.json --outDir ./dist/cjs --module CommonJS",
 		"compile:esm": "tsc --project ./tsconfig.build.json --outDir ./dist/esm",
 		"lint": "eslint .",
-		"prepush:lint": "lint-staged",
-		"prepush:test": "jest --verbose --runInBand --onlyChanged",
+		"precommit:lint": "lint-staged",
+		"precommit:test": "jest --verbose --runInBand --onlyChanged",
```
sndrs, line 27: do we want to run tests on pre-commit? maybe these should stay pre-push?

package.json
```diff
@@ -41,7 +41,7 @@
 		]
 	},
 	"lint-staged": {
-		"*": "eslint --fix"
+		"*.js,*.ts": "eslint --fix"
```
sndrs, line 44: can you elucidate the value of this?

the thing that gives me slight pause is that it means we'll have to keep it up to date (e.g. if we add `tsx`, `svelte` files etc)

ashishpuliyel: It currently runs on other files like JSON and it doesn't know what to do with them. Like so:
Screenshot 2020-10-12 at 11 48 24

And presumably there is a longer list of extensions to ignore? So figured this was the way forward

package.json
ashishpuliyel: You think so? I figure if the test fails you've got to fix it, which is another code change. And I'd rather commit code that passes the test. But it does hurt ones ability to check-in partially completed work. I could be  persuaded either way on this.

sndrs: fair enough, let's try it and see!

sndrs (approved):
nice one!

sndrs (approved):

PR #49, opened 2020-10-12
Improve commit lint messaging

sndrs:
## What does this change?

- move the error message to a script file
- updates eslint to treat `*.js` as node files, and so allow `require`

## Why?

- better control of messaging

Screenshot 2020-10-12 at 12 13 10


sndrs:
superseded by #53

PR #53, opened 2020-10-12
ci: Semantic PRs

sndrs:
## What does this change?

- lints PRs for conventional commit compliance, rather than individual commits

## Why?

- we should get the benefit of automatic semantic release without getting too involved with individual developer's ways of working


mxdvl (approved):
I like it :)

PR #56, opened 2020-10-13
feat: Export more lotame data

ioanna0:
needed by frontend


mxdvl:
Do you need `export type` and could you not include it in the above export?

ioanna0:
> Do you need `export type` and could you not include it in the above export?

Yes when exporting a type the `type` declaration is needed which is separate from the exported modules
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html

mxdvl (approved):

PR #57, opened 2020-10-14
fix: move vendor back into src

sndrs:
## What does this change?

- moves `vendor` back under `src`, with a `__` prefix to keep it separate from our own code
- converts vendor files to JS, since TS isn't catching anything in there

## Why?

multiple `rootDirs` in the tsconfig meant everything was published under the directory it lives in (src, vendor etc). this broke importing. 

@ioanna0 and i mulled this over and decided we could either do this or change the main/module paths. that felt wrong, since it's not src at that point, it's our compiled artefact.

also, vendor code _is_ source code of this package, just not ours.

this attempts to mark it as such while keeping a clean published artefact

we should probably add some validation for the built package before publishing it


ioanna0 (approved):
👍

PR #131, opened 2021-01-04
chore: Automatically merge minor dependencies bumps

mxdvl:
## What does this change?

When `patch` or `minor` updates are raised by @dependabot, merge them automatically if all tests pass.

As the PR titles start with `chore:`, these automatic merges should never trigger a package release.

Uses https://github.com/marketplace/actions/dependabot-auto-merge

## Why?

- Less noise for developers, requiring attention only if tests do not pass.
- No more [bundled updates](https://github.com/guardian/commercial-core/pull/130), which make semantic release notes less useful.

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -1,0 +1,14 @@
+name: auto-merge
+on:
+  pull_request:
+    branches-ignore:
+      - 'gh-pages'
+jobs:
+  auto-merge:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ahmadnassri/action-dependabot-auto-merge@v2
+        with:
+          target: minor
+          github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-merge.yml
```diff
@@ -0,0 +1,14 @@
...
+      - 'gh-pages'
+jobs:
+  auto-merge:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ahmadnassri/action-dependabot-auto-merge@v2
+        with:
+          target: minor
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
sndrs, line 14: I'm pretty sure this token does not have enough permissions to merge https://github.com/marketplace/actions/dependabot-auto-merge#token-scope

you'd need to make a personal access token, which would then be like you merging them all manually. do you want that responsibility? :)

mxdvl: Ah, I see, that’s the same issue as `semantic-release`, then… Could we use a @commercial-dev token?

sndrs: I'm pretty sure they're per-user, it's an interesting question though...

mxdvl:
This is not going to be possible just yet…

PR #138, opened 2021-01-13
feat!: Remove Lotame

mxdvl:
⚠️ Do not merge before Jan 14th, 2021 ⚠️

## What does this change?

Removes Lotame as a vendor.

## Why?

Lotame was used as part of Ozone, but this is no longer the case. See [Trello card](https://trello.com/c/64ae583b/1241-remove-lotame-ozone-on-the-14th-jan-2021).


ioanna0 (approved):
Bye Lotame! 🧹

mxdvl:
⚠️ Do not merge before Jan 14th ⚠️

mxdvl:
This should actually have triggered a major release, as it’s a breaking change for anyone depending on Lotame…

sndrs:
i think the build was not successful, so no publish https://github.com/guardian/commercial-core/commit/686ecefc024965c866c79340b54b066e499d06e9

mxdvl:
And only because coverage went down…
Note to self: check the build is successful and package is _actually_ published.

PR #155, opened 2021-02-03
feat:Better api for EventTimer

ashishpuliyel:
## What does this change?
Adds a EventTimer.get() method 

## Why?
Easier to access the singleton.


ioanna0 (approved):
looks good 👍

mxdvl:
Maybe we could also make use of the [`log('commercial', event)`][1] to log commercial-specific info?

[1]: https://github.com/guardian/libs/blob/main/src/logger.README.md

ioanna0:
I didn't trigger a release for some reason 🤔 should we push an empty one in main?

mxdvl:
Coverage decreased, meaning the build was unsuccessful, meaning it did not release.

Same as #138. 😢

PR #156, opened 2021-02-03
fix: correct usage of template literals

ioanna0:


ashishpuliyel (approved):
Duh! What kind of idiot wrote this code?!

PR #157, opened 2021-02-04
feat: Add slot ready event

ioanna0:
## What does this change?
Add slot ready event

## Why?
To capture the time when slot is ready


ashishpuliyel (approved):
Spelling mistakes! Terrible.

PR #171, opened 2021-02-16
feat: Track  events in GA

ioanna0:
## What does this change?
Tracks commercial events in google analytics. So far, only slotReady and slotInitialized will be tracked but is easily configureable in gaConfig object

## Why?
To measure commercial events


src/GoogleAnalytics.ts
```diff
@@ -0,0 +1,23 @@
...
+	timingLabel: string,
+	trackerName: string,
+): void => {
+	const { ga } = window;
+
+	if (typeof ga === 'undefined') {
+		console.error("Can't track GA event - GA library not loaded");
+		return;
+	}
+	const timeSincePageLoad: number = Math.round(window.performance.now());
```
arelra, line 13: Hi just wondering why the `window.performance.now()` value is rounded?
From performance api docs it looks like browsers already round to x ms to mitigate timing attacks

ioanna0: Hi, we round on frontend as well and wanted to keep it for consistency: https://github.com/guardian/frontend/blob/6f5897587dc2f5b03596e0843dcc229763a32ef2/static/src/javascripts/projects/common/modules/analytics/google.js#L117

But since you pointed out we may not need it? Although be reading the [docs](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) it says:
```
To offer protection against timing attacks and fingerprinting, the precision of performance.now() **might** get rounded **depending on browser settings**.
```
I concerned this might not be a constant behaviour?

package.json
```diff
@@ -53,6 +53,7 @@
     "@semantic-release/github": "7.2.0",
     "@types/doubleclick-gpt": "^2019041801.0.4",
     "@types/jest": "^26.0.20",
+		"@types/google.analytics": "^0.0.41",
```
mxdvl, line 56: I think the linting rules should be applied, here for example we have tabs instead of spaces.

```suggestion
    "@types/google.analytics": "^0.0.41",
```

src/EventTimer.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+		);
+		expect(trackEvent).toHaveBeenCalledWith(
+			'gu.commercial.slotReady',
+			'top-above-nav-slotReady',
+			'new',
+			'test',
+		);
+	});
+
+	it('not trigget a GA event if not in GA config', () => {
```
mxdvl, line 64: Small typo:

```suggestion
	it('not trigger a GA event if not in GA config', () => {
```

src/GoogleAnalytics.ts
arelra: Hi the "might" is concerning!

The W3C spec says browsers _should_ add randomness and also caniuse:

https://www.w3.org/TR/2019/REC-hr-time-2-20191121/#sec-domhighrestimestamp
https://caniuse.com/high-resolution-time

But you're right, best to be cautious 👍

mxdvl (commented):
Just a couple of typos, but otherwise 👍

src/EventTimer.ts
```diff
@@ -32,24 +43,26 @@ export class EventTimer {
...
-	 *
+	 * @param {string} trackerName - Name of the GA tracker
 	 * @returns {EventTimer} Instance of EventTimer
 	 */
-	static init(): EventTimer {
-		return (window.guardian.commercialTimer ||= new EventTimer());
+	static init(gaTrackerName: string): EventTimer {
+		return (window.guardian.commercialTimer ||= new EventTimer(
+			gaTrackerName,
+		));
```
mxdvl, line 52: What happens if you try to init a second time with a different `gaTrackerName`? It just uses the previous singleton, right?

mxdvl (approved):
I’m a little fuzzy on why `gaTrackerName` is an argument, as you only ever get one instance of the event timer: the one registered on the `window.guardian`.

Does that not mean it’s a constant, for the purposes of this code?

src/EventTimer.ts
ioanna0: Yeah I kind of broke the singleton in that way since it accepts params. In reality we wouldn't call it twice with different param but it does break the design pattern.
I've added it in as param because we haven't migrated [config lib](https://github.com/guardian/frontend/blob/9a89d67f4c7d867f0dc5dcafdb940fbc7e1dfb3c/static/src/javascripts/lib/config.js#L8) from frontend into commercial core yet, but I guess I _could_ just directly get it from the window without going via the lib.

ioanna0: I've changed it to get the tracker name from the window object directly instead of passing it as a param inside the init

src/EventTimer.spec.ts
```diff
@@ -0,0 +1,81 @@
+import { EventTimer } from './EventTimer';
+import { trackEvent } from './GoogleAnalytics';
+
+jest.mock('./GoogleAnalytics');
+
+describe('EventTimer', () => {
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access -- delete performance to mock it as is readonly
+	delete (window as any).performance;
```
mxdvl, line 8: GitHub Actions is now flagging this on every new PR as a linting error. Is there any way we could fix it or ask to ignore it?

ioanna0: removed here: https://github.com/guardian/commercial-core/pull/176/files

PR #176, opened 2021-02-23
feat: Add commercial start and commercial end events

ioanna0:
## What does this change?
Adds commercial start and commercial end timings.
Also refactored the category/variable/label of ga timings


ashishpuliyel (approved):
Looks fine to me

PR #177, opened 2021-02-26
feat: make trackername optional

ioanna0:
## What does this change?
Tracker name will default to the one set in `ga.create` if not set 

## Why?
To keep monitoring events without tracker name specified in window.config


arelra (commented):
lgtm 👍

arelra (approved):

PR #187, opened 2021-03-04
fix: Improve some edge cases

ioanna0:
## What does this change?
Improves some edge cases.

- Don't push an empty mark if it doesn't exist
- Don't throw an error if ga doesn't exist - it could happen possibly - we just don't measure the event in this case

## Why?
It will remove some errors we have seen recently

### DIFF
--- a/src/GoogleAnalytics.ts
+++ b/src/GoogleAnalytics.ts
@@ -7,7 +7,6 @@ export const trackEvent = (
 	const trackerName: string | undefined =
 		guardian.config?.googleAnalytics?.trackers.editorial;
 	if (typeof ga === 'undefined') {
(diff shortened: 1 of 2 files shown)


src/GoogleAnalytics.ts
```diff
@@ -7,7 +7,6 @@ export const trackEvent = (
 	const trackerName: string | undefined =
 		guardian.config?.googleAnalytics?.trackers.editorial;
 	if (typeof ga === 'undefined') {
-		console.error("Can't track GA event - GA library not loaded");
```
mxdvl, line 10: Maybe a `console.warn`?

mxdvl (approved):
Nice improvement.

Let’s see if #180 is picked up correctly!

PR #193, opened 2021-03-09
fix: call mark if performance API exists

ioanna0:
## What does this change?
Performance API seems to be undefined on mobiles. 
It got picked up by Sentry here:
https://sentry.io/organizations/the-guardian/issues/2246881950/?project=35463&query=is%3Aunresolved

Also refactored mark to void as the return is not used

## Why?
Less errors in sentry

### DIFF
--- a/src/EventTimer.ts
+++ b/src/EventTimer.ts
@@ -106,18 +106,21 @@ export class EventTimer {
 		};
 	}
 
-	mark(name: string): PerformanceEntry {
+	mark(name: string): void {
 		const longName = `gu.commercial.${name}`;
-		window.performance.mark(longName);
-
-		// Most recent mark with this name is the event we just created.
-		const mark = window.performance
-			.getEntriesByName(longName, 'mark')
-			.slice(-1)[0];
-		if (typeof mark !== 'undefined') {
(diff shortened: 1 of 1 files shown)


src/EventTimer.ts
```diff
@@ -106,18 +106,21 @@ export class EventTimer {
...
-			this.events.push(new Event(name, mark));
+		if (
+			typeof window.performance !== 'undefined' &&
+			'mark' in window.performance
+		) {
+			window.performance.mark(longName);
+			// Most recent mark with this name is the event we just created.
+			const mark = window.performance
+				.getEntriesByName(longName, 'mark')
+				.slice(-1)[0];
```
arelra, line 119: I assume this will always be undefined and not throw a `Uncaught TypeError: Cannot read property '0' of undefined`

arelra (approved):

src/EventTimer.ts
ioanna0: I think we are good in that case because if `getEntriesByName` returns an empty array so in the worst case this will return undefined:
`[].slice(-1)[0]` which we later test against

PR #214, opened 2021-03-30
docs: Import Advertising Docs from `frontend`

mxdvl:
## What does this change?

Import docs from `frontend`, and update DFP &rarr; GAM.

- Consolidate/rewrite information from the original documents on Header Bidding into `header-bidding.md`, to minimise confusion.
- Commit the diagram to version control–however, I didn’t think it was worth turning it into an SVG.

## Why?

The knowledge should live here!


arelra:
Theres also [header-bidding](https://github.com/guardian/frontend/blob/main/static/src/javascripts/projects/common/modules/commercial/docs/header-bidding.md) we can bring in too

arelra:
I hope over time we will add and expand to these docs so they become obsolete.

I'm thinking is it worthwhile to say "Imported from ..." to signify their provenance and historical context?

arelra (approved):

arelra (approved):

PR #235, opened 2021-04-20
docs: New docs for Header bidding and Glossary

arelra:
## What does this change?

Updated header bidding docs following knowledge sharing with @ioanna0 

Header bidding document:
https://github.com/guardian/commercial-core/blob/3e88dd272e83d86aeb1db5bfb110be1049b582b1/docs/header-bidding.md

I've also added a glossary:
https://github.com/guardian/commercial-core/blob/3e88dd272e83d86aeb1db5bfb110be1049b582b1/docs/glossary.md

The "code process" section I felt was necessary to reduce the time getting up to speed with the Frontend codebase and helping future migration efforts.

## Why?

Document knowledge for team and helpful for new joiners


docs/assets/header-bidding-process.excalidraw
```diff
@@ -0,0 +1,2215 @@
...
+      "height": 23,
+      "seed": 1051540609,
+      "groupIds": [
+        "elCBXVXy7oiDdkUWHL6g4"
+      ],
+      "strokeSharpness": "sharp",
+      "boundElementIds": [],
+      "fontSize": 17.62193886776144,
+      "fontFamily": 1,
+      "text": "Xasis",
```
mxdvl, line 990: This is wonderful if we can update the source directly.

Can we get the export to be generated by the CI?

```suggestion
      "text": "Xaxis",
```

arelra: I used the gui: https://excalidraw.com/

Are those SVGs? :)

I'm not sure how we could automate without something like puppeteer etc.

Also there are additional steps like cropping required.

I'm not sure its worth the dev time for what would be a relatively stable diagram...?

MarSavar (commented):
This is brilliant, thank you so much!

docs/assets/header-bidding-process.excalidraw
arelra: Automation would be cool though if we create more docs.

ioanna0 (approved):
great diagram and doc! 👍 🔝 📄

PR #249, opened 2021-05-18
feat: Add connection properties to event timer

zekehuntergreen:
## What does this change?
Adds information about a client's connection (bandwidth estimate and effective type) to the EventTimer object. 
The feature relies on  [Navigator.connection](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/connection) which is not yet supported by Firefox, Safari, or Internet Explorer.

## Why?
To supplement the performance data already being collected by EventTimer.


mxdvl (changes requested):
This is looking good, but I think we need to add a method to retrieve these properties. Maybe something like `getProperties()`?

Also, note that this is a `minor` version bump, so the title of the PR should be `feat: …` without an exclamation mark. Otherwise you're marking this release as a breaking change, which it isn't.

src/types.ts
```diff
@@ -49,5 +55,7 @@ declare global {
 			config?: GuardianWindowConfig;
 		};
 		ga: UniversalAnalytics.ga;
+		// navigator must be readonly
```
arelra, line 58: Minor - do we need this comment ? :)

zekehuntergreen: Yeah I suppose it's not all that useful anymore 😛

zekehuntergreen:
@mxdvl Is there somewhere where we'll need to use a new `EventTimer. getProperties` method immediately or do you think we should implement it in anticipation of needing it later? 
For the moment can we get away with accessing the attributes like `window.guardian.commercialTimer.properties.downlink ` ?

mxdvl (approved):

PR #254, opened 2021-05-26
feat: Send commercial metrics

zekehuntergreen:
## What does this change?
Moves logic around sending commercial metrics from front end

## Why?
- Keeps more commercial-specific code in commercial-core
- Makes logic more reusable especially in migration from frontend to DCR

### DIFF
--- a/src/sendCommercialMetrics.ts
+++ b/src/sendCommercialMetrics.ts
@@ -1,0 +1,66 @@
+import { log } from '@guardian/libs';
+import { EventTimer } from './EventTimer';
+
+type CommercialMetrics = {
+	browser_id?: string;
+	page_view_id: string;
+	received_timestamp: string;
+	received_date: string;
+	platform: string;
+	metrics: Metrics[];
+	properties: Properties[];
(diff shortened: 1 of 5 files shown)


MarSavar (approved):
👍 👍 👍

src/sendCommercialMetrics.ts
```diff
@@ -0,0 +1,66 @@
...
+		received_timestamp: timestamp,
+		received_date: date,
+		platform: 'NEXT_GEN',
+		metrics,
+		properties,
+	};
+
+	log('commercial', 'About to send commercial metrics', commercialMetrics);
+	return navigator.sendBeacon(endpoint, JSON.stringify(commercialMetrics));
+}
```
arelra, line 66: Do we have unit tests for this?

If not is it possible to have one simple test that will capture the main functionality?

zekehuntergreen: Good call - I've added a few tests

arelra (approved):

PR #267, opened 2021-06-08
fix: Filter eventTimer properties with undefined values

chrislomaxjones:
## What does this change?

Filters event timer properties with undefined values.

~Also introduces a `PropertyEntry` type and annotation in order to prevent a [no-unnecessary-condition
](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unnecessary-condition.md) eslint errror.~

## Why?

Prevents undefined properties being included and sent with commercial metrics.

### DIFF
--- a/src/sendCommercialMetrics.spec.ts
+++ b/src/sendCommercialMetrics.spec.ts
@@ -37,11 +37,7 @@ describe('sendCommercialMetrics', () => {
 					received_date: '2021-01-01',
 					platform: 'NEXT_GEN',
 					metrics: [],
-					properties: [
-						{ name: 'downlink', value: 'undefined' },
-						{ name: 'effectiveType', value: 'undefined' },
-						{ name: 'isDev', value: 'localhost' },
-					],
+					properties: [{ name: 'isDev', value: 'localhost' }],
 				}),
 			],
 		]);
--- a/src/sendCommercialMetrics.spec.ts
+++ b/src/sendCommercialMetrics.spec.ts
@@ -1,9 +1,34 @@
 import MockDate from 'mockdate';
+import { EventTimer } from './EventTimer';
 import { sendCommercialMetrics } from './sendCommercialMetrics';
+
+const DEFAULT_DATE = '2021-01-01T12:00:00.000Z';
+const PAGE_VIEW_ID = 'pv_id_1234567890';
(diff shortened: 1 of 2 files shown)


mxdvl:
Hey @chrislomaxjones, I'm not sure why we need to introduce a new type. I think the properties type should take care of this.

Maybe we need to refactor a little, as `downlink` and `effectiveType` are only available on Chrome.

cc. @zekehuntergreen

chrislomaxjones:
@mxdvl I get an eslint error if the type of the argument isn't annotated (see screenshot below). There may be a better solution to this though as the types do overlap!

Screenshot 2021-06-09 at 08 24 33

src/sendCommercialMetrics.spec.ts
```diff
@@ -19,39 +44,108 @@ describe('sendCommercialMetrics', () => {
...
-				JSON.stringify({
-					browser_id: 'browser id',
-					page_view_id: 'page view id',
-					received_timestamp: '2021-01-01T12:00:00.000Z',
-					received_date: '2021-01-01',
-					platform: 'NEXT_GEN',
-					metrics: [],
-					properties: [{ name: 'isDev', value: 'localhost' }],
-				}),
+				'//performance-events.guardianapis.com/commercial-metrics',
```
mxdvl, line 53: We could extract this in a constant, like `PROD_ENDPOINT`.

mxdvl (approved):
Increasing coverage, excellent 🎉!

src/sendCommercialMetrics.spec.ts
zekehuntergreen: Non-blocking: might be worth doing the same here? https://github.com/guardian/commercial-core/pull/267/files#diff-c076222fcae3be8e833dc2a29ce1870cce642fe2aab8c5504875182b46de95c6R34

zekehuntergreen (approved):

src/sendCommercialMetrics.spec.ts
mxdvl: Yes, and then exporting that `const` to reuse it in the test!

mxdvl:
Next up: raise a PR in `frontend` and `dotcom-rendering` to update to the latest!

PR #272, opened 2021-06-11
test: Additional unit tests for event timer, Inizio and Google Analytics

chrislomaxjones:
## What does this change?

This PR adds:
- Additional unit tests to EventTimer, testing multiple events and `mark`.
- Additional unit tests to the Inizio third party tag.
- Unit tests for google analytics.

## Why?

Increase code coverage!

### DIFF
--- a/src/EventTimer.spec.ts
+++ b/src/EventTimer.spec.ts
@@ -5,21 +5,21 @@ jest.mock('./GoogleAnalytics', () => ({
 	trackEvent: jest.fn(),
 }));
 
+const performance = {
+	now: jest.fn(),
+	mark: jest.fn(),
+	getEntriesByName: jest.fn().mockReturnValue([
+		{
+			duration: 1,
+			entryType: 'mark',
+			name: 'commercial event',
+			startTime: 1,
+		},
+	]),
+};
+
 describe('EventTimer', () => {
 	beforeEach(() => {
-		const performance = {
-			now: jest.fn(),
-			mark: jest.fn(),
-			getEntriesByName: jest.fn().mockReturnValue([
-				{
-					duration: 1,
-					entryType: 'mark',
-					name: 'commercial event',
-					startTime: 1,
-				},
-			]),
-		};
-
 		Object.defineProperty(window, 'performance', {
 			configurable: true,
 			enumerable: true,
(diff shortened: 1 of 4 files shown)


src/EventTimer.spec.ts
```diff
@@ -6,27 +6,27 @@ jest.mock('./GoogleAnalytics', () => ({
...
+			mark: jest.fn(),
+			getEntriesByName: jest.fn().mockReturnValue([
+				{
+					duration: 1,
+					entryType: 'mark',
+					name: 'commercial event',
+					startTime: 1,
+				},
+			]),
+		};
```
mxdvl, line 21: This `const performance` can still live outside of the `beforeEach`. You don’t need to redefine this variable each time, only assign to `window.performance`.

*not a blocker*

src/third-party-tags/inizio.ts
```diff
@@ -1,19 +1,20 @@
...
-		survey: { measurementId: string },
-	) => {
-		if (surveyAvailable) {
-			if (window.googletag) {
-				window.googletag.cmd.push(() => {
-					window.googletag?.pubads().setTargeting('inizio', 't');
-				});
-			}
-			console.log(`surveyAvailable: ${survey.measurementId}`);
+export const handleQuerySurveyDone = (
```
mxdvl, line 3: The convention for testing is to export an object called `_` (underscore) at the end of the file, like so:

```ts
export const _ = {
  handleQuerySurveyDone,
  onLoad,
}
```

See some examples here:
- https://github.com/guardian/frontend/blob/40c9d66bf4832b9aef7ca389db535b5e282f9185/static/src/javascripts/projects/common/modules/commercial/geo-utils.js#L26
- https://github.com/guardian/consent-management-platform/blob/2edc662cf92880cd38ad0e1d02e5a997baddf8df/src/onConsentChange.ts#L60

mxdvl (approved):
Excellent, our coverage is now 💯

src/third-party-tags/inizio.ts
chrislomaxjones: Great, good to know!

mxdvl:
Merge away!

PR #277, opened 2021-06-14
fix: Stop sending client-side date metrics

chrislomaxjones:
## What does this change?

Stop sending `received_timestamp` and `received_date` as part of commercial metrics.

## Why?

Commercial metrics will receive timestamps server-side (see https://github.com/guardian/fastly-edge-cache/pull/911).


mxdvl (approved):
Great, thanks!

PR #278, opened 2021-06-14
docs: update coverage and remove bundlephobia link

mxdvl:
## What does this change?

Updates the ReadMe.

## Why?

- We now have 100% coverage, and coveralls wasn’t picking up the main branch, it seemed
- We don’t release a cjs module, so bundlephobia cannot do its thing.


chrislomaxjones (approved):
+1!

PR #279, opened 2021-06-15
feat: Include CMP events in EventTimer

MarSavar:
Co-authored-by: Chris Jones <christopher.jones@guardian.co.uk>

## What does this change?
This change adds CMP-related timers to `EventTimer`, allowing us to send SourcePoint timers to the data lake.

CMP events are not tracked by `EventTimer`'s `trigger` function (see example [here](https://github.com/guardian/consent-management-platform/blob/c404fd68dc629525f4a70e25dfc97fc419a1eb97/src/tcfv2/sourcepoint.ts#L46)), so a static array including CMP event names has been added to to the fields of the `EventTimer` class, as well as a getter function that concatenates `EventTimer`'s private events to the external consent timers.

The tracked events are:
- `cmp-[framework]-init`
-  `cmp-[framework]-got-consent`

where `[framework]` is one of `tcfv2`, `ccpa`, or `aus`. 

Because CMP events are now included in `EventTimer`, no changes are needed to `sendCommercialMetrics`, so we will start seeing these new metrics flow into the lake when this PR is merged.

## Why?
Tracking CMP timers will give us a better understanding of how SP metrics relate to commercial timers.

### DIFF
--- a/src/EventTimer.spec.ts
+++ b/src/EventTimer.spec.ts
@@ -35,6 +35,11 @@ const performance = {
 		.mockReturnValue([]),
 };
 
+const MARK_NAME = 'mark_name';
+const MARK_LONG_NAME = `gu.commercial.${MARK_NAME}`;
+const CMP_INIT = 'cmp-tcfv2-init';
+const CMP_GOT_CONSENT = 'cmp-tcfv2-got-consent';
+
 describe('EventTimer', () => {
 	beforeEach(() => {
 		Object.defineProperty(window, 'performance', {
@@ -61,10 +66,7 @@ describe('EventTimer', () => {
 		const performanceCMP = {
 			now: jest.fn(),
 			mark: jest.fn(),
-			getEntriesByName: mockGetEntriesByName([
-				'cmp-tcfv2-init',
-				'cmp-tcfv2-got-consent',
-			]),
+			getEntriesByName: mockGetEntriesByName([CMP_INIT, CMP_GOT_CONSENT]),
 		};
 		Object.defineProperty(window, 'performance', {
 			configurable: true,
@@ -74,8 +76,8 @@ describe('EventTimer', () => {
 		});
 		const eventTimer = EventTimer.get();
 		expect(eventTimer.events).toHaveLength(2);
-		expect(eventTimer.events.map((event) => event.name)).toEqual(
-			expect.arrayContaining(['cmp-tcfv2-init', 'cmp-tcfv2-init']),
+		expect(eventTimer.events.map((event) => event.name).sort()).toEqual(
+			[CMP_GOT_CONSENT, CMP_INIT].sort(),
 		);
 	});
 
@@ -84,9 +86,9 @@ describe('EventTimer', () => {
 			now: jest.fn(),
 			mark: jest.fn(),
 			getEntriesByName: mockGetEntriesByName([
-				'gu.commercial.mark_name',
-				'cmp-tcfv2-init',
-				'cmp-tcfv2-got-consent',
+				MARK_LONG_NAME,
+				CMP_INIT,
+				CMP_GOT_CONSENT,
 			]),
 		};
 		Object.defineProperty(window, 'performance', {
@@ -96,22 +98,18 @@ describe('EventTimer', () => {
 			writable: true,
 		});
 		const eventTimer = EventTimer.get();
-		eventTimer.mark('mark_name');
+		eventTimer.mark(MARK_NAME);
 		expect(eventTimer.events).toHaveLength(3);
-		expect(eventTimer.events.map((event) => event.name)).toEqual(
-			expect.arrayContaining([
-				'cmp-tcfv2-init',
-				'cmp-tcfv2-got-consent',
-				'mark_name',
-			]),
+		expect(eventTimer.events.map((event) => event.name).sort()).toEqual(
+			[CMP_INIT, CMP_GOT_CONSENT, MARK_NAME].sort(),
 		);
 	});
 
 	it('mark produces correct event', () => {
 		const eventTimer = EventTimer.get();
-		eventTimer.mark('mark_name');
+		eventTimer.mark(MARK_NAME);
 		expect(eventTimer.events).toHaveLength(1);
-		expect(eventTimer.events[0].name).toBe('mark_name');
+		expect(eventTimer.events[0].name).toBe(MARK_NAME);
 		expect(eventTimer.events[0].ts).toBe(1);
 	});
 
@@ -123,7 +121,7 @@ describe('EventTimer', () => {
 			writable: true,
 		});
 		const eventTimer = EventTimer.get();
-		eventTimer.mark('mark_name');
+		eventTimer.mark(MARK_NAME);
 		expect(eventTimer.events).toEqual([]);
 	});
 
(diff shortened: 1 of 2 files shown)


src/EventTimer.ts
```diff
@@ -66,8 +74,32 @@ export class EventTimer {
...
+								new Event(
+									eventName,
+									window.performance.getEntriesByName(
+										eventName,
+									)[0] ?? 0,
+								),
+						)
+						.filter((event) => event.ts > 0),
+			  ]
+			: [...this._events];
```
arelra, line 98: Is this intended to create a copy of the array or could you just return `this._events` which is an array already?

MarSavar: Very good point - no need to create a copy. Thanks

src/EventTimer.ts
```diff
@@ -66,8 +74,32 @@ export class EventTimer {
...
+								new Event(
+									eventName,
+									window.performance.getEntriesByName(
+										eventName,
+									)[0] ?? 0,
+								),
+						)
+						.filter((event) => event.ts > 0),
+			  ]
+			: this._events;
```
arelra, line 98: Just wondering if this spreading logic is captured by any tests?

src/EventTimer.ts
```diff
@@ -66,8 +74,32 @@ export class EventTimer {
 		return this.init();
 	}
 
+	/**
+	 * Returns all commercial timers. CMP-related timers are not tracked
+	 * by EventTimer so they need to be concatenated to EventTimer's private events array.
+	 */
+	public get events(): Event[] {
+		return typeof window.performance !== 'undefined' &&
+			'mark' in window.performance
```
chrislomaxjones, line 83: Just checking again and noticed this should be `'getEntriesByName' in window.performance`

src/EventTimer.ts
```diff
@@ -66,8 +74,32 @@ export class EventTimer {
...
+						.map(
+							(eventName) =>
+								new Event(
+									eventName,
+									window.performance.getEntriesByName(
+										eventName,
+									)[0] ?? 0,
+								),
+						)
+						.filter((event) => event.ts > 0),
```
chrislomaxjones, line 96: We could probably be more explicit here and filter when `window.performance.getEntriesByName(eventName)` is empty rather than setting the timestamp to `0` and filter on `event.ts > 0`.

It might be clearer on first glance to see we're removing those with no entries rather than those with `0` timestamps.

MarSavar: That's a valid point, and something I had considered. In fact, the original implementation was exactly as you describe, but it broke when I tested it locally because of the way the `Event` constructor worked (it couldn't read `startTime` of `undefined`). But on second thought, this happened because I was `map`ping before `filter`ing. Swapping the order around solved the issue. Tried it locally and it works, will commit in a second.

src/EventTimer.ts
mxdvl: I think you can use optional chaining here:

```ts
window.performance?.mark ? 'a' : 'b';
```

mxdvl (commented):
Looking good, curious about the tests.

src/EventTimer.spec.ts
```diff
@@ -50,6 +57,56 @@ describe('EventTimer', () => {
...
+		Object.defineProperty(window, 'performance', {
+			configurable: true,
+			enumerable: true,
+			value: performanceCMP,
+			writable: true,
+		});
+		const eventTimer = EventTimer.get();
+		expect(eventTimer.events).toHaveLength(2);
+		expect(eventTimer.events.map((event) => event.name)).toEqual(
+			expect.arrayContaining(['cmp-tcfv2-init', 'cmp-tcfv2-init']),
```
mxdvl, line 78: this array is different from the array above.

```suggestion
			expect.arrayContaining(['cmp-tcfv2-init', 'cmp-tcfv2-got-consent']),
```

You might want to extract this to a const inside the test function anyways? Keep it DRY?

MarSavar: Fixed it, thanks!

src/EventTimer.ts
MarSavar: Done now!

src/EventTimer.spec.ts
```diff
@@ -8,14 +8,24 @@ jest.mock('./GoogleAnalytics', () => ({
...
+			},
+		])
+		.mockReturnValue([
+			{
+				duration: 1,
+				entryType: 'mark',
+				name: 'commercial event',
+				startTime: 0,
+			},
+		]),
```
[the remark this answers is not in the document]
chrislomaxjones: This was originally to mock a `getEntriesByName` function that makes only one mark that has a startTime > 1 (and all of the ones < 1 get filtered). However this isn't actually a good way of mocking it so we've replaced it with a mock function that returns marks matching certain names (see `mockGetEntriesByName`).

src/EventTimer.spec.ts
mxdvl: I’m sort of worried that the test passed in the first place… any idea why that was?

src/EventTimer.ts
MarSavar: We tried doing this but the compiler thinks `window.performance` is always defined.

src/EventTimer.spec.ts
MarSavar: We found out it's because `arrayContaining` checks for elements being present in the array *at least* once, so it doesn't throw an error if the same element appears twice. We got around the issue by sorting the expected and the actual arrays and checking for equality instead.

src/EventTimer.ts
```diff
@@ -66,8 +74,36 @@ export class EventTimer {
 		return this.init();
 	}
 
+	/**
+	 * Returns all commercial timers. CMP-related timers are not tracked
+	 * by EventTimer so they need to be concatenated to EventTimer's private events array.
+	 */
+	public get events(): Event[] {
```
zekehuntergreen, line 81: I didn't know about this syntax for getters in js! 💡

src/EventTimer.spec.ts
```diff
@@ -109,8 +109,8 @@ describe('EventTimer', () => {
 		const eventTimer = EventTimer.get();
 		eventTimer.mark(MARK_NAME);
 		expect(eventTimer.events).toHaveLength(1);
-		expect(eventTimer.events[0].name).toBe(MARK_NAME);
-		expect(eventTimer.events[0].ts).toBe(1);
+		expect(eventTimer.events[0]?.name).toBe('mark_name');
```
mxdvl, line 112: Cheeky merging from `main`!

```suggestion
		expect(eventTimer.events[0]?.name).toBe(MARK_NAME);
```

src/EventTimer.ts
```diff
@@ -32,7 +32,15 @@ interface PageEventStatus {
 }
 
 export class EventTimer {
-	events: Event[];
+	private _events: Event[];
+	private static _externallyDefinedEventNames = [
+		'cmp-tcfv2-init',
```
mxdvl, line 37: I think it would be worth having a reference to where these come from?

- `got-consent` https://git.io/Jnnyv
- `init` https://git.io/Jnny2
- etc.?

And also add `cmp-***-ui-displayed`, as https://git.io/JnnyO ?

mxdvl (approved):
I love it, really I do.

Small comments, not blockers, can come in a subsequent PR.

src/EventTimer.ts
MarSavar: Good shout, thanks! Will refactor in the next couple of days. Merging this now 🚀

PR #288, opened 2021-06-22
chore(deps): Bump dependencies 2021.06.22

zekehuntergreen:
## What does this change?
Upgrades dependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "prettier": "^2",
     "semantic-release": "^17",
     "ts-jest": "^26",
-    "typescript": "^4"
+    "typescript": "4.3.4"
   },
   "publishConfig": {
     "access": "public"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -69,7 +69,7 @@
     "prettier": "^2",
     "semantic-release": "^17",
     "ts-jest": "^26",
-    "typescript": "^4"
```
zekehuntergreen, line 72: Is there a reason not to specify an exact version number here?

zekehuntergreen: 🤖 ok we'll do that

mxdvl (approved):

package.json
mxdvl: I think there’s a clearer separation of human 🧑 / robot 🤖  between `package.json` / `yarn.lock`

PR #289, opened 2021-06-22
fix: Error when window.ga is unavailable.

ashishpuliyel:
Sometimes window.ga is null, not undefined, and this previously caused a problem.

## What does this change?
Checks if window.ga is a function, instead of checking if it's not undefined.

## Why?
In some circumstance it appears that window.ga is set to null, when CCPA opt-out is applied, for example.


mxdvl (approved):
Seems good.

Make sure to edit the commit message when merging, otherwise it won't release.

PR #290, opened 2021-06-25
docs: ad render flow

arelra:
## What does this change?

Add documentation for current frontend ad rendering flow

## Why?

To understand and document our ad rendering stack.


mxdvl (approved):
This is awesome! Thanks @arelra 🚀

PR #311, opened 2021-07-20
chore(deps): Bump dependencies 2021.07.20

MarSavar:
## What does this change?
Bump dependencies


zekehuntergreen (approved):

mxdvl:
Nice! Running `yarn update` is also helpful to resolve even newer matching dependencies.

PR #351, opened 2021-09-06
fix: Migrate from `@guardian/types` to `@guardian/libs`

jamie-lynch:
## What does this change?

This PR migrates from the [`@guardian/types`](https://github.com/guardian/types) library to [`@guardian/libs`](https://github.com/guardian/libs) as [`@guardian/types` is now deprecated](https://github.com/guardian/types/pull/140).

For this repository, that simply means dropping `@guardian/types` from both `devDependencies` and `peerDependencies` as it wasn't being used anywhere.

## How to test

Run all validation and ensure that everything still runs as expected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,6 @@
     "@guardian/eslint-config-typescript": "^0.6",
     "@guardian/libs": "^3",
     "@guardian/prettier": "^0.6",
-    "@guardian/types": "^8",
     "@octokit/core": "^3",
     "@semantic-release/github": "^7",
     "@types/doubleclick-gpt": "^2019111201.0.0",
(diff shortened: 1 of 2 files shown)


mxdvl (approved):
Thanks for this!

jamie-lynch:
> Thanks for this!

@mxdvl Thanks for approving! I don't have permissions to merge so are you okay to merge this?

Also, I've made it a breaking change due to the change in peer deps but as it's simply dropping a dependency, maybe this isn't actually a breaking change. Do you have any thoughts of whether to make this a new major/minor/patch version?

mxdvl:
@jamie-lynch actually, `commercial-core` hasn’t even reached v1, so I’ll make this a `fix:`/`patch`.

PR #353, opened 2021-09-09
feat: Migrate ad size definitions to commercial core

chrislomaxjones:
## What does this change?

Migrate the size definitions that are currently defined separately in [frontend](https://github.com/guardian/frontend/blob/main/static/src/javascripts/projects/commercial/modules/ad-sizes.js) and [DCR](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/components/AdSlot.tsx) to commercial core.

## Why?

- Frontend and DCR size definitions won't need to be manually kept in sync.
- These are important definitions that can be appropriately documented here.
- One of the first steps in a longer-term goal to move commercial code out of frontend.

### DIFF
--- a/src/ad-sizes.test.ts
+++ b/src/ad-sizes.test.ts
@@ -1,0 +1,20 @@
+import { _ } from './ad-sizes';
+
+const { getAdSize } = _;
+
+describe('ad sizes', () => {
+	it.each([
+		[getAdSize(2, 2), 2, 2, '2,2'],
+		[getAdSize(100, 100), 100, 100, '100,100'],
+		[getAdSize(320, 50), 320, 50, '320,50'],
+		[getAdSize(970, 250), 970, 250, '970,250'],
+		[getAdSize(0, 0), 0, 0, 'fluid'],
+	])(
+		'getAdSizes outputs correct values',
+		(adSize, expectedWidth, expectedHeight, expectedString) => {
+			expect(adSize.width).toEqual(expectedWidth);
+			expect(adSize.height).toEqual(expectedHeight);
+			expect(adSize.toString()).toEqual(expectedString);
+		},
+	);
+});
--- a/src/ad-sizes.test.ts
+++ b/src/ad-sizes.test.ts
@@ -4,14 +4,15 @@ const { getAdSize } = _;
 
 describe('ad sizes', () => {
 	it.each([
-		[getAdSize(2, 2), 2, 2, '2,2'],
-		[getAdSize(100, 100), 100, 100, '100,100'],
-		[getAdSize(320, 50), 320, 50, '320,50'],
-		[getAdSize(970, 250), 970, 250, '970,250'],
-		[getAdSize(0, 0), 0, 0, 'fluid'],
+		[2, 2, 2, 2, '2,2'],
+		[100, 100, 100, 100, '100,100'],
+		[320, 50, 320, 50, '320,50'],
+		[970, 250, 970, 250, '970,250'],
+		[0, 0, 0, 0, 'fluid'],
 	])(
-		'getAdSizes outputs correct values',
-		(adSize, expectedWidth, expectedHeight, expectedString) => {
+		'getAdSize(%d,%d) outputs width %d, height %d, string %s',
+		(width, height, expectedWidth, expectedHeight, expectedString) => {
+			const adSize = getAdSize(width, height);
 			expect(adSize.width).toEqual(expectedWidth);
 			expect(adSize.height).toEqual(expectedHeight);
 			expect(adSize.toString()).toEqual(expectedString);
--- a/src/ad-sizes.test.ts
+++ b/src/ad-sizes.test.ts
@@ -4,15 +4,15 @@ const { getAdSize } = _;
 
 describe('ad sizes', () => {
 	it.each([
-		[2, 2, 2, 2, '2,2'],
-		[100, 100, 100, 100, '100,100'],
-		[320, 50, 320, 50, '320,50'],
-		[970, 250, 970, 250, '970,250'],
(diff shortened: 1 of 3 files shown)


src/ad-sizes.ts
```diff
@@ -0,0 +1,87 @@
+export type GuAdSizeString = 'fluid' | `${number},${number}`;
+
+export type GuAdSize = Readonly<{
```
chrislomaxjones, line 3: Do we want to keep adding the `Gu` prefix to types when we're migrating them?

src/ad-sizes.ts
```diff
@@ -0,0 +1,87 @@
+export type GuAdSizeString = 'fluid' | `${number},${number}`;
```
chrislomaxjones, line 1: Where should types like these live? I'd prefer to keep them as close as possible to the functions that use them rather than define them globally but could see why we might want to define them in `./types.ts`

src/ad-sizes.ts
```diff
@@ -0,0 +1,87 @@
...
+	outstreamMobile: getAdSize(300, 197),
+	merchandisingHighAdFeature: getAdSize(88, 89),
+	merchandisingHigh: getAdSize(88, 87),
+	merchandising: getAdSize(88, 88),
+	inlineMerchandising: getAdSize(88, 85),
+	fabric: getAdSize(88, 71),
+	empty: getAdSize(2, 2),
+};
+
+export const adSizes: Record<SizeKeys, GuAdSize> = {
```
chrislomaxjones, line 76: Since we don't export `getAdSize` and `GuAdSize` is readonly we can't construct types of GuAdSize outside of the ones we export in `adSizes`.

src/ad-sizes.test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+describe('ad sizes', () => {
+	it.each([
+		[getAdSize(2, 2), 2, 2, '2,2'],
+		[getAdSize(100, 100), 100, 100, '100,100'],
+		[getAdSize(320, 50), 320, 50, '320,50'],
+		[getAdSize(970, 250), 970, 250, '970,250'],
+		[getAdSize(0, 0), 0, 0, 'fluid'],
+	])(
+		'getAdSizes outputs correct values',
```
mxdvl, line 13: You can call `getAdSize` inside each call of the test, and use %s for each run to have relevant info.

For example: `getAdSize outputs %s for %n and %n`.

mxdvl (commented):
I'd be tempted to drop the `Gu` prefix and maybe use `enum` over `const`.

I find it easier as well to have types close to where they are used.

src/ad-sizes.ts
mxdvl: Close is good!

src/ad-sizes.test.ts
chrislomaxjones: Nice, didn't know you could do that!

src/ad-sizes.test.ts
```diff
@@ -0,0 +1,21 @@
...
+	it.each([
+		[2, 2, 2, 2, '2,2'],
+		[100, 100, 100, 100, '100,100'],
+		[320, 50, 320, 50, '320,50'],
+		[970, 250, 970, 250, '970,250'],
+		[0, 0, 0, 0, 'fluid'],
+	])(
+		'getAdSize(%d,%d) outputs width %d, height %d, string %s',
+		(width, height, expectedWidth, expectedHeight, expectedString) => {
+			const adSize = getAdSize(width, height);
```
mxdvl, line 15:
```suggestion
	it.each([
		[2, 2, '2,2'],
		[100, 100, '100,100'],
		[320, 50, '320,50'],
		[970, 250, '970,250'],
		[0, 0, 'fluid'],
	])(
		'getAdSize(%d,%d) outputs string "%s"',
		(expectedWidth, expectedHeight, expectedString) => {
			const adSize = getAdSize(expectedWidth, expectedHeight);
```

src/ad-sizes.ts
```diff
@@ -0,0 +1,87 @@
...
+	| '300x250'
+	| '300x600'
+	| '300x1050'
+	| '160x600';
+
+const getAdSize = (width: number, height: number): AdSize => {
+	const toString = (): AdSizeString =>
+		width === 0 && height === 0 ? 'fluid' : `${width},${height}`;
+
+	return Object.freeze({
```
arelra, line 41: Not related to the PR - I know we've replicated this across from Frontend but I'm curious to understand why we create a frozen object with a `toString`?

Just thinking ahead to futureland where we try and simplify the codebase, could this just be a string - in which case we don't need `getAdSize` and can just string template the sizes?

Something we can discuss with the team.

src/ad-sizes.ts
```diff
@@ -0,0 +1,87 @@
...
+};
+
+export const adSizes: Record<SizeKeys, AdSize> = {
+	...adSizesPartial,
+	'970x250': adSizesPartial.billboard,
+	'728x90': adSizesPartial.leaderboard,
+	'300x250': adSizesPartial.mpu,
+	'300x600': adSizesPartial.halfPage,
+	'300x1050': adSizesPartial.portrait,
+	'160x600': adSizesPartial.skyscraper,
```
arelra, line 83: Again not PR related but I'd like to understand these custom sizes that are just copies of the standard sizes. Something to consolidate? We can discuss.

arelra (approved):

src/ad-sizes.ts
chrislomaxjones: Good point - I wonder whether we still need to maintain the readonly object (we wouldn't need it in DCR) but are the `.width` and `.height` properties used in frontend? Will note this as a future change we should make.

PR #364, opened 2021-09-14
chore(tests): set testEnvironment globally

mxdvl:
## What does this change?

All our unit tests are meant to be running in the browser, so we set `testEnvironment` to `jsdom` in the Jest config.


chrislomaxjones (approved):
Looks good! Though you might have to amend your commit message to get the checks to pass

PR #369, opened 2021-09-21
feat: Add ad block property to `sendCommercialMetrics `

zekehuntergreen:
## What does this change?
Adds an extra parameter to the `sendCommercialMetrics` function which is put into the `properties` of the payload.

## Why?
We want to collect data on readers' use of ad blockers.

### DIFF
--- a/src/sendCommercialMetrics.ts
+++ b/src/sendCommercialMetrics.ts
@@ -28,6 +28,15 @@ export function sendCommercialMetrics(
 	const devProperties: Properties[] = isDev
 		? [{ name: 'isDev', value: window.location.hostname }]
 		: [];
+	const adBlockerProperties: Properties[] =
+		adBlockerInUse !== undefined
+			? [
+					{
+						name: 'adBlockerInUse',
+						value: adBlockerInUse.toString(),
+					},
+			  ]
+			: [];
 
 	const endpoint = isDev
 		? '//performance-events.code.dev-guardianapis.com/commercial-metrics'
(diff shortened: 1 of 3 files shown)


mxdvl:
Would it be possible to make this change non-breaking by making `adBlockerInUse` an optional parameter?

Otherwise I’d be inclined to make the third parameter a `Record<string, string>` with some restricted keys, akin to [sentinel.ts](https://github.com/guardian/frontend/blob/25a24aac4e11d116b17534c17bb082ecfb7a51ba/static/src/javascripts/projects/commercial/sentinel.ts#L35-L37)

src/sendCommercialMetrics.ts
```diff
@@ -41,6 +42,12 @@ export function sendCommercialMetrics(
 		.map(([name, value]) => ({ name, value: String(value) }))
 		.concat(devProperties);
 
+	if (adBlockerInUse !== undefined)
+		properties.push({
+			name: 'adBlockerInUse',
+			value: adBlockerInUse.toString(),
+		});
+
```
mxdvl, line 50: If you want to keep the functional style, without mutation, you’d extract this as a `const` and concatenate it like we do fo `devProperties`.

You could enforce this by changing the type on line 40 to be `readonly`:
```ts
const properties: readonly Properties[] = Object.entries(
```

mxdvl (commented):
Will give a green tick once the PR title is updated to be a minor version 👍

src/sendCommercialMetrics.ts
zekehuntergreen: Good call! Looks like `CommercialMetrics` also needs to be changed so that `properties` is `readonly`

src/sendCommercialMetrics.ts
```diff
@@ -6,7 +6,7 @@ type CommercialMetrics = {
 	page_view_id: string;
 	platform: string;
 	metrics: Metrics[];
```
mxdvl, line 8: If we’re there… we might as well:

```suggestion
	metrics: readonly Metrics[];
```

mxdvl (approved):

src/sendCommercialMetrics.ts
zekehuntergreen: Done

PR #371, opened 2021-09-23
fix: Handle case where `window.performance = undefined`

simonbyford:
## What does this change?

[Some older browsers don't support the Performance API](https://developer.mozilla.org/en-US/docs/Web/API/Performance#browser_compatibility). Mostly we guard against this case by checking if `window.performance` is `undefined` before calling methods on it, but we discovered one instance where this wasn't happening, so we fixed it.

## Why?

We believe this to be the cause of a number of [Sentry errors](https://sentry.io/organizations/the-guardian/issues/?groupStatsPeriod=24h&project=35463&query=is%3Aunresolved++window.performance&sort=freq)

### DIFF
--- a/src/GoogleAnalytics.ts
+++ b/src/GoogleAnalytics.ts
@@ -3,14 +3,14 @@ export const trackEvent = (
 	timingVar: string,
 	timingLabel: string,
 ): void => {
-	const { ga, guardian } = window;
+	const { ga, guardian, performance } = window;
 	const trackerName: string | undefined =
 		guardian.config?.googleAnalytics?.trackers.editorial;
 
-	if (typeof ga !== 'function') {
+	if (typeof ga !== 'function' || typeof performance == 'undefined') {
 		return;
 	}
-	const timeSincePageLoad: number = Math.round(window.performance.now());
+	const timeSincePageLoad: number = Math.round(performance.now());
 
 	const send = trackerName ? `${trackerName}.send` : 'send';
 	window.ga(
--- a/src/GoogleAnalytics.spec.ts
+++ b/src/GoogleAnalytics.spec.ts
@@ -34,6 +34,27 @@ describe('trackEvent', () => {
 		expect(ga.mock.calls.length).toBe(0);
 	});
 
+	it('trackEvent makes no call to ga when window.performance is undefined', () => {
+		const performance = window.performance;
+
+		Object.defineProperty(window, 'performance', {
+			configurable: true,
+			enumerable: true,
+			value: undefined,
+			writable: true,
+		});
+
+		trackEvent(TIMING_CATEGORY, TIMING_VAR, TIMING_LABEL);
+		expect(ga.mock.calls.length).toBe(0);
+
+		Object.defineProperty(window, 'performance', {
+			configurable: true,
+			enumerable: true,
+			value: performance,
+			writable: true,
+		});
+	});
+
 	it('trackEvent makes one call to ga with trackername from config', () => {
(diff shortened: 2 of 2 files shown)


mxdvl (approved):
Wonderful!

Make sure you edit the squash message, as if you have a single commit, it will take that commit message instead of the PR title. That’s why you have a failing check.

mxdvl:
FYI you can reference the Sentry error by its ID: `CLIENT-SIDE-PROD-BK7Z`

src/GoogleAnalytics.ts
```diff
@@ -3,14 +3,14 @@ export const trackEvent = (
 	timingVar: string,
 	timingLabel: string,
 ): void => {
-	const { ga, guardian } = window;
+	const { ga, guardian, performance } = window;
 	const trackerName: string | undefined =
 		guardian.config?.googleAnalytics?.trackers.editorial;
 
-	if (typeof ga !== 'function') {
+	if (typeof ga !== 'function' || typeof performance == 'undefined') {
```
simonbyford, line 10: Open question: is this actually the desired behaviour? 

It seems like we have two options when `performance.window` isn't available:

- return early and omit to call to `window.ga` (this is what we do in this PR)
- call `window.ga` anyway with some dummy value for `timeSincePageLoad`

Any thoughts?

simonbyford:
> FYI you can reference the Sentry error by its ID: CLIENT-SIDE-PROD-BK7Z

Ah yeah true, but this bug appears to be behind many separate Sentry issues. For example:

```
CLIENT-SIDE-PROD-BDSX
CLIENT-SIDE-PROD-BKBT
CLIENT-SIDE-PROD-BKHT
etc.
```

I believe they all have the same cause but Sentry hasn't grouped them together for some reason. Maybe because the same bug was present across several releases? Looking at the file names that seems to make sense. For example:

```
https://assets.guim.co.uk/javascripts/e116ecf4c14c5585d2cb/graun.commercial.dcr.js
https://assets.guim.co.uk/javascripts/f21bdfb712776a5847d8/graun.commercial.dcr.js
etc.
```

However, I might be totally misunderstanding how Sentry works!

simonbyford:
> Make sure you edit the squash message, as if you have a single commit, it will take that commit message instead of the PR title. That’s why you have a failing check.

Nice one, thank you! I've fixed this now

src/GoogleAnalytics.ts
mxdvl: [Polyfill.io does this](https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/performance/now/polyfill.js):

```ts
var startTime = Date.now();

if (!window.performance) {
	window.performance = {};
}

window.performance.now = function () {
	return Date.now() - startTime;
};
```

I’m wondering if the fix would actually be simply to add this feature to [our polyfill](https://github.com/guardian/frontend/blob/main/static/src/javascripts/polyfill.io)?

MDN Docs: https://developer.mozilla.org/en-US/docs/Web/API/Performance/now

simonbyford: Ahhhh interesting point

simonbyford: This makes good sense to me. I'll raise a couple of PRs (in `DCR `and `frontend`) and close this one if we collectively decide to go down this route.

simonbyford:
Closing this in favour of a different approach:

https://github.com/guardian/frontend/pull/24210
https://github.com/guardian/dotcom-rendering/pull/3463

PR #379, opened 2021-09-30
feat: Detect ad blockers

zekehuntergreen:
## What does this change?
- Move code implemented in [frontend](https://github.com/guardian/frontend/blob/3e91c6942494aabf089de63000a5eeb81ddf8405/common/app/templates/inlineJS/nonBlocking/detectAdblock.scala.js#L6) and [dcr](https://github.com/guardian/dotcom-rendering/blob/150fc2d81e6a66d9c3336185e874fc8cd0288546/dotcom-rendering/src/web/lib/detectAdBlocker.ts#L3) so that it can be shared.
- Add to the classes of the fake ad element and conditions under which we consider it to be blocked in order to make function more reliable for certain ad blockers (see [spreadsheet](https://docs.google.com/spreadsheets/d/18n2N5p8ARA4R2ySfe2-bvocCuvJjApdtbMu9fa1Ol8k/edit?usp=sharing)). Code borrowed from [just-detect-adblock](https://github.com/wmcmurray/just-detect-adblock)

## Why?
- Remove duplicated code so that future changes can be made more easily
- Increase accuracy of ad blocker detection

### DIFF
--- a/src/detectAdBlocker.ts
+++ b/src/detectAdBlocker.ts
@@ -3,6 +3,8 @@ Detect whether or not the user has an ad blocking extension enabled.
 A few ad blockers are not detectable with this approach e.g. Safari / Adblock
 Code inspired by just-detect-adblock's: https://git.io/JgL4L
 */
+/*istanbul ignore file -- adElementBlocked can't be tested without patching each of the properties of
+HTMLElement.prototype that it accesses, defeating the purpose of the test! */
 
 let adBlockInUse: boolean | undefined = undefined;
 
@@ -17,7 +19,6 @@ function adElementBlocked(ad: HTMLElement): boolean {
 		ad.clientWidth === 0
 	)
 		return true;
-
 	const adStyles = window.getComputedStyle(ad);
 
 	if (adStyles.getPropertyValue('display') === 'none') return true;
--- a/src/detectAdBlocker.ts
+++ b/src/detectAdBlocker.ts
@@ -1,0 +1,64 @@
+/**
+Detect whether of not the user has an ad-blocking extension enabled.
+A few ad blockers are not detectable with this approach e.g. Safari / Adblock
+Code inspired by just-detect-adblock's: https://git.io/JgL4L
+*/
+
+let adBlockInUse: boolean | undefined = undefined;
+
+function adElementBlocked(ad: HTMLElement): boolean {
+	if (
+		ad.offsetParent === null ||
+		ad.offsetHeight === 0 ||
+		ad.offsetLeft === 0 ||
+		ad.offsetTop === 0 ||
+		ad.offsetWidth === 0 ||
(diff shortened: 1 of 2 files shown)


src/detectAdBlocker.ts
```diff
@@ -0,0 +1,65 @@
+/**
+Detect whether or not the user has an ad blocking extension enabled.
+A few ad blockers are not detectable with this approach e.g. Safari / Adblock
+Code inspired by just-detect-adblock's: https://git.io/JgL4L
+*/
+/*istanbul ignore file -- adElementBlocked can't be tested without patching each of the properties of
+HTMLElement.prototype that it accesses, defeating the purpose of the test! */
```
zekehuntergreen, line 7: See https://github.com/jsdom/jsdom/issues/1261
Is anyone aware of a workaround for this so that this module can be meaningfully tested with jest? 
Or would we need end-to-end tests in order to properly test this functionality?

src/detectAdBlocker.ts
```diff
@@ -0,0 +1,65 @@
...
+	if (adStyles.getPropertyValue('display') === 'none') return true;
+	if (adStyles.getPropertyValue('visibility') === 'hidden') return true;
+
+	const mozBindingProp = adStyles.getPropertyValue('-moz-binding');
+	if (mozBindingProp.includes('about:')) return true;
+
+	return false;
+}
+
+export function isAdBlockInUse(): Promise<boolean> {
```
mxdvl, line 33: Could we add a comment indicating how reliable this is?

For example, `false` is maybe and `true` is definitely.

mxdvl (approved):
This is a great addition.

Is there a reason not to use the `just-detect-adblock` npm package directly?

src/detectAdBlocker.ts
mxdvl: You could simply test that removing the ad element between two `window.requestAnimationFrame` calls results in a `true` response.

zekehuntergreen:
@mxdvl - There are a few ad blockers like [Ghostery on Chrome](https://docs.google.com/spreadsheets/d/18n2N5p8ARA4R2ySfe2-bvocCuvJjApdtbMu9fa1Ol8k/edit#gid=0) for which `just-detect-adblock` gives a false negative but our current approach gives a true positive, hence the combined approaches.

src/detectAdBlocker.ts
zekehuntergreen: I thought about doing something like this but then wasn't really sure that the test would provide much value. 
Removing the element from the document wouldn't be very realistic as this isn't what the ad blockers really do. And there's a risk that that sort of test would provide a false sense of security if someone breaks `adElementBlocked` in the future but tests pass.
But happy to give that a try if there's something I'm missing.

PR #396, opened 2021-10-14
feat: Add commercial CONSTANTS

mxdvl:
## What does this change?

- add TOP_ABOVE_NAV constant
- add commercial jargon “viewability”

## Why?

Centralise core commercial decisions and past experiments.

### DIFF
--- a/src/constants/topAboveNavHeight.ts
+++ b/src/constants/topAboveNavHeight.ts
@@ -1,0 +1,43 @@
+/**
+ * Unit: pixels.
+ *
+ * The majority of ads in the top banner are 250px high. We ran an experiment
+ * in October 2021 to set the minimum height to 250, and let smaller ads be
+ * centred in the space. We did not process with this option, as it had a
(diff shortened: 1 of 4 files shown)


src/constants/topAboveNavHeight.ts
```diff
@@ -0,0 +1,43 @@
...
+There was a very positive impact on CLS (Cumulative Layout Shift), which is good
+for UX. However, the negative commercial impact meant we kept a height of 90px.
+
+We ran a 1% server-side experiment to measure CLS when dedicating 250px for the
+topAboveNav. The experiment showed this change has a significant positive impact
+on CLS, and moves average CLS for the page from 0.09 to 0.07 (a 26% improvement
+from this one change). The other way we sliced the data was to look at the
+percent of pages that Google categorised as having 'good', 'needs improvement'
+or 'poor' CLS scores.
+
```
arelra, line 25: This is really helpful. Could we also add the metrics for the change in `viewability`?

PR #397, opened 2021-10-17
feat: YouTube ad targeting builder

arelra:
## What does this change?

YouTube ad targeting builder that can used in DCR

## Why?

Allows an exportable method to create YouTube ad targeting with consent state.

Wrapped in a promise to make it easier for YouTube components to consume.

### DIFF
--- a/src/ad-targeting-youtube.ts
+++ b/src/ad-targeting-youtube.ts
@@ -10,9 +10,8 @@ import type {
 	AdsConfigCCPAorAus,
 	AdsConfigDisabled,
 	AdsConfigTCFV2,
+	MaybeArray,
 } from './types';
-
-type MaybeArray<T> = T[] | T;
 
 type CustomParams = Record<string, MaybeArray<string | number | boolean>>;
 
@@ -58,9 +57,8 @@ const buildAdsConfig = (
 		} as AdsConfigCCPAorAus;
 	}
 
-	const tcfData = cmpConsent.tcfv2;
-	// ConsentState type allows for an undefined tcfv2
-	if (tcfData) {
+	if (cmpConsent.tcfv2) {
+		const tcfData = cmpConsent.tcfv2;
 		const canTarget = Object.values(tcfData.consents).every(Boolean);
 		const mergedAdTagParameters = {
 			...defaultAdsConfig.adTagParameters,
@@ -106,4 +104,4 @@ const buildAdsConfigWithConsent = async (
 		: disabledAds;
 };
 
-export { buildAdsConfigWithConsent };
+export { buildAdsConfigWithConsent, disabledAds };
--- a/src/types.ts
+++ b/src/types.ts
@@ -40,16 +40,16 @@ export type CustomParams = Record<
 	MaybeArray<string | number | boolean>
 >;
 
-export interface AdsConfigDisabled {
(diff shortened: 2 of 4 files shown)


src/ad-targeting-youtube.ts
```diff
@@ -0,0 +1,109 @@
...
+import { getPermutivePFPSegments } from './permutive';
+import type {
+	AdsConfig,
+	AdsConfigBasic,
+	AdsConfigCCPAorAus,
+	AdsConfigDisabled,
+	AdsConfigTCFV2,
+} from './types';
+
+type MaybeArray<T> = T[] | T;
```
mxdvl, line 15: Should be imported from `../types`

src/ad-targeting-youtube.ts
```diff
@@ -0,0 +1,109 @@
...
+		return {
+			...defaultAdsConfig,
+			restrictedDataProcessor: !canTarget,
+		} as AdsConfigCCPAorAus;
+	}
+
+	const tcfData = cmpConsent.tcfv2;
+	// ConsentState type allows for an undefined tcfv2
+	if (tcfData) {
+		const canTarget = Object.values(tcfData.consents).every(Boolean);
```
mxdvl, line 64: For consistency, I would follow the same pattern as above:

```suggestion
	if (cmpConsent.tcfv2) {
		const tcfData = cmpConsent.tcfv2;
		const canTarget = Object.values(tcfData.consents).every(Boolean);
```

src/types.ts
```diff
@@ -32,3 +32,43 @@ export type GoogleTrackConversionObject = {
...
+export type MaybeArray<T> = T | T[];
+
+export type CustomParams = Record<
+	string,
+	MaybeArray<string | number | boolean>
+>;
+
+export interface AdsConfigDisabled {
+	disableAds: true;
+}
```
mxdvl, line 45: Why the mix & match between interfaces and types?

yarn.lock
```diff
@@ -479,6 +479,13 @@
   resolved "https://registry.yarnpkg.com/@gar/promisify/-/promisify-1.1.2.tgz#30aa825f11d438671d585bd44e7fd564535fc210"
   integrity sha512-82cpyJyKRoQoRi+14ibCeGPu0CwypgtBAdBhq1WfvagpCZNKqwXbKwXllYSMG91DhmG4jt9gN8eP6lGOtozuaw==
 
+"@guardian/consent-management-platform@^8":
+  version "8.0.0"
+  resolved "https://registry.yarnpkg.com/@guardian/consent-management-platform/-/consent-management-platform-8.0.0.tgz#91b26f941025043194e921d6f6f2914c5bbadaaa"
+  integrity sha512-eZDLKdibB/3htdOBO9Oz7ccffJg4MrIqwqBu1s15fWy18sIDpXdEgGeZRYx6yaOaqNgB2rbC4EsVJOM+LNldbA==
+  dependencies:
+    "@guardian/libs" "^1"
```
mxdvl, line 487: This should really be "^1 || ^2 || ^3". Will raise a PR.

mxdvl (approved):
If you run ~`yarn` again~ `yarn upgrade @guardian/consent-management-platform`, it should work with the latest version of libs :)

PR #405, opened 2021-10-18
feat: Update `youtube-ad-targeting` builder to accept `consentState` parameter

arelra:
Update `youtube-ad-targeting` builder to accept `consentState` parameter rather than call CMP `onConsentChange`


chrislomaxjones (approved):
👍

PR #414, opened 2021-11-02
chore(deps): Bump dependencies 2021.11.1

chrislomaxjones:
## What does this change?

Cherrypick Dependabot PR and run `pnpm up`. 

This PR also also includes a run of `pnpm install web-vitals`. This package is marked as a peer dependency in `@guardian/libs` and not installing `web-vitals` causes files that import this package to give an error.

This PR also avoids cherrypicking the commit from #410 as it causes an error when used in conjunction with [@guardian/eslint-config-typescript](https://github.com/guardian/configs/tree/e51ebc3d7ec185c2f39eff6ecd985f5449dfc872/packages/eslint-config-typescript). This is mostly likely due to the major version change.

## Why?

Keep our dependencies up to date.


zekehuntergreen (approved):

mxdvl:
Do we really need `core-vitals` as a dependency? I think it should be added as a peer dependency.

chrislomaxjones:
> Do we really need `core-vitals` as a dependency? I think it should be added as a peer dependency.

So each package that imports `commercial-core` should have `core-vitals` as a dependency in that case? Is that the behaviour we're after?

mxdvl:
@chrislomaxjones if this is only an issue with Jest, I would actually only include it as a Dev Dependency. You’re right in that we don’t want to force people using commercial-core to have to install `core-vitals`.

PR #415, opened 2021-11-05
feat: Handle Ad Targeting

mxdvl:
## What does this change?

Brings frontend’s `build-page-targeting` to `@guardian/commercial-core`, and fully type it, with extensive JSDoc comment.

Most of these are now pure functions, except for `document.referrer` in `SessionTargeting` and local storage access in `PersonalisedTargeting`.

Subsets of targeting have been created based on their requirements:
- `getAdFreeTargeting`: used only in ad-free contexts.
- `getContentTargeting`: based on CAPI content. Takes an optional `videoLength` parameter.
- `getPersonalisedTargeting`: requires a `ConsentState` to be passed to it.
- `getSessionTargeting`: requires A/B test participations.
- `getViewportTargeting`: requires to know if a privacy banner will be shown.
- `getUnsureTargeting`: these can/should probably be removed.

## Why?

Centralise knowledge, enforce types, link Ad Manager with source code.

IDE completion

## Next steps

Start using these helpers inside frontend

Figure out a way to handle the asynchronous nature of ad targeting. Callbacks, maybe?


package.json
```diff
@@ -78,6 +80,8 @@
     "access": "public"
   },
   "peerDependencies": {
+    "@guardian/ab-core": "^2",
+    "@guardian/consent-management-platform": "^8",
```
mxdvl, line 84: For `Participations` and `ConsentState` types, respectively.

mxdvl:
This has become quite unwieldy, so I’ll simply split it out in separate PRs:
- Content Targeting: https://github.com/guardian/commercial-core/pull/421
- …

PR #419, opened 2021-11-09
chore(deps-dev): Use `@types/googletag`

mxdvl:
## What does this change?

Use `@types/googletag` instead of `@types/doubleclick-gpt`. See guardian/frontend#24349.

## Why?

The types are more restrictive, hence better.

### DIFF
--- a/src/third-party-tags/inizio.ts
+++ b/src/third-party-tags/inizio.ts
@@ -5,11 +5,9 @@ const handleQuerySurveyDone = (
 	survey: { measurementId: string },
 ): void => {
 	if (surveyAvailable) {
-		if (window.googletag) {
-			window.googletag.cmd.push(() => {
-				window.googletag?.pubads().setTargeting('inizio', 't');
-			});
-		}
+		window.googletag.cmd.push(() => {
+			window.googletag.pubads().setTargeting('inizio', 't');
+		});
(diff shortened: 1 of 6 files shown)


src/third-party-tags/inizio.ts
```diff
@@ -5,9 +5,10 @@ const handleQuerySurveyDone = (
 	survey: { measurementId: string },
 ): void => {
 	if (surveyAvailable) {
+		// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- @types/googletag declares it, but it may be missing
```
mxdvl, line 8: This is slightly annoying, but I think it’s worth keeping our types in sync.

chrislomaxjones: I'm curious about this, is it only possibly undefined in this particular case? Because throughout the rest of the source in frontend we removed the `window.googletag?.` optional chaining?

mxdvl: In this case, we aren’t mocking it, whereas in `frontend` it’s always mocked (in tests) or actually imported (in prod).

Here the tests are failing if it’s not defined.

chrislomaxjones: Ah gotcha, that makes sense!

chrislomaxjones (approved):
Looks good!

PR #420, opened 2021-11-09
fix: Fix @guardian/libs to version 3.3.0

chrislomaxjones:
## What does this change?

This PR reverses the version bump of [@guardian/libs](https://github.com/guardian/libs) applied in #414. Instead we fix the version to `3.3.0` for both the development and peer dependency.

## Why?

This resolves an an issue caused when using this package as a dependency of Dotcom Rendering. A runtime error occurred when trying to run the DCR server and was of the form `localStorage is not defined`. This appears to be caused by an attempt to use a browser API on the server.

The version can be returned to `^3` when a more suitable fix for the error is found upstream.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "@commitlint/config-conventional": "^14",
     "@guardian/consent-management-platform": "^8",
     "@guardian/eslint-config-typescript": "^0.7",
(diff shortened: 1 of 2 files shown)


mxdvl (approved):
oh no 😞

PR #421, opened 2021-11-11
feat: Ad Targeting – Content

mxdvl:
## What does this change?

```ts
/**
 * Content Targeting comes from the server
 *
 * For a specific URL, it will only change on
 * - a Composer/CAPI update
 * - a rendering platform capability update
 * - a main media update
 * - a series tag update
 * - a surge in page views per minute
 */
export type ContentTargeting = { /* … */ }
```

### `getContentTargeting`

Returns key-value pairs based on CAPI content. Takes an optional `videoLength` parameter.

[Explore the IDE IntelliSense](https://github.dev/guardian/commercial-core/blob/6c10c1ea1624b9e2c9c67dabc3326a849402531b/src/targeting/content.ts)

## Why?

### DIFF
--- a/src/targeting/content.spec.ts
+++ b/src/targeting/content.spec.ts
@@ -1,0 +1,83 @@
+import type { ContentTargeting } from './content';
+import { getContentTargeting } from './content';
+
+describe('Content Targeting', () => {
+	test('should output the same thing', () => {
+		const content: ContentTargeting = {
+			bl: ['a', 'b'],
+			br: 'f',
+			co: ['Max Duval'],
+			ct: 'article',
+			dcre: 'f',
+			edition: 'uk',
+			k: ['a', 'b'],
+			ob: null,
+			p: 'ng',
+			rp: 'dotcom-platform',
+			s: 'uk-news',
+			se: ['one'],
+			sens: 'f',
+			su: '0',
+			tn: 'something',
+			url: '/some/thing',
+			urlkw: ['a', 'b'],
+			vl: '60',
+		};
+
+		expect(getContentTargeting(content)).toEqual(content);
+	});
+
+	const videoLengths: Array<[number, ContentTargeting['vl']]> = [
+		[10, '30'],
+		[25, '30'],
+		[30, '30'],
+
+		[31, '60'],
+		[59, '60'],
+		[60, '60'],
+
+		[90, '90'],
+		[120, '120'],
+		[150, '150'],
+		[180, '180'],
+		[210, '210'],
+		[240, '240'],
+
+		[300, '300'],
+		[301, '300'],
+		[999, '300'],
+
+		[-999, null],
+		[NaN, null],
+	];
+	test.each(videoLengths)('Video Length (vl) %f => %s', (videoLength, vl) => {
+		const expected: Partial<ContentTargeting> = {
+			vl,
+		};
+
+		const targeting = getContentTargeting(
+			{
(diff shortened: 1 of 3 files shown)


src/targeting/content.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+
+		[-999, null],
+		[NaN, null],
+	];
+	test.each(videoLengths)('Video Length (vl) %f => %s', (videoLength, vl) => {
+		const expected: Partial<ContentTargeting> = {
+			vl,
+		};
+
+		const targeting = getContentTargeting(
```
zekehuntergreen, line 58: Non-blocking: it might be simpler to call `getVideoLength` directly here as that's the functionality being tested.

src/types.ts
```diff
@@ -72,3 +72,8 @@ export type AdsConfigEnabled =
 export type AdsConfig = AdsConfigEnabled | AdsConfigDisabled;
 
 export type AdTargetingBuilder = () => Promise<AdsConfig>;
+
+type True = 't';
+type False = 'f';
+type NotApplicable = 'na';
```
zekehuntergreen, line 78: Where is this used? I think it might Not be Applicable to this PR 😉

zekehuntergreen: Never mind! I see it's used in #422

src/targeting/content.ts
```diff
@@ -0,0 +1,245 @@
...
+	/**
+	 * **U**niform **R**esource **L**ocator - [see on Ad Manager][gam]
+	 *
+	 * Relative to `www.theguardian.com`, starts with `/`
+	 *
+	 * Type: _Dynamic_
+	 *
+	 * [gam]: https://admanager.google.com/59666047#inventory/custom_targeting/detail/custom_key_id=174327
+	 */
+	url: `/${string}`;
```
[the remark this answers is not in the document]
mxdvl: I think they aren’t, but it’s worth double-checking with @Amouzle

src/types.ts
mxdvl: Yeah, I’ve just cherry-picked the same commit in all of them.

src/targeting/content.ts
```diff
@@ -0,0 +1,245 @@
...
+	 * [gam]: https://admanager.google.com/59666047#inventory/custom_targeting/detail/custom_key_id=180207
+	 */
+	p: typeof platform[keyof typeof platform];
+
+	/**
+	 * Rendering Platform - [see on Ad Manager][gam]
+	 *
+	 * [gam]: https://admanager.google.com/59666047#inventory/custom_targeting/detail/custom_key_id=11881005
+	 */
+	rp: 'dotcom-rendering' | 'dotcom-platform';
```
mxdvl, line 157: @Amouzle could this value be embedded inside of `p` (Platform)?

src/targeting/content.spec.ts
mxdvl: I’m a fan of testing the public API. This also ensures that 100% coverages gets trough every branch of your code without doing anything magical.

zekehuntergreen (approved):
Nice job!

src/targeting/content.ts
mxdvl: I guess currently `p` is generated by [`commercial-shared`](https://github.com/guardian/commercial-shared/blob/master/src/main/scala/com/gu/commercial/display/AdTargetParam.scala#L138-L146).

PR #422, opened 2021-11-11
feat: Ad Targeting – Personalised

mxdvl:
## What does this change?

```ts
/**
 * Personalised Targeting requires user consent
 *
 * It allows or prevents personalised advertising, restrict data processing
 * and handles access to cookies and localstorage
 */
export type PersonalisedTargeting = { /* … */ }
```

### `getPersonalisedTargeting`

Returns key-value pairs based on user consent. Will remove existing data in local storage if no consent is given.

[Explore the IDE IntelliSense](https://github.dev/guardian/commercial-core/blob/fc234df1c466f586df6e6f364077bac82903ad74/src/targeting/personalised.ts)

## Why?

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -7,5 +7,8 @@
   },
   "extends": ["@guardian/eslint-config-typescript"],
   "globals": { "googletag": "readonly" },
-  "ignorePatterns": ["dist", "coverage", "src/__vendor"]
+  "ignorePatterns": ["dist", "coverage", "src/__vendor"],
+  "rules": {
+	  "no-use-before-define": ["error", { "functions": true, "classes": true }]
+  }
 }
--- a/src/targeting/personalised.ts
+++ b/src/targeting/personalised.ts
@@ -21,7 +21,6 @@ const frequency = [
 	'16-19',
 	'20-29',
 	'30plus',
-	'5plus', // TODO: remove it (?)
 ] as const;
 
 const AMTGRP_STORAGE_KEY = 'gu.adManagerGroup';
--- a/src/targeting/personalised.ts
+++ b/src/targeting/personalised.ts
@@ -137,7 +137,7 @@ const getFrequencyValue = (
 	state: ConsentState,
 ): PersonalisedTargeting['fr'] => {
 	const rawValue = getRawWithConsent('gu.alreadyVisited', state);
-	if (!rawValue) return '0'; // TODO: should we return `null` instead?
+	if (!rawValue) return '0';
 
 	const visitCount: number = parseInt(rawValue, 10);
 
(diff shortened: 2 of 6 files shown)


src/targeting/personalised.ts
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	const existingGroup = storage.local.getRaw(AMTGRP_STORAGE_KEY);
+
+	return isAdManagerGroup(existingGroup)
+		? existingGroup
+		: createAdManagerGroup();
+};
+
+const createAdManagerGroup = (): AdManagerGroup => {
```
chrislomaxjones, line 231: This is just personal taste but I'd define this function before it's used in `getAdManagerGroup`. If we wanted to enforce that I think we could add the [no-use-before-define](https://eslint.org/docs/rules/no-use-before-define) eslint rule.

src/targeting/personalised.ts
```diff
@@ -0,0 +1,256 @@
...
+	'2',
+	'3',
+	'4',
+	'5',
+	'6-9',
+	'10-15',
+	'16-19',
+	'20-29',
+	'30plus',
+	'5plus', // TODO: remove it (?)
```
chrislomaxjones, line 24: Is the `5plus` a value that's still in GAM? Can we remove it from here if `getFrequencyValue` never returns a value of `5plus`?

mxdvl: https://github.com/guardian/frontend/pull/11297 I can confirm it’s been deprecated since Dec 2015

src/targeting/personalised.ts
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	return null;
+};
+
+const getFrequencyValue = (
+	state: ConsentState,
+): PersonalisedTargeting['fr'] => {
+	const rawValue = getRawWithConsent('gu.alreadyVisited', state);
+	if (!rawValue) return '0'; // TODO: should we return `null` instead?
```
chrislomaxjones, line 141:
```suggestion
	if (!rawValue) return '0';
```
I think personally this is fine as we aren't exporting this function and it's only being used directly in `getPersonalisedTargeting`. And we can always refactor it if necessary later without needing the `TODO: // ...`

chrislomaxjones (approved):
Looks good! The new documentation is really great

.eslintrc.json
```diff
@@ -7,5 +7,8 @@
   },
   "extends": ["@guardian/eslint-config-typescript"],
   "globals": { "googletag": "readonly" },
-  "ignorePatterns": ["dist", "coverage", "src/__vendor"]
+  "ignorePatterns": ["dist", "coverage", "src/__vendor"],
+  "rules": {
+	  "no-use-before-define": ["error", { "functions": true, "classes": true }]
+  }
```
mxdvl, line 13: Thanks @chrislomaxjones for the suggestions

PR #423, opened 2021-11-11
feat: Ad Targeting – Session

mxdvl:
## What does this change?
Adds types around Sessions Targeting, which as per the JSDoc:

```ts
/**
 * Session Targeting is based on the browser session
 *
 * Includes information such as the country of origin, referrer, page view ID.
 *
 * These values identify a browser session are either generated client-side,
 * read from a cookie or passed down from the server.
 */
export type SessionTargeting = { /* … */ }
```

### `getSessionTargeting`

Returns key-value pairs based on the browser session. Requires an object containing server-side and client-side A/B test participations.

[Explore the IDE IntelliSense](https://github.dev/guardian/commercial-core/blob/9023662/src/targeting/session.ts)

## Why?

### DIFF
--- a/src/targeting/session.ts
+++ b/src/targeting/session.ts
@@ -8,7 +8,7 @@ import type { False, True } from '../types';
 const referrers = ['facebook', 'twitter', 'reddit', 'google'] as const;
 
 /**
- * #### Targeting on browser session
+ * Session Targeting is based on the browser session
  *
  * Includes information such as the country of origin, referrer, page view ID.
  *
--- a/src/targeting/session.ts
+++ b/src/targeting/session.ts
@@ -5,7 +5,24 @@ import type { False, True } from '../types';
 
 /* -- Types -- */
 
-const referrers = ['facebook', 'twitter', 'reddit', 'google'] as const;
+const referrers = [
+	{
+		id: 'facebook',
+		match: 'facebook.com',
+	},
+	{
+		id: 'google',
+		match: 'www.google',
+	},
+	{
+		id: 'twitter',
+		match: '/t.co/',
+	},
+	{
+		id: 'reddit',
+		match: 'reddit.com',
+	},
+] as const;
 
 /**
  * Session Targeting is based on the browser session
@@ -77,7 +94,7 @@ export type SessionTargeting = {
 	 *
 	 * [gam]: https://admanager.google.com/59666047#inventory/custom_targeting/detail/custom_key_id=228567
 	 */
-	ref: typeof referrers[number] | null;
+	ref: typeof referrers[number]['id'] | null;
 
 	/**
 	 * **S**igned **I**n – [see on Ad Manager][gam]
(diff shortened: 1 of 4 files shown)


src/targeting/session.ts
```diff
@@ -0,0 +1,173 @@
...
+	/**
+	 * **AB** Tests – [see on Ad Manager][gam]
+	 *
+	 * Type: _Dynamic_
+	 *
+	 * Values: typically start with `ab`
+	 *
+	 * [gam]: https://admanager.google.com/59666047#inventory/custom_targeting/detail/custom_key_id=186327
+	 */
+	ab: string[] | null;
```
zekehuntergreen, line 28: Is the `| null` important here? Or could we change this and the return type of `experimentsTargeting` to `string[]`?

src/targeting/session.ts
```diff
@@ -0,0 +1,173 @@
+import type { Participations } from '@guardian/ab-core';
+import type { CountryCode } from '@guardian/libs';
+import { isString } from '@guardian/libs';
+import type { False, True } from '../types';
+
+/* -- Types -- */
+
+const referrers = ['facebook', 'twitter', 'reddit', 'google'] as const;
```
zekehuntergreen, line 8: Do we want to add `googleplus` here and in referrerTypes? It does seem to be one of the possible targeting values in GAM.

zekehuntergreen: It might be neater to do something [like this](https://www.typescriptlang.org/play?#code/MYewdgzgLgBATgUwGYLouAVAngBwRAYXGhgF4YBtAKBlpgG8a7mBLAEwC4YByJAQ2AIARiBABrbgBomzWgFs+UYAAsuvAcNFiAdKDlSZMAL7TmjWXXZqA5qOsAbBAYvzFKtQHcv22yAdPTOhNDcxcrHigPFigoVGcXBSVVHgB6KF0QFPjaYLNDVk4eRDY2aOzZRPcihBLojP1AnOkAXRg+CBhQSCgqKFwEeGQUdGw8Mhg+vBAkQeGR-sJiHsmBxDnUcZXp2dR5vEXuijAAVzkhVGaKACJ2K+agA) than to have these 4-5 string defined here and in  `referrerTypes` as well?
(Yes I'm plagiarising your comment from https://github.com/guardian/frontend/pull/24370 😄 )

src/targeting/session.ts
mxdvl: I think it’s nicer to have `null` than an empty array `[]`.
This means we can filter out nulls more easily in the future.

src/targeting/session.ts
mxdvl: According to Wikipedia: [Google+ is defunct](https://en.wikipedia.org/wiki/Google%2B)

> ### Current status:
> **Defunct**: Discontinued for personal and brand accounts (April 2, 2019; 2 years ago)All users transitioned to Google Currents (G-Suite enterprise accounts)

zekehuntergreen: I didn't realize! In that case can we remove it from GAM and the comment on line 91?

zekehuntergreen (approved):

PR #424, opened 2021-11-11
feat: Ad Targeting – Viewport

mxdvl:
## What does this change?

```ts
/**
 * Viewport Targeting
 *
 * Includes values related to the viewport:
 * - breakpoint (deprecated?)
 * - whether a CMP banner will show
 * - size of page skin
 */
export type ViewportTargeting = { /* … */ }
```

### `getViewportTargeting`

Returns key-value pairs based on the size of the viewport and whether a privacy banner will show.

[Explore the IDE IntelliSense](https://github.dev/guardian/commercial-core/blob/621d956e27fcdfbeed9b9f8b2ca43da75c69d2da/src/targeting/viewport.ts)

## Why?

### DIFF
--- a/src/targeting/viewport.spec.ts
+++ b/src/targeting/viewport.spec.ts
@@ -1,0 +1,64 @@
+import type { ViewportTargeting } from './viewport';
+import { getViewportTargeting } from './viewport';
+
+describe('Viewport targeting', () => {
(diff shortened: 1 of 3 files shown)


arelra (approved):
👍

PR #425, opened 2021-11-12
feat: Ad Targeting – Unsure

mxdvl:
## What does this change?
Adds types around Targeting which we’re not sure about, which as per the JSDoc:

```ts
/**
 * I think all of these are deprecated?!?
 */
export type UnsureTargeting = { /* … */ }
```

### `getUnsureTargeting`

You get back what you pass to it, but I think it should all go, go, go away!

## Why?

### DIFF
--- a/src/targeting/unsure.ts
+++ b/src/targeting/unsure.ts
@@ -1,5 +1,8 @@
 /* -- Types -- */
 
+/**
+ * I think all of these are deprecated?!?
+ */
 export type UnsureTargeting = {
 	/**
 	 * **G**uar**d**ia**n** **C**ustomer **R**elation **M**anagement – [see on Ad Manager][gam]
@@ -11,6 +14,8 @@ export type UnsureTargeting = {
 	 * - `"PXHCI2"`
 	 *
 	 * [gam]: https://admanager.google.com/59666047#inventory/custom_targeting/detail/custom_key_id=184047
+	 *
+	 * @deprecated (?)
 	 */
 	gdncrm: string | string[];
 
@@ -25,6 +30,8 @@ export type UnsureTargeting = {
 	 * - `"a-rational-fear"`
 	 *
 	 * [gam]: https://admanager.google.com/59666047#inventory/custom_targeting/detail/custom_key_id=186807
+	 *
+	 * @deprecated - We no longer use Video.js (?)
 	 */
 	ms: string;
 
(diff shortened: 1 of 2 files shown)


src/targeting/unsure.ts
```diff
@@ -0,0 +1,67 @@
...
+	/**
+	 * Kru**x** user segments – [see on Ad Manager][gam]
+	 *
+	 * * Type: _Dynamic_
+	 *
+	 * [gam]: https://admanager.google.com/59666047#inventory/custom_targeting/detail/custom_key_id=200247
+	 *
+	 * @deprecated We no longer use Krux
+	 */
+	x: string;
```
mxdvl, line 58: @Amouzle confirmed this can be removed

src/targeting/unsure.ts
```diff
@@ -0,0 +1,56 @@
...
+	 */
+	slot: string; // TODO: Narrow to known IDs
+};
+
+/* -- Methods -- */
+
+/* -- Targeting -- */
+
+export const getUnsureTargeting = (
+	targeting: UnsureTargeting,
```
zekehuntergreen, line 55: Have we thought about creating a log when any of these properties are used in order to confirm whether or not they're deprecated?

mxdvl: Yes: https://docs.google.com/spreadsheets/d/10vsny8P_hdtBSNBp3DSRpRV49e_mVJ24Q43NY4askeY/

Amouzle: Happy with removing `gdrm` & `ms`. Just checking that `slot` will only be removed from the [BuildPageTargeting](https://github.com/guardian/frontend/blob/8bf15042f04c15875afbba5c5f93fcfa979e4357/static/src/javascripts/projects/common/modules/commercial/build-page-targeting.ts#L289)?

mxdvl: Yes, only from the page targeting, not for individual slots.

mxdvl:
Closing this in favour of the deprecating these values in frontend:
- `gdncrm` https://github.com/guardian/frontend/pull/24424
- `ms` TBA
- `slot` TBA

src/targeting/unsure.ts
mxdvl: https://github.com/guardian/frontend/pull/24431

PR #436, opened 2021-11-30
chore(deps): manually update dependencies 30/11/2021

simonbyford:
## What does this change?

- bumps dependencies manually by running `pnpm update --latest`
- fixes some eslint violations
- updates `actions/cache` (cherry-picked from https://github.com/guardian/commercial-core/pull/433)

## Why?

Dependabot doesn't support `pnpm` which means most of our dependabot PRs have been broken [since we introduced `pnpm` on 18th October](https://github.com/guardian/commercial-core/pull/404). They are failing with the error:

```
 ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up-to-date with package.json
```

This is a known issue: https://github.com/dependabot/dependabot-core/issues/1736

To temporarily resolve this, I closed a bunch of affected PRs (see below) and ran `pnpm update --latest` manually. This bumped all the relevant dependencies in one go.

Open questions:
- is there a workaround to get `pnpm` working with dependabot? 
- if not, is it worth ditching either dependabot or `pnpm` on this repo?
- what would the main drawbacks of simply running `pnpm update --latest` every week instead of relying on the usual dependabot workflow? Off the top of my head: having individual PRs for each dependency gives us more control and makes issues easier to isolate


mxdvl (approved):
If this works, it’s great.

If we cannot get @dependabot to play ball, we might have to drop `pnpm` and go back to `yarn`.

PR #439, opened 2021-12-03
docs: Restructure docs folder

frederickobrien:
This rearranges existing documentation and supporting assets to fit with a more long-term structure. Rather than having a flat docs folder with entries pulling through images from an assets subfolder, every topic now has its own folder in which all files relevant to that topic are stored.


mxdvl (approved):
Nice and tidy. 🗄️

PR #444, opened 2021-12-06
feat: `EventTimer.mark` and `.trackInGA` are private methods

zekehuntergreen:
## What does this change?
Makes `trigger` and `trackInGA` private methods of `EventTimer`
Note: contrary to the PR title, this isn't a really feature but needs to be labeled as such in order to trigger a minor release when merged.

## Why?
- Consumers should use `EventTimer.trigger` rather than `.mark` so that we 
   -  track events with Google Analytics
   - de-dup events (?)

We can enforce the use of `trigger` by making the other two methods private.

### DIFF
--- a/src/EventTimer.ts
+++ b/src/EventTimer.ts
@@ -168,7 +168,7 @@ export class EventTimer {
 				: {};
 	}
 
-	mark(name: string): void {
+	private mark(name: string): void {
 		const longName = `gu.commercial.${name}`;
 		if (
 			typeof window.performance !== 'undefined' &&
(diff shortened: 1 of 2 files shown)


src/EventTimer.ts
```diff
@@ -228,7 +228,7 @@ export class EventTimer {
 		}
 	}
 
-	trackInGA(eventName: string, label = ''): void {
+	private trackInGA(eventName: string, label = ''): void {
```
arelra, line 231: This is super minor and feel free to ignore:

Can we list the private & public methods together?

arelra (approved):

src/EventTimer.ts
zekehuntergreen: Good call

PR #445, opened 2021-12-08
feat: Ad Targeting – Shared

mxdvl:
## What does this change?

Clarifies the distinction between _content_ and _shared_ targeting. Exports all targeting from the top level.

Shared targeting comes from [guardian/commercial-shared/…/`AdTargetParam.scala`](https://github.com/guardian/commercial-shared/blob/a692e8b2eba6e79eeeb666e5594f2193663f6514/src/main/scala/com/gu/commercial/display/AdTargetParam.scala)

## Why?

We’re not generating these values client-side.


arelra (approved):
👍

PR #446, opened 2021-12-08
feat: Pick defined values from object

mxdvl:
## What does this change?

Introduces a type-safe object picker, inspired by [`removeFlaseyValues`][] in Frontend.

[`removeFlaseyValues`]: https://github.com/guardian/frontend/blob/035b8ff00fdd2740e33d4d1ab65da5fb0bb78deb/static/src/javascripts/projects/commercial/modules/header-bidding/utils.ts#L48

## Why?

Code Golfing is fun 🏌️

### DIFF
--- a/src/lib/pick-defined-values.ts
+++ b/src/lib/pick-defined-values.ts
@@ -1,0 +1,42 @@
+import { isString } from '@guardian/libs';
+
+type ValidTargeting<T, K> = '' extends T
+	? never
+	: [] extends T
+	? never
+	: [''] extends T
+	? never
+	: T extends boolean
+	? never
+	: T extends NonNullable<T>
+	? K
+	: never;
+type DefinedKeys<T> = { [K in keyof T]-?: ValidTargeting<T[K], K> }[keyof T];
+type ObjectWithDefinedValues<T> = Pick<T, DefinedKeys<T>>;
+
+const isValidTargeting = (value: unknown): value is string | string[] => {
+	if (isString(value) && value !== '') return true;
+	if (
+		Array.isArray(value) &&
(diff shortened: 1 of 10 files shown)


src/lib/pick-defined-values.ts
```diff
@@ -0,0 +1,42 @@
...
+	? never
+	: [''] extends T
+	? never
+	: T extends boolean
+	? never
+	: T extends NonNullable<T>
+	? K
+	: never;
+type DefinedKeys<T> = { [K in keyof T]-?: ValidTargeting<T[K], K> }[keyof T];
+type ObjectWithDefinedValues<T> = Pick<T, DefinedKeys<T>>;
```
arelra: Wow 😎 

5x3bzr

src/lib/pick-defined-values.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const isValidTargeting = (value: unknown): value is string | string[] => {
+	if (isString(value) && value !== '') return true;
+	if (
+		Array.isArray(value) &&
+		value.filter(isString).filter(Boolean).length > 0
+	)
+		return true;
```
arelra: 'Dangling' returns scare me a bit.

Also can we improve the readability a bit and create named variables for the boolean conditions?

src/lib/pick-defined-values.ts
mxdvl: 🤣  Yeah I explored these ideas in 10% time… it’s definitely right on the edge of usefulness.

ashishpuliyel: wow is right.

ashishpuliyel (approved):

PR #448, opened 2021-12-09
fix(types): Pick targeting keys + use `type-fest`

mxdvl:
## What does this change?

- Fix an issue with arrays of valid and invalid strings
- Lean `type-fest` for the funny type
- Update tests to cover more explicitly what the type does

## Why?

Easier to read!


arelra (approved):
Nice :)

PR #450, opened 2021-12-10
fix: A little more type safety

mxdvl:
## What does this change?

Disallow `never[]`

## Why?

If you don’t use the `as const` assertion, you still get quite a lot of type safety.

Follow up on #xxx


simonbyford (approved):
This helped me learn about `as const`

PR #455, opened 2021-12-16
feat: unify getSessionTargeting API

mxdvl:
## What does this change?

Unify the API to match other `get*Targeting` methods

## Why?

Consistency!


simonbyford (approved):
Makes sense 👌

PR #462, opened 2021-12-21
feat: Add `getTuple` helper function to `ad-sizes.ts`

frederickobrien:
As outlined in this [ticket](https://trello.com/c/nakVaVsO/1659-add-gettuple-method-to-commercial-core-ad-sizes), now that ad sizes have moved over to commercial core it makes sense for some helper functions to join them. This adds a `getTuple` function and `AdSizeTuple` type to `ad-sizes.ts`.

### DIFF
--- a/src/ad-sizes.test.ts
+++ b/src/ad-sizes.test.ts
@@ -1,4 +1,4 @@
-import { _ } from './ad-sizes';
+import { _, getTuple } from './ad-sizes';
 
 const { getAdSize } = _;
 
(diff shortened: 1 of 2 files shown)


mxdvl (commented):
That’s going to be a useful addition, @frederickobrien!

Could you write some tests for this? Jest’s `test.each` might come in handy to test several values at once. At the very least we’d check the MPU size works!

frederickobrien:
Like this, @mxdvl?

src/ad-sizes.test.ts
```diff
@@ -19,3 +19,20 @@ describe('ad sizes', () => {
...
+
+describe('getTuple', () => {
+	it.each([
+		['mpu' as const, [300, 250]],
+		['fluid' as const, [0, 0]],
+		['googleCard' as const, [300, 274]],
+		['outstreamGoogleDesktop' as const, [550, 310]],
+		['video' as const, [620, 1]],
+		['300x600' as const, [300, 600]],
+	])(
```
mxdvl, line 31: Not sure the `as const` assertion is doing anything for you here. You need to have a mutable object like an array of object to get more type safety, like so:

```ts
const sizes = [
		['mpu', [300, 250]],
		['fluid', [0, 0]],
		['googleCard', [300, 274]],
		['outstreamGoogleDesktop', [550, 310]],
		['video', [620, 1]],
		['300x600', [300, 600]],
	] as const;

it.each(sizes)(/* … */);
```

But personally I’d go for something like, so you get type safety as you build your object: 
```ts
const sizes: Array<[SizeKeys, AdSizeTupe]> = [
		['mpu', [300, 250]],
		['fluid', [0, 0]],
		/* etc… */
]
```

mxdvl (approved):
Great.

Maybe we can add a `toTuple` method to the `AdSize` type, as well? Let’s see how it’s actually used in our codebase first.

src/ad-sizes.test.ts
frederickobrien: Ah thank you, that's much better!

PR #463, opened 2021-12-22
feat: Replace `getTuple` with `getAdSize`

frederickobrien:
Refactor previous PR by replacing `getTuple` with `getAdSize` and changing the old `getAdSize` to a more accurate `createAdSize`. The tuples are no more. Also fix exports so imports from the @guardian/commercial-core package works properly.

This was the product of a mobbing session with @ashishpuliyel, @arelra, and @simonbyford in which we questioned what a tuple ( `[width: number, height: number]` ) offered which the `AdSize` object below doesn't: 

```
{
        width: number;
	height: number;
	toString: () => AdSizeString;
}
``` 

Both contained width and height values, the latter just has a handy custom `toString` function. This seemed like a good opportunity to streamline the code, and will be followed up on with a corresponding PR in Frontend.

### DIFF
--- a/src/ad-sizes.test.ts
+++ b/src/ad-sizes.test.ts
@@ -1,8 +1,7 @@
-import type { AdSizeTuple } from './ad-sizes';
-import { _, getTuple } from './ad-sizes';
+import { _, getAdSize } from './ad-sizes';
 import type { SizeKeys } from '.';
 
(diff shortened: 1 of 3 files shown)


ashishpuliyel (approved):
This matches what we discussed

PR #471, opened 2022-01-04
fix(EventTimer): private constructor and rely on `performance.mark()`’s return

mxdvl:
### 1. `private constructor()`
Make the constructor private, to prevent people from doing `new EventTimer()` instead of `EventTimer.get()`.

### 2. `performance.mark()` returns a `mark`
Rely on the fact `performance.mark` returns the mark it created, so use that instead of going through all the marks and getting the last one. Fixed the tests so the mocks more closely match browser implementation.

### 3. `filter().map()` &rarr; `map().filter()`
The filtering of events is flipped so as to get rid of a type assertion. For `_externallyDefinedEventNames`, we now check if an entry exists, return an `Event` from it if id does, and return `undefined` otherwise. Then filter the resulting array for entries that are instances of `Event`.

## Why?

Cleaner, probably faster, uses returns types from the API.

### DIFF
--- a/src/EventTimer.spec.ts
+++ b/src/EventTimer.spec.ts
@@ -19,20 +19,21 @@ const mockGetEntriesByName = (names: string[]) =>
 			: [],
 	);
 
+const mockMark = jest.fn(
+	(name: string): PerformanceMark => ({
+		name,
+		duration: 1,
+		entryType: 'mark',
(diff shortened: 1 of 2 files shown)


ashishpuliyel (approved):
Looks fine to me

PR #475, opened 2022-01-05
feat: Order ad size info alphabetically

frederickobrien:
Done while pairing with @mxdvl looking at `AdSize` types and tuples, just feels a bit easier to read.

### DIFF
--- a/src/ad-sizes.ts
+++ b/src/ad-sizes.ts
@@ -7,32 +7,32 @@ export type AdSize = Readonly<{
 }>;
 
 export type SizeKeys =
+	| '160x600'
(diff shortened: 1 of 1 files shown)


chrislomaxjones (approved):
Nice!

As an aside we could enforce this behaviour with eslint using https://eslint.org/docs/rules/sort-keys.

chrislomaxjones:
@frederickobrien You'll also want to prefix your first commit with something like `chore:` to get the CI to pass.

frederickobrien:
Smashing, thanks @chrislomaxjones!

PR #481, opened 2022-01-20
refactor!: commercialMetrics

lucymonie:
## What does this change?
This changes the interface of sendCommercialMetrics, so instead of having a single `sendCommercialMetrics` function export, it now has `initCommercialMetrics`, and `bypassCommercialMetricsSampling`, which allows us to record data that's outside the sampling rate.
 
## Why?
This follows the clearer patterns established in the [coreWebVitals API](https://github.com/guardian/libs/pull/280). 

Link to [trello card](https://trello.com/c/V4S87Q9y/1557-integrate-corewebvitals-updates-to-commercialmetrics)


src/sendCommercialMetrics.ts
```diff
@@ -1,72 +1,186 @@
...
+type Metric = {
 	name: string;
 	value: number;
 };
 
-type Properties = {
+type Property = {
 	name: string;
-	value: string;
+	value: string | number;
```
mxdvl, line 11: I’d be tempted to keep this a `string` as that’s what the backend expects.

src/sendCommercialMetrics.ts
```diff
@@ -1,72 +1,186 @@
...
+	downlink?: number;
+	effectiveType?: string;
+};
+
+type CommercialMetricsPayload = {
+	page_view_id: string | null;
+	browser_id?: string | null;
+	platform: 'NEXT_GEN' | null;
+	metrics: readonly Metric[] | null;
+	properties: readonly Property[] | null;
```
mxdvl, line 30: Maybe introducing `null` here is just confusing. Either it’s `undefined` or it’s the expected type?

```suggestion
	page_view_id?: string;
	browser_id?: string;
	platform?: 'NEXT_GEN';
	metrics?: readonly Metric[];
	properties?: readonly Property[];
```

src/sendCommercialMetrics.ts
```diff
@@ -1,72 +1,186 @@
...
-			  ]
-			: [];
-
-	const endpoint = isDev
-		? '//performance-events.code.dev-guardianapis.com/commercial-metrics'
-		: '//performance-events.guardianapis.com/commercial-metrics';
-	if (document.visibilityState !== 'hidden') return false;
+	if (document.visibilityState !== 'hidden') {
+		return false;
+	}
```
mxdvl, line 134: This should be wrapped inside a `visibilitychange` event listener.

src/sendCommercialMetrics.ts
```diff
@@ -1,72 +1,186 @@
...
+};
+
+const mapEventTimerPropertiesToString = (
+	properties: Array<[string, string | number]>,
+): Property[] => {
+	return properties.map(([name, value]) => ({
+		name: String(name),
+		value: String(value),
+	}));
+};
```
mxdvl, line 83: We already have an object cleaner in ad-targeting:  https://github.com/guardian/commercial-core/blob/main/src/targeting/pick-targeting-values.ts.

If this is something we end up doing a lot of, maybe we’d want to have a general utility to remove undefined keys from an object.

mxdvl (commented):
Nice work Lucy!

I think there’s only a few bits missing:
- the PR title should start with `refactor!: ` (with a `!`), because it will introduce a breaking change
- getting the metrics should be done in an event listener at page unload. This will be `visibilitychange` with a fallback on `pagehide` for Safari.
- I would be tempted to have a single function that returns the `eventTimer` properties: one that takes no arguments and handles `EventTimer.get().properties` internally. This would make it clearer that the function is _impure_.

src/sendCommercialMetrics.ts
lucymonie: Yes, the value is a string either way so this is an unnecessary alternative 👍🏻

src/sendCommercialMetrics.ts
lucymonie: Thank you - good suggestion 👍🏻

src/sendCommercialMetrics.ts
lucymonie: 👍🏻

src/sendCommercialMetrics.ts
lucymonie: Yes, that's a good point. Do you think it's at that point now? Or just something to keep in mind?

mxdvl: I think just keep in mind at this point 😁

src/sendCommercialMetrics.ts
```diff
@@ -1,72 +1,186 @@
...
+				},
+		  ]
+		: [];
+};
+
+const filterUndefinedEventTimerProperties = (
+	eventTimerProperties: EventProperties,
+): Array<[string, string | number]> => {
+	// Transforms object {key: value} pairs into an array of [key, value] arrays
+	return Object.entries(eventTimerProperties).filter(
```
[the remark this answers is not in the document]
lucymonie: Sure thing - shall do.

src/sendCommercialMetrics.ts
```diff
@@ -107,15 +123,49 @@ function sendMetrics(isDev: boolean) {
...
 	if (!initialised) {
 		console.warn('initCommercialMetrics not yet initialised');
 		return;
 	}
 
-	sendMetrics(isDev);
+	sendMetrics();
 }
 
+export function gatherMetricsOnPageUnload(): void {
```
mxdvl, line 117: Is this only exported for testing? If so, it would be preferable to make it part of the `_` export.

src/sendCommercialMetrics.ts
```diff
@@ -23,54 +23,71 @@ type EventProperties = {
...
 
 enum Endpoints {
 	CODE = '//performance-events.code.dev-guardianapis.com/commercial-metrics',
 	PROD = '//performance-events.guardianapis.com/commercial-metrics',
 }
 
-const commercialMetricsPayload: CommercialMetricsPayload = {
-	page_view_id: null,
-	browser_id: null,
+let commercialMetricsPayload: CommercialMetricsPayload = {
```
mxdvl, line 38: A `const` can still have its keys mutated. You might be able to keep it as such. 

Not a blocker, but maybe we can omit the `metrics` and `properties` keys here, as they will only ever be populated once and have the metrics be a parameter of `sendMetrics`? In coreWebVitals, we were constrained by the inner workings of Google’s `web-vitals` package, but here we can strive for a pure approach.

mxdvl (approved):
LGTM! 🎉 The API is quite clear and should make it easier to use in `frontend` and `dotcom-rendering`.

Left a comment on whether we can use a more functional approach, but I don’t think it’s worth fussing over too much–it’s more important to gather proper metrics and have a consistent API.

src/sendCommercialMetrics.ts
lucymonie: Actually, I don't think it needs exporting at all.

src/sendCommercialMetrics.ts
lucymonie: There's an update to the whole object in the reset, so I switched it to let here.

PR #495, opened 2022-02-02
feat: Export ad targeting and adjust `getViewportTargeting` arguments

chrislomaxjones:
## What does this change?

Export functions and types related to ad targeting, introduced in feature branches
- https://github.com/guardian/commercial-core/pull/421
- https://github.com/guardian/commercial-core/pull/422
- https://github.com/guardian/commercial-core/pull/423
- https://github.com/guardian/commercial-core/pull/424
- https://github.com/guardian/commercial-core/pull/455

This will replace the ad targeting implementations in Frontend and DCR (where targeting on AMP is different to that in the commercial bundle), after the changes in [Frontend #24498](https://github.com/guardian/frontend/pull/24498).

Secondly, this PR includes a small change to the type of `getViewportTargeting`, making it take an object rather than two arguments. This is so that it has a similar interface to the other targeting functions (note personalised does not have this interface as it only takes consent state as its sole argument).

## Why?

So that we can access these types/functions in Frontend/DCR from the root of the package rather than a long path, for example

```typescript
import { getSessionTargeting } from '@guardian/commercial-core'
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -26,3 +26,14 @@ export type {
 	CustomParams,
 } from './types';
 export * as constants from './constants';
+export type { ContentTargeting } from './targeting/content';
+export { getContentTargeting } from './targeting/content';
+export type { PersonalisedTargeting } from './targeting/personalised';
+export { getPersonalisedTargeting } from './targeting/personalised';
+export type { SessionTargeting } from './targeting/session';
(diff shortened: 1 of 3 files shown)


arelra (approved):

lucymonie (approved):
Great explanation - very newbie-friendly 👍🏻

PR #497, opened 2022-02-07
fix(targeting): Change minimum tablet width

chrislomaxjones:
## What does this change?

Adjust the minimum width of the viewport for `findBreakpoint` to output `tablet` from `740` (the width of the `tablet` breakpoint) to `660` (the width of the `phablet` breakpoint).

Update the tests to reflect this change.

## Why?

To match the current implementation in [Frontend](https://github.com/guardian/frontend/blob/main/static/src/javascripts/projects/common/modules/commercial/build-page-targeting.ts#L113-L128), where `phablet` width `660` also maps to `tablet` when generating page targeting.


lucymonie (approved):
LGTM 👍🏻

PR #498, opened 2022-02-08
feat: Update event timer names

zekehuntergreen:
## What does this change?
Makes changes 2 and 3 from the diagram below:
- Renames the commercial timer event `commercialEnd` to `commercialModulesLoaded`
- Adds two commercial metrics, `commercialExtraModulesLoaded` and `commercialBaseModulesLoaded`

Screenshot 2022-02-08 at 17 36 59

## Why?
- The change from `commercialEnd` to `commercialModulesLoaded` reflects what has actually occurred when the metric is triggered; All commercial modules have been loaded (but haven't necessarily finished executing), but since more commercial events (e.g. initialising the first slot and showing the first ad) have yet to happen, `commercialEnd` could be misleading.
- `commercialExtraModulesLoaded` and `commercialBaseModulesLoaded` are [already being triggered](https://github.com/guardian/frontend/blob/fed5c0cbf116e432d1bdb58d64ce7d8ad904b386/static/src/javascripts/bootstraps/standalone.commercial.ts#L170) so their addition here is just to keep commercial-core up to date with frontend

### DIFF
--- a/src/EventTimer.ts
+++ b/src/EventTimer.ts
@@ -28,6 +28,8 @@ interface SlotEventStatus {
 
 interface PageEventStatus {
 	commercialStart: boolean;
+	commercialExtraModulesLoaded: boolean;
+	commercialBaseModulesLoaded: boolean;
 	commercialModulesLoaded: boolean;
 }
 
@@ -124,6 +126,8 @@ export class EventTimer {
 			},
 			page: {
 				commercialStart: false,
+				commercialExtraModulesLoaded: false,
+				commercialBaseModulesLoaded: false,
 				commercialModulesLoaded: false,
 			},
 		};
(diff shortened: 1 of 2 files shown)


src/EventTimer.ts
```diff
@@ -141,7 +145,7 @@ export class EventTimer {
 					timingLabel: 'Commercial start parse time',
 				},
 				{
-					timingVariable: 'commercialEnd',
+					timingVariable: 'commercialModulesLoaded',
```
zekehuntergreen, line 148: What is the impact of changing the name of the `timingVar` we pass to `window.ga`?

zekehuntergreen: There's no harm in renaming the label we send to Google Analytics.

arelra (approved):

PR #507, opened 2022-02-15
feat!: Add properties `numInlineSlots` and `totalSlots` to commercial metrics

zekehuntergreen:
## What does this change?
- introduce two new commercial metrics
   - `numInlineSlots` - the number of inline ad slots on the page
   - `totalSlots` - the total number of ad slots on the page
- use parameter destructuring to simplify `initCommercialMetrics` api

## Why?
- It's necessary to collect `numInlineSlots` and `totalSlots` properties for the analysis of future A/B test around lazily-loaded ads on articles.
- adding the two new parameters to `initCommercialMetrics` brought the total arguments a caller could use to 7. Leaving these as positional arguments made the function more difficult to use.

### DIFF
--- a/src/sendCommercialMetrics.spec.ts
+++ b/src/sendCommercialMetrics.spec.ts
@@ -37,13 +37,13 @@ const defaultMetrics = {
 };
 
 const mockSendMetrics = () =>
-	initCommercialMetrics(
-		PAGE_VIEW_ID,
-		BROWSER_ID,
-		IS_NOT_DEV,
-		ADBLOCK_NOT_IN_USE,
-		USER_IN_SAMPLING,
-	);
+	initCommercialMetrics({
+		pageViewId: PAGE_VIEW_ID,
+		browserId: BROWSER_ID,
+		isDev: IS_NOT_DEV,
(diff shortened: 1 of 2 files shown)


ashishpuliyel (approved):
It looks fine to me.
I know naming things is a never ending pain. But for two metrics that are effectively parallel, i wonder if more parallel names would be better. For example `slotsTotal` and `slotsInline` instead?

ashishpuliyel:
Lo

> It looks fine to me. I know naming things is a never ending pain. But for two metrics that are effectively parallel, i wonder if more parallel names would be better. For example `slotsTotal` and `slotsInline` instead?

Or even `adSlotsTotal` and `adSlotsInline` to make it clear what kind of slots we're talking about if someone just sees the key name in the data somewhere.

lucymonie (approved):
This looks good! I completely agree that the api of `initCommercialMetrics` was getting unwieldy 👍🏻

zekehuntergreen:
> Or even `adSlotsTotal` and `adSlotsInline` to make it clear what kind of slots we're talking about if someone just sees the key name in the data somewhere.

Good idea - those names are much clearer

PR #508, opened 2022-02-17
feat!: Move slot metrics to EventTimer properties

zekehuntergreen:
## What does this change?
Before:
Since https://github.com/guardian/commercial-core/pull/507 `initCommercialMetrics` takes two new arguments - `adSlotsInline` and `adSlotsTotal`

After: 
These parameters are removed. `adSlotsInline` and `adSlotsTotal` can instead be added to the event timer via `EventTimer. setProperty`

## Why?
As of https://github.com/guardian/commercial-core/pull/481, `initCommercialMetrics` rather than `sendCommercialMetrics` is the interface with frontend and DCR. At the point when fronted calls `initCommercialMetrics` the number of ad slots on the page is not yet known. `EventTimer. setProperty` allows us to collect extra properties after the commercial modules have initialised.

### DIFF
--- a/src/sendCommercialMetrics.spec.ts
+++ b/src/sendCommercialMetrics.spec.ts
@@ -255,7 +255,7 @@ describe('send commercial metrics code', () => {
 			]);
 		});
 
-		it('should accept ad slot properties', () => {
+		it('should handle ad slot properties', () => {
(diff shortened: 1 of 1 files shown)


chrislomaxjones (approved):
+1, I think this method of setting slot metrics makes a lot of sense.

PR #514, opened 2022-03-02
feat: Collect commercial metrics on scroll depth

zekehuntergreen:
## What does this change?
Adds a function `initTrackScrollDepth` which collects commercial metrics on user scroll depth. It does this by inserting hidden elements at intervals of 1 viewport height, then using an intersection observer to mark the time when the viewport intersects with these elements with `EventTimer.trigger`.

## Why?
We'd like to be more accurate in finding the optimal moment to load lazily-loaded ads and suspect that there may be a possible solution that considers scroll velocity. We're temporarily collecting scroll depth metrics in order as background research to help us with this.

## Known issues
- There will be some noise from inaccurate scroll depth metrics due to users resizing their window after the page loads

### DIFF
--- a/src/EventTimer.ts
+++ b/src/EventTimer.ts
@@ -39,6 +39,8 @@ interface EventTimerProperties {
 	effectiveType?: string;
 	adSlotsInline?: number;
 	adSlotsTotal?: number;
+	// the height of the page / the viewport height
+	pageHeightVH?: number;
 }
 
 export class EventTimer {
@@ -173,7 +175,16 @@ export class EventTimer {
 				: {};
 	}
 
-	setProperty(name: 'adSlotsInline' | 'adSlotsTotal', value: number): void {
+	/**
+	 * Adds an event timer property
+	 *
+	 * @param {string} name - the property's name
+	 * @param {value} number - the property's value
+	 */
+	setProperty(
+		name: 'adSlotsInline' | 'adSlotsTotal' | 'pageHeightVH',
+		value: number,
+	): void {
 		this.properties[name] = value;
 	}
 
--- a/src/track-scroll-depth.spec.ts
+++ b/src/track-scroll-depth.spec.ts
@@ -1,0 +1,61 @@
+import { EventTimer } from './EventTimer';
+import { initTrackScrollDepth } from './track-scroll-depth';
+
+describe('initTrackScrollDepth', () => {
+	const observe = jest.fn();
+	class IntersectionObserver {
+		constructor() {
+			return Object.freeze({
+				observe: observe,
+				unobserve: () => {
+					return;
+				},
+			});
+		}
+	}
+
+	const originalOffsetHeight = Object.getOwnPropertyDescriptor(
+		HTMLElement.prototype,
+		'offsetHeight',
+	);
+	beforeAll(() => {
+		Object.defineProperty(HTMLElement.prototype, 'offsetHeight', {
+			configurable: true,
+			value: 1000,
+		});
+
+		Object.defineProperty(global, 'IntersectionObserver', {
+			value: IntersectionObserver,
+			writable: true,
+		});
+	});
+
+	afterAll(() => {
+		Object.defineProperty(
+			HTMLElement.prototype,
+			'offsetHeight',
+			originalOffsetHeight ?? 0,
+		);
+	});
+
+	test('Hidden elements are inserted and observed', () => {
+		// Given a 1000px page and a viewport height of 100px,
+		// 10 hidden elements are inserted
+		// NOTE we can't simulate scrolling down the page and assert that the callback
+		// to IntersectionObserver works as expected.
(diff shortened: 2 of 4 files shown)


src/EventTimer.ts
```diff
@@ -173,7 +175,16 @@ export class EventTimer {
...
 
-	setProperty(name: 'adSlotsInline' | 'adSlotsTotal', value: number): void {
+	/**
+	 * Adds an event timer property
+	 *
+	 * @param {string} name - the property's name
+	 * @param {value} number - the property's value
+	 */
+	setProperty(
+		name: 'adSlotsInline' | 'adSlotsTotal' | 'pageHeightVH',
```
zekehuntergreen, line 185: Is there a clever typescripty way to do this? We need to limit `name` to all of the members of `EventTimerProperties` whose type is `number`

src/track-scroll-depth.ts
```diff
@@ -0,0 +1,39 @@
...
+	// how many viewports tall is the page?
+	const pageHeightVH = Math.floor(pageHeight / intViewportHeight);
+	const eventTimer = EventTimer.get();
+	eventTimer.setProperty('pageHeightVH', pageHeightVH);
+
+	const observer = new IntersectionObserver((entries) => {
+		entries.forEach((entry) => {
+			if (entry.isIntersecting) {
+				const currentDepthVH = entry.target.getAttribute('data-depth');
+				console.log('current depth ', currentDepthVH);
```
arelra, line 22: 👀 maybe `libs/logger` instead?

src/track-scroll-depth.spec.ts
```diff
@@ -0,0 +1,61 @@
+import { EventTimer } from './EventTimer';
+import { initTrackScrollDepth } from './track-scroll-depth';
+
+describe('initTrackScrollDepth', () => {
+	const observe = jest.fn();
+	class IntersectionObserver {
+		constructor() {
+			return Object.freeze({
+				observe: observe,
```
arelra, line 9: Might be able to inline this:

```suggestion
				observe,
```

src/EventTimer.ts
chrislomaxjones: There might be, but I'm inclined to leave it as it is here. It's quite clear, and also if you tried to pull in all the keys with values of type `number`, you'd end up with `'downlink'` as well?

arelra: Something like (?):

```js
type ValidKeys<Type> = keyof Type;

setProperty(
	name: ValidKeys<
		Pick<
			EventTimerProperties,
			'adSlotsInline' | 'adSlotsTotal' | 'pageHeightVH'
		>
	>,
	value: number,
): void {
	this.properties[name] = value;
}
```

However I'm not sure if this is crossing that line of 'too clever and completely unreadable' .

There's a bunch of TS that makes me go cross-eyed and brain-hurty but where that line is is very subjective.

src/track-scroll-depth.ts
```diff
@@ -0,0 +1,44 @@
...
+const initTrackScrollDepth = () => {
+	const pageHeight = document.body.offsetHeight;
+	const intViewportHeight = window.innerHeight;
+	// how many viewports tall is the page?
+	const pageHeightVH = Math.floor(pageHeight / intViewportHeight);
+	const eventTimer = EventTimer.get();
+	eventTimer.setProperty('pageHeightVH', pageHeightVH);
+
+	const observer = new IntersectionObserver(
+		/* istanbul ignore next */
```
chrislomaxjones, line 18: Does this line exclude the following function from code coverage? And this is included because the intersection observer is mocked in the test?

If so, I wonder if it's better to leave this off? I think it's fine that this isn't tested, and aiming for 100% coverage is unrealistic, so I'd personally be happy to have a value more reflective of the true coverage of our tests. 

If it's for some other reason, please ignore me!

chrislomaxjones (commented):
This looks good to me! I just had one question about the code coverage.

src/track-scroll-depth.ts
zekehuntergreen: Yes, `istanbul ignore next` excludes the following function from coverage since there isn't a way to test the intersection observer with jest.
The pattern I've seen in our code bases is to ignore untestable / difficult-to-test code and third party code so that our coverage numbers reflect the ratio of tested code to testable code. To me the `ignore` signals that we're happy with this function being untested whereas removing it and leaving the coverage of this file at ~75% would signify that we'd like to improve that number at some point.
I feel like the best thing to do would be to test this functionality with cypress tests where we wouldn't need to mock intersection observer at all, but I think setting up this repo's first cypress tests for this feature might be overkill.

chrislomaxjones: I see, I hadn't thought of the distinction between testable code and the source as a whole. Happy to leave this in then!

chrislomaxjones (approved):
Excited to see this in action🖱️

PR #518, opened 2022-03-09
chore(deps): Bump dependencies 2022.03.09

zekehuntergreen:
## What does this change?
Updates dependencies by cherry-picking dependabot commits

```
git cherry-pick e511cbe
git cherry-pick 36a42b4
git cherry-pick 134210d
git cherry-pick cc4d448
git cherry-pick 2537769
git cherry-pick 016cbea
git cherry-pick 98b01fd
```

PR #523, opened 2022-03-14
feat: release v3.1.0

zekehuntergreen:
## What does this change?
- merging the PR should create a tag for commercial core and publish it to npm

## Why?
[Three previous attempts](https://github.com/guardian/commercial-core/runs/5510057262?check_suite_focus=true) to release v3.1.0 after merging #514 were unsuccessful for the following reasons
1. due to failed npm authentication
2. because semantic release couldn't find relevant changes to release (since on attempt 1, a github tag had been created)
3. main was ahead of a448133 since #518 was merged in the interim

Creating a new pull request and letting semantic release handle github tags and npm publishing seems like the easiest and cleanest way forward.


arelra (approved):
👍

PR #527, opened 2022-03-25
fix: Remove `video` ad size

simonbyford:
## What does this change?

Remove the `video` ad size

## Why?

It is never used


ashishpuliyel (approved):
Deprecate usually means it still works, but is discouraged.
Vs what we're doing here, which is dropping it entirely.

PR #531, opened 2022-03-28
refactor: Kebab case files

arelra:
## What does this change?

Kebab-case any camel/pascal-case files. 🐫 -> 🍖  

## Why?

Adhere to our convention


yarn-error.log
```diff
@@ -0,0 +1,7445 @@
+Arguments:
```
zekehuntergreen, line 1: Should this file have been checked in?

arelra: Oops! Will remove and add to `.gitignore`

PR #532, opened 2022-03-29
chore: Exports at bottom of files

chrislomaxjones:
## What does this change?

Introduce a new Eslint rule [`import/exports-last`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/exports-last.md) which enforces that all exports must be declared at the bottom of the file. Fix all occurrences where this isn't the case.

## Why?

As part of a new [migration strategy](https://github.com/guardian/commercial-core/pull/530), we'd like to standardise on how we manage file exports. This rule ensures all exports are at the bottom of the file, making it easy where to check which values are exported.

We enforce this with an Eslint rule in order to automate checks for this.

## Note on the rule

Note that this rule only forces that all exports must occur at the bottom of the file, and that you can still use named exports, so long as they appear at the end. Hence the following would be valid under this new rule:

```typescript
const a = 10;

export type A = 'hello';

export const foo = () => {
  // ...
}
``` 

```typescript
const a = 10;

export const foo = () => {
  // ...
}

export const bar = () => {
  // ...
}
```

Although this isn't enforced we typically group exports at the bottom of files, like so:

```typescript
const a = 10;

const foo = () => {
  // ...
}

const bar = () => {
  // ...
}

export { foo, bar };
```

There are exceptions to this such as [types.ts](https://github.com/guardian/commercial-core/blob/2c5ab42f816e1bb9ebd2396ff27a9faf3936892a/src/types.ts), since we export every value anyway it makes less sense to export them all again at the bottom.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,11 +5,15 @@
     "browser": true,
     "node": true
   },
-  "extends": ["@guardian/eslint-config-typescript"],
+  "extends": [
+    "@guardian/eslint-config-typescript",
+    "plugin:import/recommended"
+  ],
   "globals": { "googletag": "readonly" },
   "ignorePatterns": ["dist", "coverage", "src/__vendor"],
   "rules": {
     "curly": ["error", "multi-line"],
-    "no-use-before-define": ["error", { "functions": true, "classes": true }]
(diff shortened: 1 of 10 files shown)


arelra (approved):
👏

PR #535, opened 2022-04-07
feat: Get consent before sending commercial metrics

simonbyford:
## What does this change?

When in TCFv2 mode, check the user has consented to purpose 7 (measure ad performance) and 8 (measure content performance) of the TCF framework before sending commercial metrics to the server.

When in non-TCFv2 mode, no need to check anything.

Also makes some improvements to `send-commercial-metrics.spec.ts` which are easiest reviewed commit-by-commit. Some of these are down to taste and I'm definitely happy to be challenged if these go against any testing principles I'm not aware of!

## Why?

We shouldn't be collecting metrics without consent

### DIFF
--- a/src/send-commercial-metrics.spec.ts
+++ b/src/send-commercial-metrics.spec.ts
@@ -4,7 +4,6 @@ import {
 	bypassCommercialMetricsSampling,
 	initCommercialMetrics,
 } from './send-commercial-metrics';
-import type { Metric, Property, TimedEvent } from './send-commercial-metrics';
 
 const {
 	Endpoints,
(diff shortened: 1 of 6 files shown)


zekehuntergreen (commented):
Thanks for the additions and improvements to the tests 👏

src/send-commercial-metrics.ts
```diff
@@ -105,6 +108,8 @@ const roundTimeStamp = (events: TimedEvent[]): Metric[] => {
 };
 
 function sendMetrics() {
+	if (!consented) return;
```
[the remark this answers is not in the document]
simonbyford: I like this idea very much, for the reasons you gave. Thanks, will update!

src/send-commercial-metrics.ts
```diff
@@ -159,6 +164,21 @@ const addVisibilityListeners = (): void => {
 	window.addEventListener('pagehide', listener);
 };
 
+const addConsentListener = (): void => {
+	onConsentChange((consentState: ConsentState) => {
```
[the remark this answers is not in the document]
simonbyford: Cool. Let me investigate this and see if I can port it over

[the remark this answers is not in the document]
arelra: fyi @zekehuntergreen and I raised a PR for this 🙂  :
https://github.com/guardian/consent-management-platform/pull/597

simonbyford:
I've decided to hold off on this until the port of `getEnhancedConsent` to `consent-management-platform` is complete:

https://github.com/guardian/consent-management-platform/pull/597

We can then replace usages of `onConsentChange` with `getEnhancedConsent` in this PR

src/send-commercial-metrics.ts
```diff
@@ -75,15 +77,18 @@ const transformToObjectEntries = (
...
-const filterUndefinedProperties = <T>(
-	transformedProperties: Array<[string, T | undefined]>,
-): Array<[string, T]> =>
-	transformedProperties.reduce<Array<[string, T]>>((acc, [key, value]) => {
-		if (typeof value !== 'undefined') {
-			acc.push([key, value]);
-		}
-		return acc;
-	}, []);
+const filterUndefinedProperties = (
```
arelra, line 80: Just wondering if we could simplify this to a regular `array.filter()` ?

In which case I think the comparison would simplify down to `!!value` ?

src/send-commercial-metrics.spec.ts
```diff
@@ -288,66 +287,64 @@ describe('send commercial metrics code', () => {
...
-		type: undefined,
-		downlink: 1,
-		effectiveType: '4g',
-	};
-
-	const transformedProperties: Array<[string, string | number | undefined]> =
-		[
+	it('can transform event timer properties into object entries', () => {
+		expect(
+			transformToObjectEntries({
```
zekehuntergreen, line 450: I'm a fan of inlining the function calls and expected values like this - cuts down on jumping around the file to find these values.

yarn.lock
```diff
@@ -449,46 +455,48 @@
...
 "@guardian/ab-core@^2.0.0":
   version "2.0.0"
   resolved "https://registry.yarnpkg.com/@guardian/ab-core/-/ab-core-2.0.0.tgz#d8a9408d5a66ff1f8c684ed458c99517bbee15b1"
   integrity sha512-E5p6l0l37hY0GNFaZmA5/22TlVuavzlMHR3OeweD5eDcY2gB3nGgzPK/d3M67dkbDAJkGjB4a0W84Z5N2YScvQ==
 
 "@guardian/consent-management-platform@^10.1.0":
-  version "10.2.2"
-  resolved "https://registry.yarnpkg.com/@guardian/consent-management-platform/-/consent-management-platform-10.2.2.tgz#bffa22192c68b308574d6c6abfb22277058d3e32"
-  integrity sha512-lLdVTrDhxtf9BC3S2PNLMJf7/cR0Zo5RnfstWaynGerwOILwM8Ppl5xdHX/i6RdZrt2O4Z7GuB6mxhNAeAw27A==
+  version "10.5.0"
```
zekehuntergreen, line 484: I think we should update `package.json` to specify that we need CMP version ^10.5.0 since that version or higher is now required. 
Now that we're using `onConsent`, introduced in 10.5, installing 10.1..10.4 of the CMP would break commercial-core.

src/send-commercial-metrics.ts
zekehuntergreen: +1 for `filter` if this is possible but I think we'd still need to test for `value !== 'undefined'`. I think we want to preserve the behavior of sending metrics whose values are defined but falsy like 0, null, empty string.

arelra: Ah yes you're right. Good old falsy!

yarn.lock
simonbyford: Good point, I've made the change!

src/send-commercial-metrics.spec.ts
```diff
@@ -141,18 +275,27 @@ describe('send commercial metrics code', () => {
...
 			eventTimer.properties = {
 				downlink: 1,
 				effectiveType: '4g',
 			};
-			setVisibility();
+
+			setVisibility('hidden');
 			global.dispatchEvent(new Event('visibilitychange'));
 
-			expect(sentMetrics).toEqual(true);
```
zekehuntergreen, line 155: Why strip out these assertions that the return value of `initCommercialMetrics` is what we expect it to be? (Sorry I think I missed this earlier)

src/send-commercial-metrics.spec.ts
```diff
@@ -36,97 +43,224 @@ const defaultMetrics = {
...
+		});
+
+		setVisibility('hidden');
+		global.dispatchEvent(new Event('visibilitychange'));
+
+		expect((navigator.sendBeacon as jest.Mock).mock.calls).toEqual([]);
+	});
+
+	it('sends metrics when consent is non-TCFv2 (i.e. when user is in USA or Australia)', async () => {
+		mockOnConsent(nonTcfv2Consent);
```
zekehuntergreen, line 195: Testing that metrics are sent for CCPA and Aus users regardless of consent state, it might be more complete to add another test case where `doNotSell` is true.

simonbyford: Yup good shout, I will add!

src/send-commercial-metrics.spec.ts
simonbyford: I wanted to keep the tests very focused and this didn't seem super valuable to me, not worth the extra assertion. Whether or not the metrics actually get sent seemed more important

simonbyford: For what it's worth we never actually check this value anywhere! 😀

zekehuntergreen: Ok that makes sense - let's remove then

PR #548, opened 2022-04-27
chore: Replace git.io links with resolved links

arelra:
Co-authored-by: Max Duval <max.duval@theguardian.com>

## What does this change?

Replace git.io links with resolved links as git.io will be deprecated on 2022-04-29:

https://github.blog/changelog/2022-04-25-git-io-deprecation

Used @mxdvl 's guide:

https://gist.github.com/mxdvl/eae14de45152bfec04c9d7c9786ad997


chrislomaxjones (approved):
Thanks for doing this!

PR #549, opened 2022-04-28
feat: add size mappings to ad-sizes.ts

Jakeii:
## What does this change?
Add size mappings to `ad-sizes.ts` so they can be imported from DCR, rather than inline in the component.

I think right to move to this package? But not totally sure

## Why?
Simplify how size mappings are managed.


arelra (approved):
Nice 👍

PR #576, opened 2022-05-20
feat: Collect commercial metrics on GPC signal

kenoir:
## What does this change?

Adds a function `initTrackGpcSignal` which collects metrics on the [GPC signal](https://globalprivacycontrol.org/#gpc-spec) when handed a `ConsentState` from the [guardian/consent-management-platform](https://github.com/guardian/consent-management-platform) library. 

## Why?

We'd like to understand the commercial impact of the GPC signal on our platform and track any changes over time.

Depends on https://github.com/guardian/consent-management-platform/pull/606

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "@commitlint/cli": "^16.1.0",
     "@commitlint/config-conventional": "^17.0.0",
     "@guardian/ab-core": "^2.0.0",
-    "@guardian/consent-management-platform": "^10.5.0",
+    "@guardian/consent-management-platform": "guardian/consent-management-platform#notice-gpc-navigator",
     "@guardian/eslint-config-typescript": "^1.0.0",
     "@guardian/libs": "5.1.0",
     "@guardian/prettier": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "@commitlint/cli": "^16.1.0",
     "@commitlint/config-conventional": "^17.0.0",
     "@guardian/ab-core": "^2.0.0",
-    "@guardian/consent-management-platform": "guardian/consent-management-platform#notice-gpc-navigator",
+    "@guardian/consent-management-platform": "^10.6.0",
     "@guardian/eslint-config-typescript": "^1.0.0",
     "@guardian/libs": "5.1.0",
     "@guardian/prettier": "^1.0.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -45,7 +45,7 @@
     "@commitlint/cli": "^16.1.0",
     "@commitlint/config-conventional": "^16.0.0",
     "@guardian/ab-core": "^2.0.0",
-    "@guardian/consent-management-platform": "^10.5.0",
+    "@guardian/consent-management-platform": "guardian/consent-management-platform#notice-gpc-navigator",
```
kenoir, line 48: This should be updated when the CMP lib is released.

src/track-gpc-signal.ts
```diff
@@ -0,0 +1,21 @@
...
+import { EventTimer } from './event-timer';
+
+/**
+ * Collect metrics on gpcSignal presence and value
+ * https://globalprivacycontrol.github.io/gpc-spec/
+ */
+const initTrackGpcSignal = (consentState: ConsentState) => {
+	// If undefined we set the property value to -1, false is 0, true is 1
+	const gpcSignal =
+		consentState.gpcSignal === undefined ? -1 : +consentState.gpcSignal;
```
ashishpuliyel, line 12: Why do we actually need to create this dependency on CMP here? What does the indirection give us?
Can't we just do `navigator.globalPrivacyControl` ?

kenoir: We could, but I prefer keeping the CMP the source of truth. 

For example the GPC can also be signalled by a HTTP header, and we might implement some mechanism to notice that and we'd want to keep the logic for saying the "GPC is ON" in one place, likewise future potential extensions or whatever might lead us down the same path.

ashishpuliyel: Fair point and well made.

arelra, line 12: Do we need to coerce with `+`?

```suggestion
		consentState.gpcSignal === undefined ? -1 : consentState.gpcSignal;
```

arelra, line 12: Minor point for a follow up but if we always expect the `gpcSignal` property to be present should the type from CMP be `gpcSignal: undefined | boolean` rather than optional?

arelra (approved):
👍 Just a couple of minor comments

src/track-gpc-signal.ts
kenoir: I agree, it should be `gpcSignal: undefined | boolean` as you describe. Are you happy to merge this and i'll follow up with a fix on the CMP lib afterwards?

kenoir: Yes, this seems to be required otherwise the build fails complaining i've got the wrong type here.

arelra: Sure that's fine

PR #585, opened 2022-05-31
feat: Migrate create-ad-slot.ts

Jakeii:
## What does this change?
Migrate create-ad-slot.ts and it's test to commerical-core

### DIFF
--- a/src/create-ad-slot.ts
+++ b/src/create-ad-slot.ts
@@ -1,0 +1,311 @@
+import type { AdSize, SizeMapping } from './ad-sizes';
+import { adSizes } from './ad-sizes';
+
+const adSlotIdPrefix = 'dfp-ad--';
+
+type AdSlotConfig = {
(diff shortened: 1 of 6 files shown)


src/create-ad-slot.ts
```diff
@@ -0,0 +1,311 @@
...
+		label: true,
+		refresh: true,
+		name: 'mobile-sticky',
+		sizeMappings: {
+			mobile: [adSizes.mobilesticky],
+		},
+	},
+};
+
+/*
```
chrislomaxjones, line 167:
```suggestion
/**
```

chrislomaxjones (commented):
Great stuff! I had a few thoughts off the back of our mobbing below.

Personally, I also think it's fine to tolerate the drop in test coverage. I'd prefer to get this in and we can think about bringing the coverage back up later in separate PRs.

arelra (approved):
🎉

PR #598, opened 2022-06-13
feat: Copy frontend size mappings into ad-sizes

Jakeii:
## What does this change?
Copy frontend size mappings into ad-sizes

## Why?
Sorry, yet another PR on ad sizes

Didn't do this initially because of the clash with the 'top-above-nav' name, but have noticed it's actually safe to do as frontend only uses the `mobile` size in the browser and dotcom the rest

This is needed in order to lookup the ad sizes in frontend, they _could_ be kept in separate objects but unnecessary I think and would require different object names etc.

### DIFF
--- a/src/ad-sizes.ts
+++ b/src/ad-sizes.ts
@@ -74,7 +74,16 @@ type SlotName =
 	| 'mostpop'
 	| 'merchandising'
 	| 'merchandising-high'
-	| 'survey';
+	| 'merchandising-high-lucky'
+	| 'survey'
+	| 'im'
+	| 'inline'
+	| 'mostpop'
+	| 'comments'
+	| 'top-above-nav'
+	| 'carrot'
+	| 'epic'
+	| 'mobile-sticky';
 
 type Breakpoint = 'mobile' | 'desktop' | 'phablet' | 'tablet';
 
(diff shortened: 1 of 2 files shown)


domlander (approved):

chrislomaxjones (commented):
I'm not entirely sure it's safe to merge the `top-above-nav` size mappings, even though its separate sets of breakpoints. 

I think merging them would be the equivalent of saying "on both DCR and Frontend, all the sizes at all these breakpoints would be eligible", whereas I think we want to say "on DCR these phablet/desktop ones are eligible, and on Frontend the mobile ones are eligible". Someone else on the team may want to chime if I'm mistaken (cc @guardian/commercial-dev)!

I agree though it's best to avoid exporting two sets of size mappings. Could we possibly have an object `staticSizeMappings` (DCR) and `dynamicSizeMappings` (Frontend)? Then I think we can export the static ones for DCR, and `create-ad-slot` can use the dynamic ones?

EDIT: It may make sense to have a `commonSizeMappings` object that contains all those that exist in both static/dynamic? This may be a trade-off between reducing duplication and having separate overloaded slot names on different platforms.

chrislomaxjones (approved):
As discussed, if this looks good when testing it against Frontend, and since these slots only exist at certain breakpoints anyway, I think we should be fine to merge `top-above-nav`.

PR #599, opened 2022-06-16
fix: Only send commercial metrics once

domlander:
## What does this change?

Send commercial metrics a maximum of one time for each pageview.

## Why?

@ashishpuliyel has a hypothesis that some of the outliers in the timing data is related to browsers that suspend the window, and then resume (tabs put into the background and resumed the next day for example).

The timers become somewhat meaningless once the tab has been put into the background, so we should stop the timers after that.

### DIFF
--- a/src/send-commercial-metrics.ts
+++ b/src/send-commercial-metrics.ts
@@ -145,10 +145,10 @@ const listener = (e: Event): void => {
 
 const addVisibilityListeners = (): void => {
 	// Report all available metrics when the page is unloaded or in background.
-	window.addEventListener('visibilitychange', listener);
+	window.addEventListener('visibilitychange', listener, { once: true });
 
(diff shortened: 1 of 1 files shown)


ashishpuliyel (approved):
Lol this was easy. Let's see if this hurts or helps. Send d&I a note when it's merged and they can tell us

domlander:
> Lol this was easy. Let's see if this hurts or helps. Send d&I a note when it's merged and they can tell us

One line changes are the best 😎. But yeah, Chris and I agreed that sending this live and then monitoring the amount of outliers was a good way forward. We're not sure whether this will fix the issue by reducing the amount of outliers, but it's a reasonable attempt, a sensible change and we'll gain information going forward.

I will make D&I aware of this change when it is merged

PR #603, opened 2022-06-20
feat: Add logic for setting Ad Manager group cookie

domlander:
## What does this change?

- Set cookie `amtgrp` if the user is in CCPA or aus, regardless of the user's consent state. This is because this cookie is not used for targeting/personalisation. It is used for A/B tests.
- Remove superfluous functions: `personalisedAdvertising` and `tcfv2AllPurposesConsented`.

## Why?

- To correct logic whilst migrating commercial code over from frontend to commercial-core.


arelra (approved):
🎉

PR #619, opened 2022-07-18
feat: Add crossword banner slot to slot size definitions

emma-imber:
## What does this change?

Adding a new slot size mapping for the crossword banner ads that appear on tablet displays.

## Why?

At the moment the crossword banner ad slot is not getting the right sizes of adverts, causing ads to overlap comments on the page. Adding this new size mapping is the first step towards fixing this bug.


Jakeii (approved):

arelra (approved):
👍

PR #626, opened 2022-07-25
feat: Refactor and export `concatSizeMappings`

chrislomaxjones:
## What does this change?

Deal with one of our `// TODO` markers by refactoring the `concatSizeMappings` function.

This requires the introduction of an `isBreakpoint` type guard - to ensure we don't concat additional keys in objects. Note that because of the use of `reduce` we won't affect the original size mappings.

## Why?

> TODO

### DIFF
--- a/src/ad-sizes.ts
+++ b/src/ad-sizes.ts
@@ -88,6 +88,9 @@ type SlotName =
 
 type Breakpoint = 'mobile' | 'desktop' | 'phablet' | 'tablet';
 
+const isBreakpoint = (s: string): s is Breakpoint =>
+	s === 'mobile' || s === 'phablet' || s === 'tablet' || s === 'desktop';
+
 type SizeMapping = Partial<Record<Breakpoint, AdSize[]>>;
 
 type SlotSizeMappings = Record<SlotName, SizeMapping>;
@@ -322,4 +325,4 @@ export type {
 	SlotSizeMappings,
 	SlotName,
 };
-export { adSizes, getAdSize, slotSizeMappings, createAdSize };
+export { adSizes, getAdSize, slotSizeMappings, createAdSize, isBreakpoint };
--- a/src/ad-sizes.ts
+++ b/src/ad-sizes.ts
@@ -1,3 +1,5 @@
+import type { Breakpoint } from './lib/breakpoint';
+
 type AdSizeString = 'fluid' | `${number},${number}`;
 
 /**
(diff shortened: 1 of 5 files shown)


src/create-ad-slot.ts
```diff
@@ -148,28 +148,21 @@ const createClasses = (
  */
 const concatSizeMappings = (
```
domlander, line 149: It would be good to have some tests around this breakpoint behaviour if possible

src/create-ad-slot.ts
```diff
@@ -148,28 +148,21 @@ const createClasses = (
...
-		if (optionSizeMappings[device] && defaultSizeMappings[device]) {
-			const optionSizeMappingsForDevice = optionSizeMappings[device];
-			const defaultSizeMappingsForDevice = defaultSizeMappings[device];
-
-			if (defaultSizeMappingsForDevice && optionSizeMappingsForDevice) {
-				// TODO can we do concatenatedSizeMappings[device]?.concat ?
-				concatenatedSizeMappings[device] = (
-					concatenatedSizeMappings[device] ?? []
-				).concat(optionSizeMappingsForDevice);
+	optionSizeMappings?: SizeMapping,
```
Jakeii, line 151: would defaulting to `{}` instead of the coalesce bellow be more readable?
```suggestion
	optionSizeMappings: SizeMapping = {},
```

chrislomaxjones: Agreed, Much nicer 👍

src/create-ad-slot.ts
chrislomaxjones: One of the existing tests covers rejecting invalid keys (see the `invalid` mapping in the test), but I'll add some tests around mutating the original size mappings.

src/ad-sizes.ts
```diff
@@ -88,6 +88,9 @@ type SlotName =
 
 type Breakpoint = 'mobile' | 'desktop' | 'phablet' | 'tablet';
 
+const isBreakpoint = (s: string): s is Breakpoint =>
```
arelra, line 91: This feels like it should be in `/lib` rather than being exported from here?

src/create-ad-slot.ts
chrislomaxjones: Actually I may delay this to a follow-up PR, I thought it'd be easy but looks like something isn't quite right with the ad sizes when running Jest. For example, running the following gives me `TypeError: undefined is not a function` rather than a proper diff of the expected/actual

```typescript
expect({ desktop: [adSizes.mobilesticky] }).toEqual({
	desktop: [adSizes.googleCard],
});
```

arelra (approved):
Nice improvement!

src/ad-sizes.ts
chrislomaxjones: Good idea. I think the type and this function should stay close together so I've moved both to a file in lib.

PR #628, opened 2022-07-29
feat: createAdSlot no longer adds size mappings data attrs

Jakeii:
## What does this change?
Removes anything to do with sizeMappings from `createAdSlot`, as they no longer need to be written to the data attributes.

## Why?
They are redundant now, size mappings are now loaded from https://github.com/guardian/commercial-core/blob/main/src/ad-sizes.ts#L143 as of this frontend PR https://github.com/guardian/frontend/pull/25124


arelra (approved):
Great to see all this code being deleted! 🚀

PR #633, opened 2022-08-12
feat: add new function to return the array of an ad size

emma-imber:
## What does this change?

Adding a new function to return the array of a given ad size, and added some relevant tests to make sure the function behaves as expected.

## Why?

This will be useful in frontend to make sure that we send arrays with working array methods on them, such as splice.

### DIFF
--- a/src/ad-sizes.ts
+++ b/src/ad-sizes.ts
@@ -97,6 +97,10 @@ const createAdSize = (width: number, height: number): AdSize => {
 	return new AdSize([width, height]);
 };
 
+const getAdSizeArray = (adSize: AdSize): number[] => {
+	return [adSize.width, adSize.height];
+};
+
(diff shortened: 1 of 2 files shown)


src/ad-sizes.ts
```diff
@@ -40,6 +40,10 @@ class AdSize extends Array<number> {
 	get height(): number {
 		return this[1];
 	}
+
+	get array(): number[] {
```
chrislomaxjones, line 44: What do you think about aligning this with the way we call `toString` above by having a `public toArray()` method rather than a getter?

emma-imber: Great shout, that's definitely more consistent! I'll make the change

chrislomaxjones (approved):
👍

PR #641, opened 2022-08-24
chore: Bump deps 2022-08-24

arelra:
## What does this change?

Bump deps 2022-08-24

Bumping TypeScript seems to have flagged a new error:

```
error TS2564: Property '[0]' has no initializer and is not definitely assigned in the constructor
```

```
FAIL src/ad-sizes.test.ts
  ● Test suite failed to run
    src/ad-sizes.ts:[23](https://github.com/guardian/commercial-core/runs/7992252393?check_suite_focus=true#step:5:24):11 - error TS2564: Property '[0]' has no initializer and is not definitely assigned in the constructor.
    23  readonly [0]: number;
                 ~~~
    src/ad-sizes.ts:[24](https://github.com/guardian/commercial-core/runs/7992252393?check_suite_focus=true#step:5:25):11 - error TS[25](https://github.com/guardian/commercial-core/runs/7992252393?check_suite_focus=true#step:5:26)64: Property '[1]' has no initializer and is not definitely assigned in the constructor.
    24  readonly [1]: number;
                 ~~~
```

I've fixed this by updating the constructor of `ad-sizes.ts` as follows:

```diff
-		this.push(width, height);
+		this[0] = width;
+		this[1] = height;
```

### DIFF
--- a/src/ad-sizes.test.ts
+++ b/src/ad-sizes.test.ts
@@ -16,6 +16,7 @@ describe('ad sizes', () => {
 			const adSize = createAdSize(expectedWidth, expectedHeight);
 			expect(adSize.width).toEqual(expectedWidth);
 			expect(adSize.height).toEqual(expectedHeight);
+			expect(adSize.length).toEqual(2);
 			expect(adSize.toString()).toEqual(expectedString);
 		},
 	);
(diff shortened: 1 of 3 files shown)


chrislomaxjones (approved):
Nice, thanks for fixing that error 👍

PR #665, opened 2022-09-09
feat: Collect commercial metrics on labs containers

simonbyford:
## What does this change?

todo

## Why?

todo


SiAdcock (approved):
LGTM 💎

chrislomaxjones (approved):
Even without full context on the labs tracking work, this is very clear 👍

src/track-labs-container.ts
```diff
@@ -0,0 +1,31 @@
...
+ * - when the element is scrolled into view
+ */
+const initTrackLabsContainer = () => {
+	const target = document.querySelector('section.dumathoin');
+	if (target === null) return;
+
+	const eventTimer = EventTimer.get();
+
+	log('commercial', 'Page has labs container');
+	eventTimer.trigger('hasLabsContainer');
```
zekehuntergreen, line 16: Non-blocking: it might be worth considering using `event.Timer.setProperty` and treating `hasLabsContainer` and possibly `labsContainerInView` as properties of the page rather than as metrics. 
`labsContainerInView` seems like it could be considered an "event" and that we might be interested in analyzing the timestamps associated with it. I'm not sure that the same is true of `hasLabsContainer` though.

PR #670, opened 2022-09-15
feat: Capture `labsUrl` when container is present

simonbyford:
## What does this change?

Follows on from https://github.com/guardian/commercial-core/pull/665

Modifies `initTrackLabsContainer` to also capture `labsUrl` when a Guardian Labs container is present. 

In the below case, it would set `labsUrl` to `/a-vision-for-better-food`

Screenshot 2022-09-15 at 09 03 05

## Why?

So we know which campaign the reader saw.


chrislomaxjones (approved):
Looks good

PR #350, opened 2021-09-06
chore(deps-dev): Bump deps 2021-09-06

mxdvl:
## What does this change?

Bump dependencies and update the `NavigatorConnection` type.

- chore(deps-dev): bump @guardian/libs from 3.0.0 to 3.2.0
- chore(deps-dev): bump @guardian/types from 8.0.0 to 8.1.0
- chore(deps-dev): bump typescript from 4.3.5 to 4.4.2
- chore(deps-dev): bump eslint-plugin-import from 2.24.1 to 2.24.2
- chore(deps-dev): bump semantic-release from 17.4.6 to 17.4.7
- rebuild yarn.lock

## Why?

Keep things fresh!

PR #675, opened 2022-09-16
feat: upgrade Typescript and add new global types

SiAdcock:
## What does this change?

- upgrade typescript to v4.8
- add types that appear to have been dropped (builds on #350)

## Why?

Keeping fresh, and alignment with csnx

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "semantic-release": "^19.0.2",
     "ts-jest": "^27.0.7",
     "type-fest": "^2.8.0",
-    "typescript": "^4.5.2",
+    "typescript": "^4.8.3",
     "web-vitals": "^2.1.2"
   },
   "publishConfig": {
--- a/src/global.ts
+++ b/src/global.ts
@@ -6,10 +6,28 @@ import type {
 import type { EventTimer } from '.';
 
 declare global {
+	/**
+	 * lib.dom.d.ts only includes features that are supported in at least two main browsers.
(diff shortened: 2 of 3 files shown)


arelra (approved):
👍

SiAdcock:
There's something going on here that I don't fully understand. Might be worth getting to the bottom of. 

`NetworkInformation` and `ConnectionType` were removed from TypeScript's DOM library in microsoft/TypeScript-DOM-lib-generator#1311 as a result of an upgrade to `@mdn/browser-compat-data`. 

It looks like in mdn/browser-compat-data#15363, the `NetworkInformation` was disabled. I'll have to do some more digging to find out why, and whether it's causing the removal of type information in microsoft/TypeScript-DOM-lib-generator#1311

SiAdcock:
TypeScript's `lib.dom.d.ts` only includes features that are supported in at least [two main browsers](https://github.com/microsoft/TypeScript-DOM-lib-generator#why-is-my-fancy-api-still-not-available-here). Currently, `NetworkInformation` is supported in [Chrome only](https://github.com/mdn/browser-compat-data/blob/main/api/NetworkInformation.json) – it's behind a flag in Firefox.

If / when the `NetworkInformation` API is more widely supported, we can remove this shim. I'll add a comment explaining this.

arelra:
Hey @SiAdcock  sorry I didn't see this so the types were also updated as part of: https://github.com/guardian/commercial-core/pull/704

So I think this PR can be closed?

arelra:
Closing as no longer required

PR #677, opened 2022-09-21
feat: Migrate messenger

Jakeii:
## What does this change?
Migrate `messenger.ts`, it's test and `post-message.ts` to commercial-core

### DIFF
--- a/src/messenger.ts
+++ b/src/messenger.ts
@@ -1,0 +1,531 @@
+import { postMessage } from './messenger/post-message';
+
+/**
+ * The type of iframe messages we accept
+ */
+type MessageType =
+	| 'background'
+	| 'click'
+	| 'disable-refresh'
+	| 'get-page-targeting'
+	| 'get-page-url'
+	| 'get-styles'
+	| 'measure-ad-load'
+	| 'passback'
+	| 'resize'
+	| 'set-ad-height'
+	| 'scroll'
+	| 'type'
+	| 'viewport';
+
+/**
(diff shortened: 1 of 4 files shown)


src/messenger.ts
```diff
@@ -0,0 +1,531 @@
...
+			errorHandler: MessengerErrorHandler,
+		) => void
+	>,
+	persistentListeners: Array<
+		(
+			register: RegisterPersistentListener,
+			errorHandler: MessengerErrorHandler,
+		) => void
+	>,
+	errorHandler: MessengerErrorHandler,
```
Jakeii, line 513: I thought passing in the raven/sentry error handler would be better rather than copying it over, but not totally sure.

arelra (approved):
A big one!

Very minor but what do you think about putting these files into `/messenger`? The root is getting quite busy now.

Jakeii:
> A big one!
> 
> Very minor but what do you think about putting these files into `/messenger`? The root is getting quite busy now.

Have put post-message in the folder already :), but I was thinking about how all the exports are at the root, might start getting busy there.

I'm not sure how to get some sort of alias so that `import type { RegisterListener } from '@guardian/commercial-core/messenger';` would work, as `import type { RegisterListener } from '@guardian/commercial-core/dist/esm/messenger';` feels a bit gross.
`

arelra:
I don't personally mind the `index.ts` file being busy and exporting everything from the root path as long as it was well organised.

We could have intermediate `index.ts`s in each folder so the the root just exports everything from each subfolder.

PR #686, opened 2022-10-04
feat: Migrate `build-page-targeting` from frontend

simonbyford:
Co-authored-by: Ravi <7014230+arelra@users.noreply.github.com>
Co-authored-by: Zeke Hunter-Green <zeke.huntergreen@guardian.co.uk>

## What does this change?

Moves the logic to build page targeting from [frontend](https://github.com/guardian/frontend/blob/main/static/src/javascripts/projects/common/modules/commercial/build-page-targeting.ts). A subsequent PR in frontend will remove the old logic and instead call this method. After that, we'll do the same for DCR.

## Why?

So we can standardise the way we build page targeting across DCR and frontend. Currently there is a lot of duplication and inconsistency.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -60,3 +60,4 @@ export type {
 	RespondProxy,
 } from './messenger';
 export { postMessage } from './messenger/post-message';
+export { buildPageTargeting } from './targeting/build-page-targeting';
--- a/src/targeting/build-page-targeting.ts
+++ b/src/targeting/build-page-targeting.ts
@@ -73,14 +73,6 @@ const buildPageTargeting = (
 	countryCode: CountryCode,
 	clientSideParticipations: Participations,
 ): PageTargeting => {
-	if (
-		!window.guardian.config.page ||
-		!window.guardian.config.ophan ||
-		window.guardian.config.isDotcomRendering === undefined
-	) {
-		return {};
-	}
-
 	const { page, isDotcomRendering } = window.guardian.config;
 
 	const adFreeTargeting: { af?: True } = adFree ? { af: 't' } : {};
--- a/src/targeting/build-page-targeting.ts
+++ b/src/targeting/build-page-targeting.ts
@@ -1,0 +1,150 @@
+import type { Participations } from '@guardian/ab-core';
+import { cmp } from '@guardian/consent-management-platform';
+import type { ConsentState } from '@guardian/consent-management-platform/dist/types';
+import type { CountryCode } from '@guardian/libs';
+import { getCookie, isString } from '@guardian/libs';
+import type { False, True } from '../types';
+import type { ContentTargeting } from './content';
+import { getContentTargeting } from './content';
+import type { AdManagerGroup, Frequency } from './personalised';
+import { getPersonalisedTargeting } from './personalised';
+import type { SessionTargeting } from './session';
+import { getSessionTargeting } from './session';
+import type { SharedTargeting } from './shared';
+import { getSharedTargeting } from './shared';
+import type { ViewportTargeting } from './viewport';
+import { getViewportTargeting } from './viewport';
+
+type PartialWithNulls<T> = { [P in keyof T]?: T[P] | null };
+
+type TrueOrFalse = True | False;
+
+type PageTargeting = PartialWithNulls<
+	{
+		ab: string[];
+		af: 't'; // Ad Free
+		amtgrp: AdManagerGroup;
+		at: string; // Ad Test
+		bp: 'mobile' | 'tablet' | 'desktop'; // BreakPoint
+		cc: CountryCode; // Country Code
+		cmp_interaction: string;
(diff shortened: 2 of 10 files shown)


src/targeting/build-page-targeting.ts
```diff
@@ -0,0 +1,151 @@
...
+	countryCode: CountryCode,
+	clientSideParticipations: Participations,
+): PageTargeting => {
+	if (
+		!window.guardian.config?.page ||
+		!window.guardian.config.ophan ||
+		window.guardian.config.isDotcomRendering === undefined
+	) {
+		return {};
+	}
```
simonbyford, line 82: This doesn't feel great, is there a better way?

arelra: Is this new functionality?

I'm not sure how 'fatal' these conditions are. Maybe there's an argument to say we should try to send as much targeting as we can even if some of the properties are empty? Might be worth asking Alex.

chrislomaxjones: There's some mismatches between what [Frontend considers present](https://github.com/guardian/frontend/blob/main/static/src/javascripts/types/global.d.ts#L376) on the window and what [commercial-core does](https://github.com/guardian/commercial-core/blob/main/src/global.ts#L23) - could we fix our types up to match what is present in Frontend?

src/targeting/build-page-targeting.ts
```diff
@@ -0,0 +1,151 @@
...
+
+	const { page } = window.guardian.config;
+
+	const { isDotcomRendering } = window.guardian.config;
+
+	const adFreeTargeting: PageTargeting = adFree ? { af: 't' } : {};
```
chrislomaxjones, line 88: I assume this is just a straight migration but whilst we're here could we apply a more narrow type to this? Maybe

```suggestion
	const adFreeTargeting: { af?: 't' } = adFree ? { af: 't' } : {};
```

src/index.ts
```diff
@@ -60,3 +60,4 @@ export type {
 	RespondProxy,
 } from './messenger';
 export { postMessage } from './messenger/post-message';
+export { buildPageTargeting } from './targeting/build-page-targeting';
```
chrislomaxjones, line 63: Not certain but it might be handy to export the `PageTargeting` type to consumers of commercial-core as well.

src/targeting/build-page-targeting.ts
arelra: I would have thought its better in general to try and send as much targeting as possible even if some properties are missing in order to maximise fulfilment.

But putting that aside:

[Dotcom](https://github.com/guardian/dotcom-rendering/blob/89706d26c0298b19449c49a9a79681b8cbe3af98/dotcom-rendering/src/model/window-guardian.ts#L9) and [Frontend](https://github.com/guardian/frontend/blob/1aaed4efc97a181f06b7a4a29d7ac528d01db7aa/static/src/javascripts/types/global.d.ts#L79) consider `guardian.config.page` (and `guardian.config.isDotcomRendering`) to be present.

[Dotcom](https://github.com/guardian/dotcom-rendering/blob/89706d26c0298b19449c49a9a79681b8cbe3af98/dotcom-rendering/src/model/window-guardian.ts#L33) and [Frontend](https://github.com/guardian/frontend/blob/1aaed4efc97a181f06b7a4a29d7ac528d01db7aa/static/src/javascripts/types/global.d.ts#L74) consider `guardian.config.ophan` to be present.

[Dotcom](https://github.com/guardian/dotcom-rendering/blob/89706d26c0298b19449c49a9a79681b8cbe3af98/dotcom-rendering/window.guardian.d.ts#L24) considers `guardian.ophan` to be optional whereas [Frontend](https://github.com/guardian/frontend/blob/1aaed4efc97a181f06b7a4a29d7ac528d01db7aa/static/src/javascripts/types/global.d.ts#L375) does not.

So as long as we use `guardian.config.ophan.pageViewId` I think we can rely on Dotcom and Frontend getting the types right and remove this if block?

simonbyford: Thanks both, that is really helpful. I think fixing the inconsistencies Chris mentioned should just fix this and remove the need for the guard statement

simonbyford: I've resolved this [here](https://github.com/guardian/commercial-core/pull/686/commits/ba47f077093492ec8a4897fa10a142d6452e6073), should be good now!

src/targeting/build-page-targeting.ts
simonbyford: Good shout, have changed!

src/index.ts
simonbyford: Good idea, I've added this. Will update consumers in the upcoming frontend PR

arelra (approved):
Nice work! 🚀

PR #688, opened 2022-10-06
feat: Export standard ad sizes

domlander:
## What does this change?

1. Exports standard ad sizes
2. Adds ad label height to constants

## Why?

1. So that in other apps we can distinguish between standard ad-sizes and sizes that are variable (e.g. fluid) or do not show up on the page (e.g. [1,1 ], [2, 2]) or look like a standard ad size but are actually variable (e.g. merchandising slot: [88, 87]). It will be useful in frontend to know if the top-above-ad is of a standard ad size or a fluid ad. I don't want to duplicate ad sizes in the frontend repo.
2. We define this constant in multiple places in both frontend and dotcom-rendering. Best to define it in one place that makes sense.

### DIFF
--- a/src/constants/index.spec.ts
+++ b/src/constants/index.spec.ts
@@ -1,4 +1,4 @@
-import { PREBID_TIMEOUT, TOP_ABOVE_NAV_HEIGHT } from '.';
+import { AD_LABEL_HEIGHT, PREBID_TIMEOUT, TOP_ABOVE_NAV_HEIGHT } from '.';
 
 // These tests ensure that we look twice when changing a constant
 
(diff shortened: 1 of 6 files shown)


src/constants/index.spec.ts
```diff
@@ -10,4 +10,8 @@ describe('Constant values are constant', () => {
 	test('PREBID_TIMEOUT', () => {
 		expect(PREBID_TIMEOUT).toBe(1500);
 	});
+
+	test('PREBID_TIMEOUT', () => {
```
chrislomaxjones, line 14:
```suggestion
	test('AD_LABEL_HEIGHT', () => {
```

chrislomaxjones (approved):
Nice improvement!

PR #689, opened 2022-10-06
feat: Build consentless page targeting

zekehuntergreen:
Co-Authored-By: Jake Lee Kennedy <jake@bodhi.io>

## What does this change?
- adds function `buildPageTargetingConsentless` which calls the consented version, `buildPageTargeting` and returns only the targeting key / values needed for Opt Out integration

## Why?
- Custom targeting functionality is needed to release our Opt Out integration, showing real ads to readers.

### DIFF
--- a/src/targeting/build-page-targeting-consentless.ts
+++ b/src/targeting/build-page-targeting-consentless.ts
@@ -34,9 +34,8 @@ type ConsentlessPageTargeting = Partial<
 	Pick<PageTargeting, ConsentlessTargetingKeys>
 >;
 
-const isConsentlessKey = (key: unknown): key is ConsentlessTargetingKeys => {
-	return consentlessTargetingKeys.includes(key as ConsentlessTargetingKeys);
-};
+const isConsentlessKey = (key: unknown): key is ConsentlessTargetingKeys =>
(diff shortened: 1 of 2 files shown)


src/targeting/build-page-targeting-consentless.ts
```diff
@@ -0,0 +1,71 @@
...
+] as const;
+
+type ConsentlessTargetingKeys = typeof consentlessTargetingKeys[number];
+
+type ConsentlessPageTargeting = Partial<
+	Pick<PageTargeting, ConsentlessTargetingKeys>
+>;
+
+const isConsentlessKey = (key: unknown): key is ConsentlessTargetingKeys => {
+	return consentlessTargetingKeys.includes(key as ConsentlessTargetingKeys);
```
arelra, line 38: This is just code golf but this could be an implicit return

arelra (approved):
Nice!

src/targeting/build-page-targeting-consentless.ts
zekehuntergreen: Good call

chrislomaxjones: 🏌️

PR #693, opened 2022-10-17
feat: Export outstreamSizes. Add isFixedSize method to AdSize

domlander:
## What does this change?

- Move outstreamSizes to object so it can be exported.
- Adds pro to adSize class to differentiate between standard ads that display within their slot and other ads where the size is not reflective of the size of the advert

## Why?

- This will avoid duplication in Frontend.
- It is helpful to know whether an advert is going to load within its slot or expand its boundaries.


src/ad-sizes.ts
```diff
@@ -38,6 +38,13 @@ class AdSize extends Array<number> {
 		return [this[0], this[1]];
 	}
 
+	// The inserted advert is a fixed size within its slot.
+	// Examples that are not a fixed size are: fluid ads, empty or
+	// out of page ads, and merch ads in a [88, ??] size format.
+	public isFixedSize(): boolean {
```
domlander, line 44: Is there a better name for this? I want to differentiate between the ad sizes that expect an advert to load normally within the slot and are constrained by the size, and adverts that can expand out of the slot or are dummy sizes.

arelra: This is tricky naming! Some options that spring to mind:

positive: `Real`, `Actual`,  `Numeric` 
negative: `Proxy`, `Symbol`, `Token`

I think `isNumeric` / `isProxy` are my favourites so far but interested in what others think.

src/ad-sizes.ts
```diff
@@ -38,6 +39,13 @@ class AdSize extends Array<number> {
...
+		return this.width > 2 && this.width !== 88;
```
arelra, line 46: While this is very smart I worry about the readability a little.

Could we do something like:

```suggestion
    const isOutOfPage = this.width === 1 && this.height === 1;
    const isFluid = this.toString() === 'fluid';
    const isProxy = this.width === 88;
		return !(isOutOfPage || isFluid || isProxy);
```

src/ad-sizes.ts
Jakeii: Could use the term `flexible` for fluid, merch etc? then `isFlexible` and `!isFlexible()`

src/ad-sizes.ts
domlander: Yeah this is much better 👍

src/ad-sizes.ts
domlander: Thanks both for your suggestions! I used fixedSize as I found that phrase is the Google definition of fluid ads.

I like isProxy the best as it more closely describes the size of the ad as opposed to the advert so I'll go with that.

OR...should this live on the Advert class in Frontend? Is this so hard to name because it doesn't belong in this AdSize class?

arelra: I'd be in favour of keeping the logic here for now rather than Frontend.

But it does feel like `AdSize` is mixing concerns between storing a single size and then knowing about 'other' sizes. We can keep as is for now and follow up with a refactor?

domlander: Yeah I agree with that there is a mixing of concerns that could use a refactor. 

Also, in frontend, we let the size equal the string "fluid", which isn't quite right because AdSize already handles the case of a fluid ad:
```
class Advert {
...
size: AdSize | 'fluid' | null = null;
...
}
```

arelra (approved):
👍

PR #694, opened 2022-10-18
feat: Remove `countryCode` parameter from `buildPageTargeting` and add spec

simonbyford:
Co-authored-by: <ravi.lal.freelancer@guardian.co.uk>

## What does this change?

- removes the parameter `countryCode` from `buildPageTargeting()`. The country code is instead calculated directly using a script [copied from `libs`](https://github.com/guardian/csnx/blob/main/libs/%40guardian/libs/src/locale/getLocale.ts) and modified to be synchronous
- adds a spec for `buildPageTargeting`. This was based on an [existing spec in frontend](https://github.com/guardian/frontend/blob/main/static/src/javascripts/projects/common/modules/commercial/build-page-targeting.spec.ts) which can now be removed

## Why?

Fewer arguments mean the method is simpler to consume.

The spec always belonged here and not in frontend, [as acknowledged at the time](https://github.com/guardian/frontend/pull/25556#discussion_r988886343).

### DIFF
--- a/src/create-ad-slot.spec.ts
+++ b/src/create-ad-slot.spec.ts
@@ -27,6 +27,7 @@ const DEFAULT_CONFIG = {
 		isSensitive: false,
 		section: 'uk-news',
 		videoDuration: 63,
+		edition: 'UK' as const,
 	},
 };
 
--- a/src/lib/get-locale.ts
+++ b/src/lib/get-locale.ts
@@ -1,0 +1,54 @@
+import type { CountryCode } from '@guardian/libs';
+import { getCookie, isString, storage } from '@guardian/libs';
+import type { Edition } from '../types';
+
+const KEY = 'GU_geo_country';
+const KEY_OVERRIDE = 'gu.geo.override';
+const COUNTRY_REGEX = /^[A-Z]{2}$/;
+
+// best guess that we have a valid code, without actually shipping the entire list
+const isValidCountryCode = (country: unknown): country is CountryCode =>
+	isString(country) && COUNTRY_REGEX.test(country);
+
+// we'll cache any successful lookups so we only have to do this once
+let locale: CountryCode | undefined;
+
+const editionToGeolocationMap: Record<Edition, CountryCode> = {
+	UK: 'GB',
+	US: 'US',
+	AU: 'AU',
+};
+
+const editionToGeolocation = (editionKey: Edition): CountryCode =>
+	editionToGeolocationMap[editionKey];
+
+// just used for tests
+export const __resetCachedValue = (): void => (locale = undefined);
+
+/**
+ * Fetches the user's current location as an ISO 3166-1 alpha-2 string e.g. 'GB', 'AU' etc
+ * Note: This has been copied from guardian-libs and made syncronous by ommiting the call to
+ * the geolocation API
+ */
+export const getLocale = (): CountryCode => {
+	if (locale) return locale;
+
+	// return overridden geo from localStorage, used for changing geo only for development purposes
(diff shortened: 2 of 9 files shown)


src/create-ad-slot.spec.ts
```diff
@@ -27,6 +28,7 @@ const DEFAULT_CONFIG = {
 		isSensitive: false,
 		section: 'uk-news',
 		videoDuration: 63,
+		edition: 'UK' as Edition,
```
chrislomaxjones, line 31: I think you could type these `as const` and TypeScript will figure out that this is a valid subtype of `Edition`. I'll leave that up to you though, I'm not sure there's very much in it.

```suggestion
		edition: 'UK' as const,
```

src/lib/get-locale.ts
```diff
@@ -0,0 +1,55 @@
+import type { CountryCode } from '@guardian/libs';
+import { getCookie, isString, storage } from '@guardian/libs';
+import type { Edition } from '../types';
+
+const KEY = 'GU_geo_country';
+const KEY_OVERRIDE = 'gu.geo.override';
+const COUNTRY_REGEX = /^[A-Z]{2}$/;
+
+// best guess that we have a valid code, without actually shipping the entire list
+const isValidCountryCode = (country: unknown) =>
```
chrislomaxjones, line 10: Could we make this a type guard? Might be able to avoid the `as CountryCode` assertion on L46 this way.

```suggestion
const isValidCountryCode = (country: unknown): country is CountryCode =>
```

src/lib/get-locale.ts
```diff
@@ -0,0 +1,55 @@
...
+
+// just used for tests
+export const __resetCachedValue = (): void => (locale = void 0);
+
+/**
+ * Fetches the user's current location as an ISO 3166-1 alpha-2 string e.g. 'GB', 'AU' etc
+ * Note: This has been copied from guardian-libs and made syncronous by ommiting the call to
+ * the geolocation API
+ */
+
```
chrislomaxjones, line 33: Can remove this newline here?

```suggestion
```

src/lib/get-locale.ts
```diff
@@ -0,0 +1,55 @@
...
+	UK: 'GB',
+	US: 'US',
+	AU: 'AU',
+};
+
+const editionToGeolocation = (editionKey: Edition): CountryCode =>
+	editionToGeolocationMap[editionKey];
+
+// just used for tests
+export const __resetCachedValue = (): void => (locale = void 0);
```
chrislomaxjones, line 26: It'd be nice to avoid defining and exporting a function used just for testing, but I can't think of a way around it in this instance!

A bit of a nitpick but I always find `void 0` a bit confusing, is this just acting as `undefined` here?

chrislomaxjones (commented):
Thanks for bringing all the tests over 🙏

src/create-ad-slot.spec.ts
simonbyford: You're right, thanks! I've made the change

src/lib/get-locale.ts
simonbyford: Yup, have changed

src/lib/get-locale.ts
simonbyford: Yes, whoopsy

src/lib/get-locale.ts
simonbyford: Agreed on the unfortunate need to expose this function!

Also agreed on `void 0`, have updated

chrislomaxjones (approved):
Awesome!

PR #695, opened 2022-10-19
test: Add test for `buildPageTargetingConsentless`

simonbyford:
## What does this change?

Adds a test for `buildPageTargetingConsentless()`

## Why?

There wasn't one before


src/targeting/build-page-targeting-consentless.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+import { buildPageTargeting as buildPageTargeting_ } from './build-page-targeting';
+import { buildPageTargetingConsentless } from './build-page-targeting-consentless';
+
+const buildPageTargeting = buildPageTargeting_ as jest.MockedFunction<
+	typeof buildPageTargeting_
+>;
+
+jest.mock('./build-page-targeting', () => ({
+	buildPageTargeting: jest.fn(),
+}));
```
arelra, line 11: The way this is mocked is a bit ye oldy worldy which we've carried over from Frontend.

I _think_ we don't need to alias the function or stub the mock but please test:

```suggestion
import { buildPageTargeting } from './build-page-targeting';
import { buildPageTargetingConsentless } from './build-page-targeting-consentless';

jest.mock('./build-page-targeting');
```

And then later:

```js
(buildPageTargeting as jest.Mock).mockReturnValue
```

arelra (approved):
Thanks for fixing this gap in our tests! 📈

src/targeting/build-page-targeting-consentless.spec.ts
simonbyford: Ahh miles better. This was pure cargo culting on my part. Thanks, have updated!

PR #698, opened 2022-10-25
feat: Vendor scripts from Frontend

chrislomaxjones:
## What does this change?

Add the following third party scripts from Frontend to the `__vendor` directory:
- Ipsos Mori stub
- A9 Apstag
- Launchpad (used by Redplanet)
- Pubmatic override (for using Permutive audiences with Pubmatic Prebid)

Note that for A9 and Launchpad we wrap the stubs in `() => void` functions, so we can defer their execution without using a dynamic import.

## Why?

Collect more of our vendor third party code in one place. Make it clearer that this should remain as JavaScript.

### DIFF
--- a/src/__vendor/a9-apstag.js
+++ b/src/__vendor/a9-apstag.js
@@ -1,0 +1,36 @@
+/*eslint-disable */
+// Load Amazon A9 library as described here: https://ams.amazon.com/webpublisher/uam/docs/web-integration-documentation/integration-guide/javascript-guide/display.html
+export const a9Apstag = () => {
+	!(function(a9, a, p, s, t, A, g) {
+		if (a[a9]) return;
+		function q(c, r) {
+			a[a9]._Q.push([c, r]);
+		}
+		a[a9] = {
+			init() {
+				q('i', arguments);
+			},
+			fetchBids() {
+				q('f', arguments);
+			},
+			setDisplayBids() {},
+			targetingKeys() {
+				return [];
+			},
+			_Q: [],
+		};
+		A = p.createElement(s);
+		A.async = !0;
+		A.src = t;
+		g = p.getElementsByTagName(s)[0];
+
+		g.parentNode.insertBefore(A, g);
+	})(
+		'apstag',
+		window,
+		document,
+		'script',
+		'//c.amazon-adsystem.com/aax2/apstag.js'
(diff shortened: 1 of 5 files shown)


src/__vendor/a9-apstag.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * Load Amazon A9 library as {@link https://ams.amazon.com/webpublisher/uam/docs/web-integration-documentation/integration-guide/javascript-guide/display.html described here}
+ */
+export const a9Apstag = () => {
+	!(function(a9, a, p, s, t, A, g) {
```
arelra, line 5: Don't think the `!` is doing anything useful here?:
```suggestion
	(function(a9, a, p, s, t, A, g) {
```

src/__vendor/launchpad.js
```diff
@@ -0,0 +1,23 @@
+/**
+ *  Loads the Launchpad js tracker (used by Redplanet) into the page,
+ *  and creates the global "launchpad" method
+ */
+export const launchpad = () => {
+	;(function (p, l, o, w, i, n, g) {
```
arelra, line 6: Don't think the `;` is doing anything useful here:

```suggestion
	(function (p, l, o, w, i, n, g) {
```

src/__vendor/a9-apstag.js
chrislomaxjones: Nice, I'm pretty sure we shouldn't ever be returning (as we're not here) and using the return value of these stubs regardless, so I agree. I wonder why this and the `;` in your other comment were there originally 🤔

Jakeii: I think the `;` was an old best practise to do with minifying/concatenating scripts, just in case the script before yours forgot it's final `;` it wouldn't break everything. Possibly a similar story for the `!` 🤔

Jakeii (approved):

src/__vendor/a9-apstag.js
arelra: I think the `!`  was an old byte-saving IIFE trick so you could do `!function(){}()` without wrapping the function and it would invoke. 

But given we have `(function(){})()` I think it's not needed regardless.

arelra (approved):

PR #699, opened 2022-10-25
feat: Remove `clientSideParticipations` parameter from `buildPageTargeting`

simonbyford:
## What does this change?

Removes the parameter `clientSideParticipations` from `buildPageTargeting()`. The `clientSideParticipations` is instead calculated directly from localstorage.

### DIFF
--- a/src/targeting/build-page-targeting.spec.ts
+++ b/src/targeting/build-page-targeting.spec.ts
@@ -171,7 +171,7 @@ describe('Build Page Targeting', () => {
 	});
 
 	it('should build correct page targeting', () => {
-		const pageTargeting = buildPageTargeting(emptyConsent, false, {});
+		const pageTargeting = buildPageTargeting(emptyConsent, false);
 
 		expect(pageTargeting.sens).toBe('f');
 		expect(pageTargeting.edition).toBe('us');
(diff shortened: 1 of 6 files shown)


src/lib/get-participations.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const isParticipations = (
+	participations: unknown,
+): participations is Participations => {
+	if (!isObject(participations)) return false;
+	return Object.values(participations).some((participation) => {
+		if (!isObject(participation)) return false;
+		if (!isString(participation.variant)) return false;
+		return true;
+	});
```
chrislomaxjones, line 14: This may come down to taste, but I personally find the early returns add some noise here.

Could the same logic be expressed by checking that `every` participation matches the criteria?

```suggestion
	isObject(participations) &&
	Object.values(participations).every(
		(participation) =>
			isObject(participation) && isString(participation.variant),
	);
```

I'll leave this up to you though!

src/lib/get-participations.ts
```diff
@@ -0,0 +1,20 @@
+import type { Participations } from '@guardian/ab-core';
+import { isObject, isString, storage } from '@guardian/libs';
+
+const participationsKey = 'gu.ab._participations';
```
chrislomaxjones, line 4: Was there always an `_` in the key for this? Wasn't this originally `gu.ab.participations`?

src/lib/get-participations.ts
```diff
@@ -0,0 +1,20 @@
...
+): participations is Participations => {
+	if (!isObject(participations)) return false;
+	return Object.values(participations).some((participation) => {
+		if (!isObject(participation)) return false;
+		if (!isString(participation.variant)) return false;
+		return true;
+	});
+};
+
+export const getParticipationsFromLocalStorage = (): Participations => {
```
arelra, line 17: Minor:

I know we lifted this file from dotcom's `ab-localstorage.ts`. I prefer that naming for a couple of reasons:
- clear that this module is something to do with ab testing
- it is a (albeit) weak link to that module in DCR which will make it easier to track should we ever migrate this functionality to libs

arelra, line 17: Just to increase our understanding of this could we check if localstorage is set with all the tests the user participates in. Also if the localstorage participations includes server side and client side tests. Thank you!

arelra (approved):
Nice! Another step forwards to having all our targeting in commercial-core 🎉

src/lib/get-participations.ts
simonbyford: Yup agreed, I'll make the change

src/lib/get-participations.ts
simonbyford: Hard to know what the original intention was because this case [isn't tested in DCR where this code was lifted from](https://github.com/guardian/dotcom-rendering/blob/67b249e29c8c45e72f5869483cbd658c1cb300ff/dotcom-rendering/src/web/experiments/lib/ab-localstorage.test.ts)

src/lib/get-participations.ts
simonbyford: It seems to differ between frontend and DCR:

https://github.com/guardian/dotcom-rendering/pull/6136#issuecomment-1270495167

src/lib/get-participations.ts
simonbyford: Yes definitely. Do you mean test on CODE? Or add unit tests here?

src/lib/get-participations.ts
simonbyford: I don't understand why, or which we should care about here. Will do some digging to better understand how participations are stored. Thanks for this prompt!

src/lib/get-participations.ts
chrislomaxjones: So [this is the type](https://github.com/guardian/ab-testing/blob/b5de0e09259d99399f192e4696470812b2aa45b1/packages/ab-core/src/types.ts#L99) that we're testing against I think, which looks like every variant value must be a string.

src/lib/get-participations.ts
arelra: Oh I just meant testing on CODE

src/lib/get-participations.ts
simonbyford: You're right, I'll update to the one you suggested, thanks

src/lib/get-participations.ts
chrislomaxjones: As I understand it, Frontend used `gu.ab.participations` and saves a participation in local storage whenever a browser is entered into a test, and this logic wasn't present on DCR. This logic was added to DCR for the long Mandatory Sign In Gate test in order to bucket people and then stop adding new people to buckets after a certain duration. However, I think because of a naming collision this needed to be called something other than `gu.ab.participations`, hence the underscore.

I think from a commercial perspective the only tests we're interesting in targeting against are the traditional Frontend ones, so `gu.ab.participations`. I'm not sure that the `_` ones are ever exposed to the commercial code.

simonbyford: Great, thanks Chris, that was the conclusion I just reached too. I've updated to use `gu.ab.participations`

src/lib/get-participations.ts
simonbyford: Confirmed localstorage participations is set with client-side tests _only_

arelra: Great thanks for checking that!

src/targeting/build-page-targeting.spec.ts
```diff
@@ -192,103 +192,84 @@ describe('Build Page Targeting', () => {
...
 
-	it('should set correct ab param', () => {
-		expect(
-			buildPageTargeting(emptyConsent, false, {
-				MtMaster: {
-					variant: 'variantName',
-				},
-			}).ab,
-		).toEqual(['MtMaster-variantName']);
+	it.only('should set correct ab param', () => {
```
mxdvl, line 253: Aren’t you skipping all other tests with this `only`?

PR #701, opened 2022-10-27
chore: Remove npm suggestion from README

Jakeii:
## What does this change?
Remove npm install suggestion from readme

## Why?
We shouldn't encourage using a package manager other than yarn as the `yarn.lock` file won't be read or written to, and could cause inconsistent dependencies.


chrislomaxjones (approved):
Good shout 👍

PR #703, opened 2022-10-31
feat: YouTube IMA AdTag builder

arelra:
## What does this change?

Adds a utility function `buildImaAdTagUrl` for our YouTube IMA ads integration.

Once we have full targeting in commercial-core (coming soon) we will update to use that. For the moment this is a foundational piece of work to shift the building of the ad tag from DCR into commercial-core

## Why?

This will construct an ad tag URL to be used by the YouTube IMA ads integration.

### DIFF
--- a/src/targeting/youtube-ima.ts
+++ b/src/targeting/youtube-ima.ts
@@ -1,24 +1,18 @@
-import type { Participations } from '@guardian/ab-core';
 import type { CustomParams, MaybeArray } from '../types';
-import { filterEmptyValues } from './build-page-targeting';
+import { filterValues } from './build-page-targeting';
 
 /**
  * @param  {Record<string, MaybeArray<string|number|boolean>>
- * do this https://support.google.com/admanager/answer/1080597
+ * Follows https://support.google.com/admanager/answer/1080597
  */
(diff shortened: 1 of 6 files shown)


src/targeting/youtube-ima.ts
```diff
@@ -0,0 +1,60 @@
...
+};
+
+const buildImaAdTagUrl = (
+	adUnit: string,
+	customParams: CustomParams,
+): string => {
+	// TODO: Add regular build-page-targeting to
+	// get all targeting including ab participations.
+	// For now hardcode adtest (at)
+	const mergedCustomParams = { ...customParams, at: 'fixed-puppies' };
```
zekehuntergreen, line 32: Are we not ready to remove the ad test parameter to show real ads?

arelra: Not yet but close - we have a separate PR in progress to derive `ab-participations` in commercial-core: #699 

I thought we could release this first and enhance as other targeting gets rolled into commercial-core. It should work the same as now - the test has not started yet as we are waiting for a bugfix from Google.

zekehuntergreen (approved):

PR #704, opened 2022-10-31
chore: Bump deps 2022.10.31

arelra:
## What does this change?

Bump deps 2022.10.31

### DIFF
--- a/src/event-timer.ts
+++ b/src/event-timer.ts
@@ -33,7 +33,7 @@ interface PageEventStatus {
 	commercialModulesLoaded: boolean;
 }
 
-type ConnectionType = "bluetooth" | "cellular" | "ethernet" | "mixed" | "none" | "other" | "unknown" | "wifi";
+type ConnectionType = 'bluetooth' | 'cellular' | 'ethernet' | 'mixed' | 'none' | 'other' | 'unknown' | 'wifi';
 
 interface NetworkInformation extends EventTarget {
     readonly type?: ConnectionType;
--- a/src/event-timer.ts
+++ b/src/event-timer.ts
@@ -33,24 +33,6 @@ interface PageEventStatus {
 	commercialModulesLoaded: boolean;
 }
 
-type ConnectionType = 'bluetooth' | 'cellular' | 'ethernet' | 'mixed' | 'none' | 'other' | 'unknown' | 'wifi';
-
-interface NetworkInformation extends EventTarget {
-    readonly type?: ConnectionType;
-	readonly downlink?: number;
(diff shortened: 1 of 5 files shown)


src/event-timer.ts
```diff
@@ -161,20 +179,21 @@ export class EventTimer {
 				},
 			],
 		};
+		const localWindow = window as unknown as LocalWindow;
```
arelra, line 182: Bumping TypeScript from v4.7.4 to v4.8.4 means we lose the `NavigatorNetworkInformation` type as it is not supported by the majority of browsers.

Not so happy about this fix. Overriding the window.navigator type does not seem possible as you then run into:

`Subsequent property declarations must have the same type` errors for `lib.dom.d.ts`

Is there a better solution?

chrislomaxjones: Could you try overwriting the `Navigator` type in the global scope directly? Something like

```typescript
declare global {
	interface Navigator {
		readonly connection: NetworkInformation;
	}
}
```

src/event-timer.ts
```diff
@@ -33,6 +33,24 @@ interface PageEventStatus {
 	commercialModulesLoaded: boolean;
 }
 
+type ConnectionType = "bluetooth" | "cellular" | "ethernet" | "mixed" | "none" | "other" | "unknown" | "wifi";
```
chrislomaxjones, line 36: I thought double quotes would get flagged by Prettier?

src/event-timer.ts
arelra: Nice idea thanks! I've tried it out but that still doesn't override `Navigator` from `lib.dom.d.ts` and you get a similar error message:

`All declarations of 'navigator' must have identical modifiers`

I shall pick your TS brain IRL :)

src/event-timer.ts
arelra: That is surprising. It isn't flagging and CI passes.

src/event-timer.ts
chrislomaxjones: I've got a [branch here](https://github.com/guardian/commercial-core/compare/test-navigator-types) with a minimal reproduction, but am not too sure it's exactly what we're after, so happy to discuss!

src/event-timer.ts
chrislomaxjones: Strange, I now wonder if we ever actually check Prettier in CI, or just rely on people auto-formatting on save 🤔

chrislomaxjones (approved):
Always keeping on top of dependencies @arelra 👍

PR #708, opened 2022-11-07
feat: add `youtube` parameter to `buildPageTargeting`

simonbyford:
## What does this change?

When `youtube` is set to `false`, `buildPageTargeting` pulls the regular permutive segments (i.e. those stored in localstorage under the key `_papns`). This was the default behaviour before this change.

When `youtube` is set to `true`, `buildPageTargeting` pulls the youtube-specific permutive segments (i.e. those stored in localstorage under the key `_pdfps`)

## Why?

So we can finally start using `buildPageTargeting` in `buildAdsConfig`, which happens in [this commit](https://github.com/guardian/commercial-core/pull/708/commits/260f96ff78ab9d4789c64ad4be05bc5e25e8b5d1).

### DIFF
--- a/src/targeting/build-page-targeting.ts
+++ b/src/targeting/build-page-targeting.ts
@@ -68,10 +68,17 @@ const filterValues = (pageTargets: Record<string, unknown>) => {
 	return filtered;
 };
 
-const buildPageTargeting = (
-	consentState: ConsentState,
-	adFree: boolean,
-): PageTargeting => {
+type BuildPageTargetingParams = {
+	consentState: ConsentState;
+	adFree: boolean;
+	youtube?: boolean;
+};
+
+const buildPageTargeting = ({
+	consentState,
+	adFree,
+	youtube = false,
+}: BuildPageTargetingParams): Record<string, string | string[]> => {
 	const { page, isDotcomRendering } = window.guardian.config;
 
 	const adFreeTargeting: { af?: True } = adFree ? { af: 't' } : {};
@@ -120,7 +127,10 @@ const buildPageTargeting = (
 		? getSharedTargeting(page.sharedAdTargeting)
 		: {};
 
-	const personalisedTargeting = getPersonalisedTargeting(consentState);
+	const personalisedTargeting = getPersonalisedTargeting({
+		state: consentState,
+		youtube,
+	});
 
 	const pageTargets: PageTargeting = {
 		...personalisedTargeting,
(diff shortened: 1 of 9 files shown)


src/targeting/build-page-targeting.ts
```diff
@@ -71,7 +71,8 @@ const filterValues = (pageTargets: Record<string, unknown>) => {
 const buildPageTargeting = (
 	consentState: ConsentState,
 	adFree: boolean,
-): PageTargeting => {
+	pfp: boolean,
```
arelra, line 74: If we make this have a default value of false then we don't need to pass this wherever we don't need youtube targeting?

```suggestion
	youtube = false,
```

src/ad-targeting-youtube.ts
```diff
@@ -14,22 +12,14 @@ import type {
...
-
 const buildAdsConfig = (
 	cmpConsent: ConsentState,
 	adUnit: string,
 	customParams: CustomParams,
 ): AdsConfig => {
 	const mergedCustomParams = {
 		...customParams,
-		...buildCustomParamsFromCookies(),
+		...buildPageTargeting(cmpConsent, false, true),
```
arelra, line 22: Minor organisational if you agree - could we move this file and its test into `/targeting` pls and rename it to `youtube` to align with the other files.

arelra (commented):
Great work, nearly at the end of this epic!

Just some minor comments

src/targeting/build-page-targeting.ts
```diff
@@ -71,7 +71,8 @@ const filterValues = (pageTargets: Record<string, unknown>) => {
...
+): Record<string, string | string[]> => {
```
chrislomaxjones, line 75: Could we still use the `PageTargeting` type here?

src/targeting/build-page-targeting.ts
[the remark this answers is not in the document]
chrislomaxjones: On a similar topic, now this function takes two boolean arguments, I was wondering if this would benefit from an object of named parameters?

Calling something like

```typescript
buildPageTargeting({ consentState, adFree: true, youtube: false });
```

instead of

```typescript
buildPageTargeting(consentState, true, false);
```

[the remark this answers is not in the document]
simonbyford: Agreed on both fronts, have made the change!

arelra (approved):
Nice work!

PR #709, opened 2022-11-08
feat: YouTube IMA ad tag URL uses page targeting

arelra:
## What does this change?

Updates `youtube-ima` to use  regular `getPageTargeting` in order to build a full targeting object to pass to GAM.

Primarily concerned with adtest (at) and ab test participations (ab) but the other targeting will be useful when IMA ads are up and running.

## Why?

We can target YouTube IMA ads more effectively.


chrislomaxjones (approved):
Looks good 👍

Jakeii (approved):

PR #710, opened 2022-11-10
feat: YouTube IMA handles exceptions from build-page-targeting

arelra:
## What does this change?

YouTube IMA handles exceptions from build-page-targeting

## Why?

Defensive error handling in case YoutubeAtom and therefore `buildImaAdTagUrl` is used in an environment where `guardian.config`, cookies, localstorage etc are not available.

### DIFF
--- a/src/targeting/youtube-ima.ts
+++ b/src/targeting/youtube-ima.ts
@@ -1,4 +1,5 @@
 import type { ConsentState } from '@guardian/consent-management-platform/dist/types';
+import { log } from '@guardian/libs';
 import type { CustomParams, MaybeArray } from '../types';
 import { buildPageTargeting, filterValues } from './build-page-targeting';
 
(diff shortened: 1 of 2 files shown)


src/targeting/youtube-ima.ts
```diff
@@ -20,14 +21,34 @@ const encodeCustomParams = (
...
+		return mergedCustomParams;
+	} catch (e) {
+		/**
+		 * Defensive error handling in case YoutubeAtom is used in an
+		 * environment where guardian.config, cookies, localstorage etc
+		 * are not available
+		 */
+		log('commercial', 'Error building YouTube IMA custom params', e);
+	}
+	return customParams;
```
domlander, line 40: Does the following make it clearer that it's specifically `buildPageTargeting(...)` that can throw an exception?

```
try {
		const pageTargeting = buildPageTargeting(consentState, false);
	} catch (e) {
		/**
		 * Defensive error handling in case YoutubeAtom is used in an
		 * environment where guardian.config, cookies, localstorage etc
		 * are not available
		 */
		log('commercial', 'Error building YouTube IMA custom params', e);
		return customParams;
	}

	const mergedCustomParams = { ...customParams, ...pageTargeting };
	return mergedCustomParams;
};
```

domlander (approved):

src/targeting/youtube-ima.ts
arelra: That's much clearer. Thanks!

PR #712, opened 2022-11-18
feat: Pass client side test participations to targeting functions

arelra:
## What does this change?

- adds back `clientSideParticipations` as a parameter to `build-page-targeting`.
- updates public YouTube targeting functions to accept an object rather than an array of parameters

## Why?

We previously thought we could read test participations from localstorage. This works on Frontend but the [library that DCR uses](https://github.com/guardian/ab-testing) for ab testing [does not write to localstorage](https://github.com/guardian/ab-testing#differences-of-this-library-vs-frontend-implementation).


chrislomaxjones (approved):
Looks good 👍

PR #718, opened 2022-11-25
feat: Creates logic for creating server side ad slots on liveblogs

domlander:
## What does this change?

Creates two functions that contain much of the logic for server-side ad slot insertion on liveblog pages.

## Why?

So that we can create ad slots server side whilst providing a good UX and maintain current ad revenue from liveblog pages. This is commercial code so ideally should live in the commercial-core repo.


domlander:
I could not find a way to untangle the types in DCR in a way that made it pragmatic to extract the logic out in to this repository. Therefore the logic still resides in DCR

PR #722, opened 2022-11-29
chore(deps-dev): dependencies bump `2022-11-29`

chrislomaxjones:
## What does this change?

Bump dependencies for this week.

Cherry pick 4 Dependabot PRs, plus upgrade `@guardian/libs` to `10.0.0` so that we can remove our development dependency on `web-vitals` (this was required to run the test since it used to be a peer dep of libs).

~The remaining TypeScript upgrade will require a change to our TS setup, so will be addressed separately.~

The upgrade to TypeScript 4.9.3 required we pull in `@types/node` explicity - see [this dicsussion](https://github.com/microsoft/TypeScript/issues/51567) for more context.

## Why?

Keep our dependencies up to date.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,14 +74,13 @@
     "semantic-release": "^19.0.2",
     "ts-jest": "^27.0.7",
     "type-fest": "^2.8.0",
-    "typescript": "^4.5.2",
-    "web-vitals": "^2.1.2"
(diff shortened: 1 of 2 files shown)


arelra (approved):
😍

PR #730, opened 2022-12-05
chore: Setup for bundling in frontend by moving to monorepo

Jakeii:
## What does this change?

Setup workspaces/monorepo in preparation for moving frontend commercial bundle source here under the `standalone` directory and `@guardian/commercial-standalone` package

The current commercial-core will be in the `core` directory and remain under the `@guardian/commercial-core` package (why this is a chore commit)

The frontend ts/eslint/etc config has been copied from frontend into `standalone`, webpack and package.json have been pruned and copied.

Added a `scripts/frontend-migration.mjs` script that will graph the frontend commercial bundle and copy it to `standalone/src`.

The frontend code is not in the repo yet as there is still a little work to do including:

- Cleanup/optimise standalone webpack config
- Get cypress tests in and working in CI and locally
- Modify script to get non bundle files in (spec.ts, d.ts, mocks etc)
- Ensure all CI we expect for frontend code works in commercial-core
- Modify frontend to use bundle from commercial core

## Why?
BUNDLEPENDANCY


standalone/tsconfig.json
```diff
@@ -0,0 +1,71 @@
...
+      "es2015",
+      "dom",
+      "dom.iterable"
+    ],
+    "module": "ESNext",
+    "moduleResolution": "node",
+    "noEmit": true,
+    "noUnusedLocals": true,
+    "paths": {
+      "admin/*": [
```
chrislomaxjones, line 20: We can probably delete most of these paths (maybe keep `common`, `commercial`, and a few others?)

standalone/package.json
```diff
@@ -0,0 +1,89 @@
...
+		"webpack-merge": "^4.1.0"
+	},
+	"publishConfig": {
+		"access": "public"
+	},
+	"peerDependencies": {
+		"@guardian/ab-core": "^2.0.0",
+		"@guardian/libs": ">6"
+	},
+	"dependencies": {
```
chrislomaxjones, line 77: This is a comfortingly small number of runtime dependencies :)

Jakeii:
> Great work, this is very exciting!
> 
> I'm just collecting some thoughts here on future steps:
> 
> * Since unit tests aren't captured in the commercial bundle graph, they'll have to be copied over as a separate step.
> * Ditto with TypeScript declaration files? And some `__mocks__` directories that are used by some tests.
> * Cypress directory and Github action as well (in my opinion running the built output of these bundles against DCR in this repo makes tons of sense compared to in Frontend).
> * Removing as much cruft as possible from the Webpack/TypeScript/Eslint configuration.
> 
> What do you propose as next steps for this PR in particular? Shall we hang off merging into main until we are ready to making this the "commercial source of truth"?

great point about the unit tests, might be a way to wrangle `madge` into grabbing additional things we want like unit tests, type declarations and the mocks like you've said🤔
I completely forgot about cypress, the github action should probably work with a couple of tweaks as well

I think we could divide this and the webpack tweaks, into a few tickets that can be merged here to move this forward!🚀

README.md
```diff
@@ -1,58 +1,16 @@
...
-If necessary, it will then automatically release a new, [semver](https://semver.org/)-compliant version of the package to NPM.
+# @guardian/commercial
 
-#### Pull requests
+Monorepo for The Guardian's commercial code
 
-Try to write PR titles in the conventional commit format, and [squash and merge](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges#squash-and-merge-your-pull-request-commits) when merging. That way your PR will trigger a release when you merge it (if necessary).
+## Packages
+* [@guardian/core](./core) - The guardian advertising libs
+* [@guardian/standalone](./standalone/) - The bundle that provides the advertising logic for the https://theguardian.com
```
SiAdcock, line 16: Forgive me if daft question... is the intention for `@guardian/standalone` to be a published package on NPM? Or will it live as a standalone bundle in S3 to be requested by DCR / Frontend at runtime?

core/README.md
```diff
@@ -0,0 +1,58 @@
...
+If your target environment does not support that, make sure you transpile this package when bundling your application.
+
+## Development
+
+### Requirements
+
+1. [Node 14](https://nodejs.org/en/download/) ([nvm][] or [fnm][] recommended)
+2. [Yarn](https://classic.yarnpkg.com/en/docs/install/)
+
+[nvm]: https://github.com/nvm-sh/nvm
```
SiAdcock, line 43: I think we can stop recommending nvm. fnm is the way (or asdf for polyglots)

.github/workflows/CI-commercial-core.yml
```diff
@@ -121,8 +128,11 @@ jobs:
           path: dist
 
       - name: Release
+        uses: cycjimmy/semantic-release-action@v2
```
arelra, line 131: Would it be preferable to use the GHA from semantic-release themselves (assuming it does the same thing):

https://github.com/semantic-release/github

scripts/frontend-migration.mjs
```diff
@@ -0,0 +1,51 @@
+
```
arelra, line 1: pedantry- blank line :)

package.json
```diff
@@ -10,87 +10,27 @@
...
-		"test:watch": "jest --watch",
-		"test:ci": "jest --coverage --ci",
-		"tsc": "tsc --noEmit",
-		"validate": "npm-run-all tsc lint test build",
-		"prepare": "husky install"
+		"prepush:test": "yarn workspace @guardian/commercial-core jest --verbose --runInBand --onlyChanged"
 	},
+	"workspaces": [
+		"core",
+		"standalone"
```
arelra, line 19: I'm not super keen on the name 'standalone' as it originates from when we used to have separate bundles for Frontend and DCR. But we can think of better names later.

arelra (approved):
This is fantastic work @Jakeii  - a big step forwards in migrating out of Frontend 👏 👏 👏

standalone/package.json
```diff
@@ -0,0 +1,90 @@
...
+		"@guardian/source-foundations": "^7.0.1",
+		"fastdom": "^1.0.11",
+		"lodash-es": "^4.17.21",
+		"ophan-tracker-js": "^1.4.0",
+		"prebid.js": "guardian/prebid.js#2e3b96d",
+		"process": "^0.11.10",
+		"raven-js": "^3.27.2",
+		"raw-loader": "^4.0.2",
+		"svg-loader": "^0.0.2",
+		"wolfy87-eventemitter": "^5.2.9"
```
SiAdcock, line 88: Would this result in DCR / Frontend shipping 2 instances of common dependencies such as source-foundations, fastdom, ophan, raven, etc – once in the commercial bundle and once in the standard JS for these platforms? Perhaps this is already happening, I'm not sure

SiAdcock (approved):
Heroic effort 💪 Great work @Jakeii!

README.md
Jakeii: As an interim solution, we'll be publishing the bundle in a package that is used by frontend, until we figure out a completely independent solution where we publish to a URL.

package.json
Jakeii: Agreed!

.github/workflows/CI-commercial-core.yml
Jakeii: I don't think that's a GHA you've linked? And I can't find an official one, TBH I can't see why I switched to the action from just running the command! I'll switch it back..

PR #735, opened 2022-12-14
feat: Add `320x480` ad size

chrislomaxjones:
## What does this change?

Add the `320x480` size to our set of known ad sizes.

A small amount of drive-by refactoring as well: sorting the size keys, adding some short comments, and using `as const` in the test.

## Why?

We'll need this size when we enable Picnic on mobile web.

### DIFF
--- a/src/ad-sizes.ts
+++ b/src/ad-sizes.ts
@@ -84,8 +84,8 @@ type SizeKeys =
 	| 'outstreamDesktop'
 	| 'outstreamGoogleDesktop'
 	| 'outstreamMobile'
-	| 'picnic'
 	| 'portrait'
+	| 'portraitInterstitial'
 	| 'skyscraper';
 
 type SlotName =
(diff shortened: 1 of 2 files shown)


arelra (approved):

src/ad-sizes.ts
```diff
@@ -83,9 +84,9 @@ type SizeKeys =
 	| 'outstreamDesktop'
 	| 'outstreamGoogleDesktop'
 	| 'outstreamMobile'
+	| 'picnic'
```
Amouzle, line 87: Could we keep this just as the dimensions? As picnic might be one ad that runs but others could in the future I.e. Google Native / house ads for example.

chrislomaxjones: Sure! It looks like we give the dimension-based ones a name first. From a bit of googling should it be a [portrait interstitial](https://www.dimensions.com/element/mobile-portrait-interstitial-ad-320-x-480)?

Amouzle: Okay TIL! Makes sense.

PR #748, opened 2023-01-04
fix: Use unified CMP event names

chrislomaxjones:
## What does this change?

This changes the set of externally defined event names, removing all of the existing per-framework names and using ones that are framework-agnostic.

The three new event names:
- [`cmp-init`](https://github.com/search?q=repo%3Aguardian%2Fconsent-management-platform+cmp-init&type=code)
- [`cmp-ui-displayed`](https://github.com/search?q=repo%3Aguardian%2Fconsent-management-platform+cmp-ui-displayed&type=code)
- [`cmp-got-consent`](https://github.com/search?q=repo%3Aguardian%2Fconsent-management-platform+cmp-got-consent&type=code)

## Why?

Some time ago the Consent Management Platform moved over to a ['unified implementation'](https://github.com/guardian/consent-management-platform/pull/561). In doing so, we stopped leaving the same unique performance marks for each of the three CMP frameworks. This PR moves commercial-core to start looking up the new correct marks, so we add them to our commercial metrics payload correctly.

Note there we will be some work required from the data-side to reconfigure the dashboard to use these measurements.

### DIFF
--- a/src/event-timer.spec.ts
+++ b/src/event-timer.spec.ts
@@ -38,8 +38,8 @@ const performance = {
 
 const MARK_NAME = 'mark_name';
 const MARK_LONG_NAME = `gu.commercial.${MARK_NAME}`;
-const CMP_INIT = 'cmp-tcfv2-init';
-const CMP_GOT_CONSENT = 'cmp-tcfv2-got-consent';
+const CMP_INIT = 'cmp-init';
+const CMP_GOT_CONSENT = 'cmp-got-consent';
 
 const DEFAULT_CONFIG = {
 	isDotcomRendering: true,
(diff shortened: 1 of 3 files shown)


domlander (commented):
Should this test be updated?

https://github.com/guardian/commercial-core/blob/b9dbf7b0da5f9c5cfa1310192b898878733256f3/src/send-commercial-metrics.spec.ts#L545

chrislomaxjones:
> Should this test be updated?
> 
> https://github.com/guardian/commercial-core/blob/b9dbf7b0da5f9c5cfa1310192b898878733256f3/src/send-commercial-metrics.spec.ts#L545-L559

Great spot thanks. I've put a dummy metric in since it isn't really necessary for the test.

domlander (approved):
Looks good

PR #758, opened 2023-01-17
chore: Include additional files in migration script like tests and get them working

Jakeii:
## What does this change?
Adds files necessary to get unit tests working, spec files, type definitions, and mocks.

I also discovered a bug with the script/madge library, it only seems to find all the files when the script/madge is executed with frontend as the root folder.

There are still a few failing bundle tests when they're run but they're due to the TS version (changes to the window type in lib.dom.d.ts), when we do the final copy, the tests will need to be fixed.


standalone/package.json
```diff
@@ -75,7 +98,6 @@
 		"@guardian/libs": ">6"
 	},
 	"dependencies": {
-		"@guardian/commercial-core": "^4.25.0",
```
chrislomaxjones, line 78: Should commercial-core still be a dependency rather than a development dependency?

standalone/tsconfig.json
```diff
@@ -29,43 +26,21 @@
...
-      ],
-      "membership/*": [
-        "projects/membership/*"
-      ],
-      "ophan/embed": [
-        "node_modules/ophan-tracker-js/build/ophan.embed"
-      ],
-      "ophan/ng": [
-        "node_modules/ophan-tracker-js"
-      ],
```
chrislomaxjones, line 52: Thanks for removing these paths!

I was just curious about this `ophan/ng` one - it looks like it's used in `ab-ophan.ts` which we'll copy over from Frontend, do we no longer need this?

chrislomaxjones (commented):
This is great, getting really close to fully migrating now!

standalone/package.json
Jakeii: I'd say it's a dev dependancy? As it's only needed for the bundle build not at runtime. That's how i think about it anyway, happy to change though

standalone/tsconfig.json
Jakeii: I think `ab-ophan.ts` is in because `ab.ts` which imports all the tests is needed for the ab testing to work.
We haven't considered how ab-testing will work have we 🤔

chrislomaxjones: That makes sense, I was just wondering if we needed to retain this `ophan/ng` path for that to work?

> We haven't considered how ab-testing will work have we 🤔

This lib will become another place we'll have to duplicate AB tests maybe?

standalone/package.json
chrislomaxjones: Maybe I'm thinking about it wrong, but I considered any code that gets pulled into the standalone bundle a dependency (so 3rd party stuff like fastdom and our stuff like CMP), since it'll end up being used at runtime? I'm not too sure now, so I'll leave it up to you.

chrislomaxjones (approved):

PR #765, opened 2023-01-23
feat: Add comments-expanded slot type to slotSizeMappings

domlander:
## What does this change?

- Adds `comments-expanded` to `SlotName` type
- Adds `comments-expanded` slot type to `slotSizeMappings`
- Removed duplicates from SlotName type: `top-above-nav` and `comments` and sort alphabetically.

## Why?

- We will be introducing a new ad slot in the comments slot in article pages when the comments are expanded, so we need a name for this new ad slot to differentiate it from other ad slots.


core/src/ad-sizes.ts
```diff
@@ -241,6 +239,17 @@ const slotSizeMappings: SlotSizeMappings = {
 			adSizes.fluid,
 		],
 	},
+	'comments-expanded': {
+		mobile: [
```
emma-imber, line 243: Do we need to define any tablet or desktop sizes for this ad slot?

domlander: I think that the mobile definition of slots encompasses breakpoints for mobile devices and above. The same pattern is used for carrot, right and the merchandising slots

chrislomaxjones (approved):
Makes sense!

PR #769, opened 2023-01-25
chore: Cypress tests for commercial bundle code

Jakeii:
## What does this change?

- Copy cypress testing boilerplate from frontend into the `e2e` directory
- Modify the migration script to include cypress tests
- Add github action for cypress tests and unit testing
- Due to a difference is versions of some node_modules (and the way yarn hoists dependencies) there are some tweaks that need to be made to configs and types of the frontend code, as these are in flux at the moment I've added the tweaks to the branch on frontend (https://github.com/guardian/frontend/compare/bundle-in-core)

I've checked in the frontend bundle code as I've added CI, and I'm pretty sure it won't pass if there's no code


arelra (approved):
Great work @Jakeii 👏

PR #770, opened 2023-01-27
fix: Remove tall ads from comments section

domlander:
## What does this change?

- Removes tall ads (skyscraper, half-page) from comments section
- Reordered breakpoints so that all are in ascending order

## Why?

There is not enough room to display a tall ad without create a large amount of whitespace, pushing the content below down. Therefore we should limit this ad slot to ad sizes that can be placed without moving the content.


SiAdcock (approved):
👌

PR #773, opened 2023-02-02
feat: record `offlineCount` in commercial metrics

sndrs:
## What does this change?

adds `offlineCount` to the network measurements in commercial metrics.

it records the offline count that will be added in https://github.com/guardian/dotcom-rendering/pull/7118 (pending the version bump to `commerical-core` from this change

## Why?

to help the @guardian/open-journalism team understand the extent of intermittent connectivity


core/src/event-timer.ts
```diff
@@ -157,23 +158,16 @@ class EventTimer {
...
+		this.properties = {
+			offlineCount: window.guardian.offlineCount,
+		};
+
+		if (window.navigator.connection) {
+			this.properties.type = window.navigator.connection.type;
+			this.properties.downlink = window.navigator.connection.downlink;
+			this.properties.effectiveType =
+				window.navigator.connection.effectiveType;
+		}
```
mxdvl, line 170: Is it true that when `navigator.connection` is available, all of these fields are, too?

mxdvl (approved):
📶

SiAdcock (approved):

PR #779, opened 2023-02-06
chore: Update lockfile

SiAdcock:
## What does this change?

Chop old typescript versions out of the yarn.lock

## Why?

We [currently resolve TypeScript to `~4.9.0`](https://github.com/guardian/commercial-core/blob/main/package.json#L40)

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -11127,20 +11127,10 @@ typescript-tuple@^2.2.1:
   dependencies:
     typescript-compare "^0.0.2"
 
-typescript@^3.9.10, typescript@^3.9.5, typescript@^3.9.7:
-  version "3.9.10"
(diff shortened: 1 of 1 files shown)


Jakeii:
did you do that manually?

SiAdcock:
@Jakeii No, I just ran `yarn` at the root. The `yarn.lock` got updated automatically. Is that happening for you?

Jakeii (approved):
Oh, cool! I find it strange the install command isn't complaining in CI 🤔

SiAdcock:
> I find it strange the install command isn't complaining in CI 🤔

Yeah, it's a good shout. Are we using [`--frozen-lockfile`](https://classic.yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-frozen-lockfile)?

Jakeii:
I seem to remember yarn uses `--frozen-lockfile` if the `CI` env variable is true, I could be wrong though

PR #783, opened 2023-02-09
fix: tidy up dependencies

Jakeii:
## What does this change?
Updated several dependencies

Updated jest in commercial-core which required updating some of the tests

Replaced uglifyjs in the standalone workspace with terser, as the uglifyjs webpack plugin doesn't officially support webpack 5 

commercial-core also had some peerDeps that were also devDeps so got rid of them.

## Why

An attempt to resolve the dependency issues preventing publishing of packages


arelra (approved):
Nice 🧹

PR #792, opened 2023-02-15
ci: Enable publishing of @guardian/commercial-bundle

Jakeii:
## What does this change?
* Update commercial files from frontend
* Uncomment commercial-bundle jobs in workflows
* added an `.npmrc` with an option that will prevent npm from running it's install command before versioning (https://docs.npmjs.com/cli/v8/commands/npm-version#workspaces-update), this is run via semantic-release

## Why?
Getting ready to use the bundle in frontend🎉


chrislomaxjones (approved):
Nice one!

PR #794, opened 2023-02-20
fix(bundle): include used process.env.* in prod build

Jakeii:
## What does this change?
webpack was only aliasing the `process` module in the dev build, so the prod build didn't actually work in the browser.

Also re-enable the E2E tests that would have caught this...

### DIFF
--- a/bundle/webpack.config.js
+++ b/bundle/webpack.config.js
@@ -13,6 +13,7 @@ module.exports = {
 	},
 	output: {
 		path: path.join(__dirname, 'dist'),
+		clean: true,
 	},
 	resolve: {
 		modules: [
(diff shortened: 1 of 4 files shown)


chrislomaxjones (approved):
👍

bundle/webpack.config.js
```diff
@@ -25,6 +25,7 @@ module.exports = {
 			svgs: path.join(__dirname, 'static', 'svg'),
 			'ophan/ng': 'ophan-tracker-js',
 			lodash: 'lodash-es',
+			process: 'process/browser',
```
arelra, line 28: Just for my understanding why do we need this in the prod build if it wasn't used in Frontend?

Jakeii: Great point
Looking at frontend's webpack it's got:
https://github.com/guardian/frontend/blob/main/webpack.config.prod.js#L31-L33

That would make sure process is defined I think
And would be better because the rest of `process.*` won't be in the bundle

arelra (approved):

PR #797, opened 2023-02-22
ci: Add workflow to publish beta releases with labels

Jakeii:
## What does this change?
Adds a workflow that will publish beta releases from a PR

1. It gets the last version from the repositories tags
2. bumps the prerelease version using `yarn version` 
3. push the tag
4. publishes to npm with the next tag

E.g. If the current version is 1.0.0, the next beta version will be 1.0.1-beta.0, then 1.0.1-beta.1, etc.

`semanic-release` is not used here as it's very opinionated and refuses to work with non-semver versions or pull requests

## Why?
Occasionally we need to test changes on CODE, this will allow us to test multiple changes separately.

### DIFF
--- a/.github/workflows/test-release.yml
+++ b/.github/workflows/test-release.yml
@@ -1,0 +1,94 @@
+name: Publish Beta Package
+
+# This workflow is triggered on a label being added to a PR, and will publish a beta version of the bundle package e.g. @guardian/commercial-bundle-v1.0.1-beta.0
+# It gets the last version from the tags, bumps the prerelease version and publishes to npm with the next tag
+# if the current version is 1.0.0, the next beta version will be 1.0.1-beta.0, then 1.0.1-beta.1, etc.
+# semanic-release is not used here as it's very opinionated and refuses to work with non-semver versions or pull requests
+on:
+  pull_request:
+    types: [labeled]
+
+jobs:
+  bundle-release:
+    name: Bundle Beta Release
+    runs-on: ubuntu-latest
+    if: github.event.label.name == 'Bundle Beta Release'
+    defaults:
+      run:
+        working-directory: ./bundle
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Setup node
+        uses: guardian/actions-setup-node@main
+
+      - name: Install dependencies
+        uses: bahmutov/npm-install@v1
+
+      - name: Build package
+        run: yarn build
+
+      - name: Grab last version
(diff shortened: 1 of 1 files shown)


.github/workflows/beta-release-on-label.yml
```diff
@@ -0,0 +1,136 @@
...
+        uses: guardian/actions-setup-node@main
+
+      - name: Install dependencies
+        uses: bahmutov/npm-install@v1
+
+      - name: Build package
+        run: yarn build
+
+      - name: Grab last version
+        run: yarn version --new-version $(git tag -l | grep @guardian/commercial-bundle-\* |  sed -E 's|.*([0-9]+\.[0-9]+\.[0-9]+(-beta.[0-9]+)?)|\1|g' | tail -1) --no-git-tag-version
```
arelra, line 46: 😎 I'm just wondering if `tail -1` will get the last (aka _numeric_) tag from 'git tag -l' or just the last version that was tagged if that makes sense. If this happens on PRs could there be collisions?

arelra: Also if this script is getting a bit gnarly then it might be worth moving to a fully fledged node script?

arelra (commented):
Nice!

.github/workflows/beta-release-on-label.yml
```diff
@@ -0,0 +1,136 @@
...
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            github.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: '🚀 ${{ steps.version.outputs.BETA_VERSION }} published to npm as a beta release'
+            })
+
+  core-release:
```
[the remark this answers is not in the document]
Jakeii: there's 2 labels one for each package, can release 1 or both!

.github/workflows/beta-release-on-label.yml
Jakeii: Not sure I follow the question 🙈?

arelra: Oh I am wondering if there is a race condition where we do tag version + 1 so that versions can get bumped out of numerical order. But I'm fine with this as is.

arelra (approved):

.github/workflows/beta-release-on-label.yml
Jakeii: hmm, seems quite unlikely, only if 2 PRs were tagged within a couple minutes of each other, in that case can just rerun I think

PR #798, opened 2023-02-23
feat: Refine data naming convention

mxdvl:
## What does this change?

Use a consistent naming convention for the type of data that we send: **Qualitative** & **Quantitative**

Automatically sort EventTimer properties in qualitative or quantitative:
- properties: qualitative, string or categorical
- metrics: quantitative, number or numerical

Keep legacy data around to prevent a breaking change

### Quantitative

For numerical data of type `number`.
Anything that can be measured discretely or continuously:
- size of element
- number of elements
- duration of event
- occurrences of event

See [Quantitative Data](https://en.wikibooks.org/wiki/Statistics/Different_Types_of_Data/Quantitative_and_Qualitative_Data#Quantitative_data) on wiki.

```ts
const data: Quantitative[] = [
	{ name: "time on page", value: 10_320 },
	{ name: "height of banner", value: 360 },
	{ name: "cumulative layout shift", value: 0.0625 },
];
```

### Qualitative
For categorical data of type `string`.

Anything that can be identified with distinct labels:
- edition
- host name
- name of element
- type of connection

See [Qualitative Data](https://en.wikibooks.org/wiki/Statistics/Different_Types_of_Data/Quantitative_and_Qualitative_Data#Qualitative_data) on wiki.

```ts
const data: Qualitative[] = [
	{ name: "edition", value: "UK" },
	{ name: "content ID", value: "/2023/feb/23/…" },
	{ name: "connection type", value: "4g" },
]
```

## Why?

The distinction between “properties” and “metrics” is confusing:
- the file is called “metrics”, but deals with both types
- event “properties” can be `number` (numerical) or `string` (categorical) but currently are only recorded as strings
- data transformation is required in our warehouse to convert some types because of the current confusion
  - Following, #773 we struggled to find the data because it was not where we expected it
- When musing about a unified approach to data collection, [defining these two distinct types of data was key](https://github.com/guardian/libs/pull/323/files#diff-20c0b52052623c9d08e4046e8930c31563414fb8ed4b2fb6596071d78cd0b02aR85-R87)

### DIFF
--- a/.changeset/late-ants-relate.md
+++ b/.changeset/late-ants-relate.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': patch
+---
+
+Refactor internal naming conventions around quantitative and qualitative data.
--- a/src/core/event-timer.ts
+++ b/src/core/event-timer.ts
@@ -251,3 +251,4 @@ class EventTimer {
 }
 
 export { EventTimer };
+export type { EventTimerProperties };
--- a/src/core/send-commercial-metrics.spec.ts
+++ b/src/core/send-commercial-metrics.spec.ts
@@ -7,13 +7,7 @@ import {
 	initCommercialMetrics,
 } from './send-commercial-metrics';
 
-const {
-	Endpoints,
-	mapEventTimerPropertiesToString,
-	reset,
-	roundTimeStamp,
-	transformToObjectEntries,
-} = _;
+const { Endpoints, reset, roundTimeStamp, transformPropertiesObjectToData } = _;
 
 jest.mock('@guardian/consent-management-platform');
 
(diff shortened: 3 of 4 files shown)


ashishpuliyel:
My thoughts:
I was initially concerned this was change in the names, but which leaves the underlying problem (which is people not understanding the difference between the two, and putting numbers in the properties/qualitative field) unsolved.

> the file is called “metrics”, but deals with both types
This is fair.

> event “properties” can be number (numerical) or string (categorical) but currently are only recorded as strings.
It _should_ be only strings, or things you're willing to treat as strings (chrome version number?), as you can see from the current type definition, which remains unchanged.

```
	name: string;
	value: string;
```
But, yeah, people were sometimes using it wrong. You've added lots of documenting comments which is great. Perhaps these names are clearer. They aren't to me, but who's to say my brain is typical.

My primarily objection is that the word 'qualitative' is not usually 'string properties of a thing'. It's 'use words to describe' and to understand the feelings people have in relation to an experience or understand their behaviour.

What if we rename them so metrics contain:
 - string_properties
   - 	name: string;
	value: string;
 - numeric_metrics
   - 	name: string;
	value: number;
	
Very very explict?

mxdvl:
> You've added lots of documenting comments which is great.

After a wider discussion and feedback from you and @emma-imber I’ve opted for focussing on the value brought by comments in #917

PR #800, opened 2023-02-24
fix(offlineCount): measure at page unload

mxdvl:
## What does this change?

Record the `offlineCount` at the point the user navigates away, part of `gatherMetricsOnPageUnload`, rather than at the time of initialisation of `EventTimer`.

## Why?

There’s a possibility that `EventTimer` is initialised before the user has experienced all their offline events.

Follow-up on #773

### DIFF
--- a/core/src/send-commercial-metrics.ts
+++ b/core/src/send-commercial-metrics.ts
@@ -23,7 +23,6 @@ type EventProperties = {
 	type?: ConnectionType;
 	downlink?: number;
 	effectiveType?: string;
-	offlineCount?: number;
 };
 
 type CommercialMetricsPayload = {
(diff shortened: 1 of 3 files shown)


core/src/send-commercial-metrics.ts
```diff
@@ -110,6 +109,25 @@ function sendMetrics() {
 
 type ArrayMetric = [key: string, value: string | number];
 
+/**
+ * Gather how many times the user has experienced the “offline” event
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/Window/offline_event
+ *
+ * This value should accessed as late as through the page lifecycle
+ * to get an accurate value.
```
sndrs, line 117:
```suggestion
 * This value should be fetched as late as possible in the page lifecycle,
 * to get an accurate value.
```

sndrs (approved):
nice one

mxdvl:
This won’t be live before #799 is merged.

mxdvl:
This is done again in #802

PR #801, opened 2023-02-24
fix(e2e): correct DCR dev server path

mxdvl:
## What does this change?

Redo of #799

## Why?

I scheduled it wrong.


arelra (approved):

ashishpuliyel (approved):
It's atomic all right.

mxdvl:
What’s up with coveralls?

PR #805, opened 2023-02-28
fix: update file location for cross icon

emma-imber:
## What does this change?
Updates the file location for the cross icon used in the mobile sticky close button

## Why?
Currently the import does not point to the correct file location, leading to 'undefined' showing where the cross icon should appear

### DIFF
--- a/bundle/src/projects/commercial/modules/dfp/render-advert-label.ts
+++ b/bundle/src/projects/commercial/modules/dfp/render-advert-label.ts
@@ -4,7 +4,7 @@
 -- "Promise returned in function argument where a void return was expected"
 */
 import { getCookie } from '@guardian/libs';
-import crossIcon from 'svgs/icon/cross.svg';
+import crossIcon from '../../../../../static/svg/icon/cross.svg';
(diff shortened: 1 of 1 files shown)


Jakeii (approved):

arelra (approved):

SiAdcock (approved):
✖️

SiAdcock:
Is [the consentless version](https://github.com/guardian/commercial/blob/d01eebafdad51bc3d9d133d0cbb85fbc2c4491c8/bundle/src/projects/commercial/modules/consentless/render-advert-label.ts#L6) suffering from the same fate?

emma-imber:
> Is [the consentless version](https://github.com/guardian/commercial/blob/d01eebafdad51bc3d9d133d0cbb85fbc2c4491c8/bundle/src/projects/commercial/modules/consentless/render-advert-label.ts#L6) suffering from the same fate?

Good point - I'm not sure if we have the format in consentless yet? If we don't, we can tidy up the render label file for consentless. Happy to investigate this tomorrow!

PR #806, opened 2023-03-01
fix(bundle): mobile sticky cross icon

Jakeii:
## What does this change?
We discovered it was due to a webpack 5 issue, replaced `svg-loader` with `raw-loader`


emma-imber (approved):
🎉

PR #809, opened 2023-03-01
feat: Create elements manager ab test. Place test ad in right slot.

domlander:
## What does this change?

- Create an Elements Manager 0% AB test. This is [mirrored in frontend](https://github.com/guardian/frontend/pull/25940).
- When in the test:
    - Don't load opt-out or googletag
    - Load an example ad in an iframe in the right slot.

To test, note that you will need to run elements manager on port 3002 to see the ad.

## Why?

This is a step towards using elements manager to place elements on the page.

### DIFF
--- a/bundle/src/bootstraps/standalone.commercial.ts
+++ b/bundle/src/bootstraps/standalone.commercial.ts
@@ -21,6 +21,7 @@ import { log } from '@guardian/libs';
 import { initTeadsCookieless } from 'commercial/modules/teads-cookieless';
 import { isInVariantSynchronous } from 'common/modules/experiments/ab';
 import { consentlessAds } from 'common/modules/experiments/tests/consentlessAds';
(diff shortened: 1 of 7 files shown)


chrislomaxjones (approved):
Very exciting!

bundle/src/bootstraps/standalone.commercial.ts
```diff
@@ -231,6 +237,12 @@ const chooseAdvertisingTag = async () => {
 	}
 };
 
+if (isInVariantSynchronous(elementsManager, 'variant')) {
+	void import('../projects/commercial/modules/elementsManager').then(
```
chrislomaxjones, line 241: Up to you but you could add a comment to instruct webpack to give this chunk a name e.g.

```
void import(/* webpackChunkName: "elements-manager" */ ...)
```

PR #815, opened 2023-03-06
fix: `dcre` should use `dcrCouldRender`

arelra:
## What does this change?

The targeting property `dcre` (dcr-eligible) should derive from `config.page.dcrCouldRender` and not `config.isDotcomRendering`


chrislomaxjones (approved):
Good spot!

Jakeii (approved):

PR #816, opened 2023-03-06
chore: Parallel e2e tests

arelra:
## What does this change?

- parallel e2e tests
- only install production deps


emma-imber (approved):
Nice!

PR #819, opened 2023-03-08
test: Visual e2e workflow & GHA refactors

arelra:
## What does this change?

Mostly:

- Get the workflow `e2e-visual` i.e. Percy visual tests working in this repo (currently only as workflow_dispatch)

But also:

- Replace `guardian/actions-setup-node` and `bahmutov/npm-install` as per:
  https://github.com/guardian/dotcom-rendering/pull/7117
  https://github.com/guardian/dotcom-rendering/pull/7352
- Use yarn to install - installs are now 30-40% faster
- Renaming to align the tests and drop 'commercial' from the names. Please do shout if these names could be better
- Rename e2e scripts
- Drop cypress-io action in favour of custom server start jobs
- Drop the 'docs' job which was only generating a TOC for /core/README.md
- Use Sourcepoint classes instead of button text


emma-imber (approved):
🎉

e2e/cypress/support/commands.ts
```diff
@@ -30,26 +30,20 @@ Cypress.Commands.add('findAdSlotIframeBySlotId', (adSlotId: string) => {
...
-const allowAllButtons = ['Yes, I’m happy', 'Accept all', 'Yes, ok']
-	.map((title) => `button[title="${title}"]`)
-	.join(',');
-const manageConsent = 'Manage my cookies';
-const rejectAll = 'Reject all';
-
 Cypress.Commands.add('rejectAllConsent', () => {
 	cy.getIframeBody('sp_message_iframe_')
-		.find(`button[title="${manageConsent}"]`)
+		.find('button.sp_choice_type_12')
```
domlander, line 35: Could we add a comment here to explain that this is Sourcepoint's special class name for the "manage cookies" button? And also maybe that this is the currently the best way available to us to identify this button.

e2e/cypress/support/commands.ts
```diff
@@ -30,26 +30,20 @@ Cypress.Commands.add('findAdSlotIframeBySlotId', (adSlotId: string) => {
...
-		.find(`button[title="${rejectAll}"]`, { timeout: 30000 })
+		.find('button.sp_choice_type_REJECT_ALL', { timeout: 30000 })
 		.click();
 	cy.wait(100);
 });
 
 Cypress.Commands.add('allowAllConsent', () => {
 	cy.getIframeBody('sp_message_iframe_')
-		.find(allowAllButtons, { timeout: 30000 })
+		.find('button.sp_choice_type_11', { timeout: 30000 })
```
domlander, line 46: Also here for "Accept cookies"

.github/workflows/e2e-visual.yml
```diff
@@ -0,0 +1,88 @@
+name: E2E Visual
+on: workflow_dispatch
+  # pull_request:
```
domlander, line 3: Are these lines intentionally left commented out?

domlander (approved):
I like the new alignment on names 👍

.github/workflows/e2e-visual.yml
arelra: Yes intentional as there is an issue with Safari so kept it as a manual trigger for now.

Jakeii (approved):
Great stuff

PR #822, opened 2023-03-13
fix: Update and fix deps

arelra:
## What does this change?

It looks like the migration from Frontend introduced regressions to dependencies by using the versions from Frontend.

- bump all @guardian packages to their latest versions
- bump typescript to align with csnx ~4.9.5
- move deps to `devDependencies` where possible (please verify they are not CI/publish related :)
- bump commercial-core in commercial-bundle


Jakeii:
I think we can get rid of the top level dependencies too https://github.com/guardian/commercial/blob/main/package.json#L35C5-L38
They were only there for the migration script, I think `madge` might be responsible for some weird typescript version resolutions

arelra:
> I think `madge` might be responsible for some weird typescript version resolutions

I noticed using `yarn workspace ...` would cause the typescript versions in yarn.lock to flip flop between old versions so uses `resolutions` everywhere.

Ok cool I'll remove the migration script too

Jakeii (approved):

bundle/package.json
```diff
@@ -88,26 +91,25 @@
...
 		"ophan-tracker-js": "^1.4.0",
 		"prebid.js": "guardian/prebid.js#2e3b96d",
 		"process": "^0.11.10",
 		"raven-js": "^3.27.2",
 		"tslib": "^2.4.1",
 		"web-vitals": "^2.1.2",
 		"wolfy87-eventemitter": "^5.2.9"
+	},
+	"resolutions": {
+		"typescript": "~4.9.5"
```
mxdvl, line 113: Was this intentionally kept in? I think that it has prevented Typescript from actually upgrading in #1222, #899 and #881

PR #823, opened 2023-03-14
feat: Move adtest clear link to the ad container

emma-imber:
## What does this change?
Moves the adtest clear link to the container level, to prevent the iv being expanded by GAM.

## Why?
This should resolve the issue of divs expanding and pushing down ads when the adtestInLabels cookie is true.

Note - draft for now, using the PR to create a release for testing on CODE.

### DIFF
--- a/bundle/src/projects/commercial/modules/dfp/render-advert-label.ts
+++ b/bundle/src/projects/commercial/modules/dfp/render-advert-label.ts
@@ -76,19 +76,17 @@ const createAdTestCookieRemovalLink = (
 ): HTMLElement => {
 	const adTestCookieRemovalLink = document.createElement('div');
 	adTestCookieRemovalLink.style.cssText =
-		'position: relative;padding: 0 0.5rem;text-align: left;box-sizing: border-box;';
+		'position: relative;padding: 0;text-align: left;box-sizing: border-box;display: block:width: 0;height: 0';
 
-	if (adTestName) {
-		const url = new URL(window.location.href);
-		url.searchParams.set('adtest', 'clear');
+	const url = new URL(window.location.href);
+	url.searchParams.set('adtest', 'clear');
 
-		const clearLink = document.createElement('a');
-		clearLink.className = 'ad-slot__adtest-cookie-clear-link';
-		clearLink.href = url.href;
-		clearLink.innerHTML = 'clear';
+	const clearLink = document.createElement('a');
+	clearLink.className = 'ad-slot__adtest-cookie-clear-link';
+	clearLink.href = url.href;
+	clearLink.innerHTML = 'clear';
 
-		adTestCookieRemovalLink.appendChild(clearLink);
-	}
+	adTestCookieRemovalLink.appendChild(clearLink);
 
 	return adTestCookieRemovalLink;
 };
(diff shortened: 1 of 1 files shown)


bundle/src/projects/commercial/modules/dfp/render-advert-label.ts
```diff
@@ -71,32 +71,33 @@ const createAdTestLabel = (
...
 const renderAdvertLabel = (adSlotNode: HTMLElement): Promise<Promise<void>> => {
 	return fastdom.measure(() => {
 		if (shouldRenderLabel(adSlotNode)) {
-			const renderAdTestLabel = shouldRenderAdTestLabel();
+			const renderAdTestLabel = shouldRenderAdTestLabel(adSlotNode);
+			const adTestClearExists =
+				adSlotNode.parentNode?.firstElementChild?.firstElementChild
+					?.nodeName === 'A'
+					? true
+					: false;
```
chrislomaxjones, line 100: Is this checking if the elment is an anchor tag? If so, it might be more idiomatic to use

```typescript
element instanceof HTMLAnchorElement
```

chrislomaxjones (approved):
Nice!

domlander:
I wonder if adding `overflow: hidden; white-space: nowrap;` might be better than hiding the ad label for skyscraper ads? This will stop the text content from overflowing its container.

It doesn't look good, but this way we still keep the message so that the user knows that they are in an ad-label test. It also is a bit more future-proof in case we use other thin ad types.

Will leave it up to you.

image

bundle/src/projects/commercial/modules/dfp/render-advert-label.ts
```diff
@@ -71,32 +71,33 @@ const createAdTestLabel = (
...
 };
 
-const createAdTestCookieRemovalLink = (
-	adTestName: string | null,
-): HTMLElement => {
+const createAdTestCookieRemovalLink = (): HTMLElement => {
 	const adTestCookieRemovalLink = document.createElement('div');
 	adTestCookieRemovalLink.style.cssText =
-		'position: relative;padding: 0 0.5rem;text-align: left;box-sizing: border-box;';
+		'position: relative;padding: 0;text-align: left;box-sizing: border-box;display: block:width: 0;height: 0';
```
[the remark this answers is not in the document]
emma-imber: Good spot!

bundle/src/projects/commercial/modules/dfp/render-advert-label.ts
emma-imber: Oh nice idea! I've made the change

emma-imber:
> I wonder if adding `overflow: hidden; white-space: nowrap;` might be better than hiding the ad label for skyscraper ads? This will stop the text content from overflowing its container.
> 
> It doesn't look good, but this way we still keep the message so that the user knows that they are in an ad-label test. It also is a bit more future-proof in case we use other thin ad types.
> 
> Will leave it up to you.
> 
> image

I'm going to leave this as I had it - where people see a skyscraper ad, they are also probably seeing a top above nav or right ad on the same page which should contain the ad test label in, so they should still be able to see that an ad test is active!

domlander (approved):

PR #826, opened 2023-03-15
feat: Client-side targeting for elements-manager

chrislomaxjones:
## What does this change?

This PR lays out the initial functionality for targeting in the GEM client.

We fetch the full set of elements, with their assets and targeting, from the endpoint introduced in [this elements-manager PR](https://github.com/guardian/elements-manager/pull/21). In production we'll retrieve an object of the same shape from the CDN. It is then the responsibility of the client to determine, for each slot, which element would be eligible to fill it.

### Summary of how this targeting works

Imagine an element has the following targeting fetched from the elements-manager UI:

```json
[
  { "key": "slot", "value": "mostpop,comments" },
  { "key": "k", "value": "commentisfree" }
]
```

For a slot to be filled by this element, the combined page and slot targeting for this slot must contain:
- the key `k` with at least one corresponding value that is `commentisfree`.
- the key `slot` with at least one corresponding value that is `mostpop` or `commentisfree`.

To compute this, we roughly follow:

- Assemble a _page_ targeting object, in the typical way. Here we check consent state and either choose the full page targeting object, or the consent-less targeting object.
- For each slot on the page, combine the previous page targeting with some slot-level targeting (at the moment just e.g. `slot: top-above-nav`.
- Check this full targeting for a slot against the retrieved elements, to determine which are eligible.
- To determine eligibility of an element, we take each of the targeting rules defined for it and check **every** one of them can be satisfied by the targeting. This is in effect _ANDing_ all of the targeting rules together. We also ensure that an element has at least one linked asset, otherwise it's considered ineligible.
- To determine if a single rule is satisfied, we check that the key is defined in the targeting and that at least one of the values  is present corresponding to that key. Checking for at least one allows users of GEM to target multiple things for a single element e.g. setting _slot is either mostpop or comments_.
- From the eligible elements, we currently just select one at random. This is considered the _winning element_.
- Select an asset at random from the winning element. This should always be present, since eligible elements have at least one linked asset.
- Fill the given slot with the asset set as an iframe `src`, as well as rendering the label.

### Caveats

- A single element's asset can appear in more than one slot on a given page. We could change this down the line, to make it more like GAM.
- No notion of _priority_ when targeting elements. At present, there's no way to target an ad test in the way you would with GAM, where it beats out all of the other elements on the page.

## Examples

Apply `k=uk` and `k=climate-crisis` targeting to target two elements to separate section fronts. Note the assets fill each of the slots on the page.

https://user-images.githubusercontent.com/8000415/225327156-3cf763c3-6141-407a-9562-ecaf45064861.mov

<hr>

Apply `slot=mostpop` and `slot=merchandising` targeting to two elements to target separate slots on an article. Note that slots default to MPU size for the moment, hence the scrolling.

https://user-images.githubusercontent.com/8000415/225328978-9dafacf5-cd91-4ef9-8c04-bacdcd769e8a.mov

## Why?

Allows for the targeting applied in the GEM UI be reflected by the placement of elements (and their assets) on the client.

### DIFF
--- a/bundle/src/projects/commercial/modules/elements-manager/targeting.ts
+++ b/bundle/src/projects/commercial/modules/elements-manager/targeting.ts
@@ -12,7 +12,6 @@ import type {
 	GuElement,
 	SerializedPayload,
 	TargetingRule,
-	TargetingRules,
 } from './types';
 
 export const fetchSelectionPayload = async (): Promise<GuElement[]> => {
@@ -105,7 +104,7 @@ const satisfiesRule = (
  */
 const satisfiesTargeting = (
 	pageTargeting: PageTargeting,
-	targetingRules: TargetingRules,
+	targetingRules: TargetingRule[],
 ): boolean =>
 	targetingRules.every((rule) => satisfiesRule(pageTargeting, rule));
 
--- a/bundle/src/projects/commercial/modules/elements-manager/targeting.ts
+++ b/bundle/src/projects/commercial/modules/elements-manager/targeting.ts
@@ -132,15 +132,12 @@ export const initTargeting = (
 	elements: GuElement[],
 	pageTargeting: PageTargeting,
 ) => {
-	const selectAssetForSlot = (slotId: string): Asset | undefined => {
-		const targeting = { ...pageTargeting, slot: slotId };
-
-		// Select all of the eligible candidates from their targeting
+	const selectAssetForSlot = (
+		slotTargeting: PageTargeting,
+	): Asset | undefined => {
+		const targeting = { ...pageTargeting, ...slotTargeting };
 		const candidates = eligibleElements(targeting, elements);
-
 		const winningElement = selectAtRandom(candidates);
-
-		// Choose an asset at random from the winning element
 		return winningElement
 			? selectAtRandom(winningElement.assets)
 			: undefined;
--- a/bundle/src/projects/commercial/modules/elements-manager/targeting.ts
+++ b/bundle/src/projects/commercial/modules/elements-manager/targeting.ts
@@ -7,25 +7,62 @@ import type { ConsentState } from '@guardian/consent-management-platform/dist/ty
 import { isString } from 'lodash-es';
 import { commercialFeatures } from 'common/modules/commercial/commercial-features';
 import { getSynchronousParticipations } from 'common/modules/experiments/ab';
-import type {
-	Asset,
-	GuElement,
-	SelectionPayload,
-	TargetingRule,
-	TargetingRules,
-} from './types';
+import type { Asset, GuElement, TargetingRule, TargetingRules } from './types';
 
-export const fetchSelectionPayload = (): Promise<SelectionPayload> => {
-	// TODO replace this with a fetch
-	return Promise.resolve({
-		elements: [],
-	});
+type SerializedPayload = Array<{
+	id: string;
+	targeting: Array<{
+		key: string;
+		// Here multiple values may occupy the same string, but are comma separated
+		// e.g. `sport,culture`
+		value: string;
+	}>;
+	assets: Asset[];
+}>;
+
+export const fetchSelectionPayload = async (): Promise<GuElement[]> => {
+	// Endpoints are required for development purposes
+	const payloadUrl = `http://localhost:3000/api/elements`;
+	const assetUrl = `http://localhost:3000/api/dev/asset`;
+
+	try {
+		const res = await fetch(payloadUrl);
+
+		if (!res.ok) {
+			console.error(
+				'Network error trying to fetch the selection payload',
+			);
+			return [];
+		}
+
+		const json = (await res.json()) as SerializedPayload;
+
+		// Transform the serialized payload into an array of elements
+		const elements: GuElement[] = json.map((element) => ({
+			...element,
+			assets: element.assets.map((asset) => ({
+				...asset,
+				path: `${assetUrl}/${asset.path}`,
+			})),
+			targeting: element.targeting.map(({ key, value }) => ({
+				key,
+				value: new Set(value.split(',')),
+			})),
+		}));
+
+		return elements;
+	} catch (error) {
+		console.error(error);
+		return [];
+	}
 };
 
 /**
  * Obtain consented or consentless page targeting object, based on consent state.
  */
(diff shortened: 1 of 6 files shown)


bundle/src/projects/commercial/modules/elements-manager/targeting.ts
```diff
@@ -0,0 +1,146 @@
...
+	GuElement,
+	SerializedPayload,
+	TargetingRule,
+	TargetingRules,
+} from './types';
+
+export const fetchSelectionPayload = async (): Promise<GuElement[]> => {
+	// Endpoints are required for development purposes
+	const payloadUrl = `http://localhost:3000/api/elements`;
+	const assetUrl = `http://localhost:3000/api/dev/asset`;
```
chrislomaxjones, line 21: We'll swap these out for Production URLs once this we're ready for that.

bundle/src/projects/commercial/modules/elements-manager/targeting.ts
```diff
@@ -0,0 +1,146 @@
...
+		return false;
+	} else if (isString(targetingValue)) {
+		// If there is a single value in the targeting for this key, check that
+		// string is present in the condition's values
+		return condition.value.has(targetingValue);
+	} else {
+		// If there's multiple values in the targeting for this key, check that
+		// at least one is present in the condition's values
+		return targetingValue.some((v) => condition.value.has(v));
+	}
```
domlander, line 98: We don't need the `else if`'s here
```
	if (!targetingValue) {
		// If there is no value in the targeting for the corresponding key then
		// this rule can't be satisfied
		return false;
	}

	if (isString(targetingValue)) {
		// If there is a single value in the targeting for this key, check that
		// string is present in the condition's values
		return condition.value.has(targetingValue);
	}

	// If there's multiple values in the targeting for this key, check that
	// at least one is present in the condition's values
	return targetingValue.some((v) => condition.value.has(v));
};
```

domlander:
Can there be some unit tests added around targeting? Specifically, ensure that a certain element is selected or not based on its targeting.

Maybe including these conditions:

Page targeting | Element targeting | Can serve?
-- | -- | --
section not included | section=culture |  
section=culture | section=culture |  
section=culture | section not included |  
section=culture | section=culture,sport |  
section=culture,sport | section=culture |  
section not included | section not included |

chrislomaxjones:
> Can there be some unit tests added around targeting? 

Great idea, on it.

bundle/src/projects/commercial/modules/elements-manager/targeting.ts
```diff
@@ -0,0 +1,146 @@
...
+			element.assets.length > 0 &&
+			// The page targeting object supplied satisfies the element's targeting rules
+			satisfiesTargeting(pageTargeting, element.targeting),
+	);
+
+const selectAtRandom = <T>(candidates: T[]) =>
+	// TODO remove this type assertion when we can enable --noUncheckedIndexedAccess compiler option
+	candidates[Math.floor(Math.random() * candidates.length)] as T | undefined;
+
+export const initTargeting = (
```
domlander, line 131: Is there perhaps is a more descriptive name for what this function is doing? It looks like it's finding an asset to serve given a list of elements and based on targeting.

bundle/src/projects/commercial/modules/elements-manager/targeting.ts
```diff
@@ -0,0 +1,146 @@
...
+		slotTargeting: PageTargeting,
+	): Asset | undefined => {
+		const targeting = { ...pageTargeting, ...slotTargeting };
+		const candidates = eligibleElements(targeting, elements);
+		const winningElement = selectAtRandom(candidates);
+		return winningElement
+			? selectAtRandom(winningElement.assets)
+			: undefined;
+	};
+	return { selectAssetForSlot };
```
domlander, line 145: Why do we return wrap the winning asset in an object when returning it, rather than return the object itself?

bundle/src/projects/commercial/modules/elements-manager/types.ts
```diff
@@ -0,0 +1,31 @@
...
+	id: string;
+	path: string;
+	clickThroughUrl: string;
+	width?: number;
+	height?: number;
+};
+
+export type GuElement = {
+	id: string;
+	targeting: TargetingRules;
```
domlander, line 15: Do we need the `TargetingRules` type? This might be convention I'm unaware of

bundle/src/projects/commercial/modules/elements-manager/targeting.ts
chrislomaxjones: Makes sense, have made the change.

bundle/src/projects/commercial/modules/elements-manager/targeting.ts
```diff
@@ -0,0 +1,146 @@
...
+import { getSynchronousParticipations } from 'common/modules/experiments/ab';
+import type {
+	Asset,
+	GuElement,
+	SerializedPayload,
+	TargetingRule,
+	TargetingRules,
+} from './types';
+
+export const fetchSelectionPayload = async (): Promise<GuElement[]> => {
```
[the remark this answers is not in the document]
chrislomaxjones: 👍

bundle/src/projects/commercial/modules/elements-manager/targeting.ts
chrislomaxjones: Have removed this function now, so naming doesn't apply anymore.

domlander (approved):

bundle/src/projects/commercial/modules/elements-manager/targeting.ts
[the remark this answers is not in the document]
Jakeii: I think there's an eslint rule for this

Jakeii (approved):

bundle/src/projects/commercial/modules/elements-manager/targeting.ts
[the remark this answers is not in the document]
chrislomaxjones: Ah yes [this one](https://github.com/guardian/commercial-core/blob/802090a9fe3b4acc44e7bf81970ea6664b1ced81/core/.eslintrc.js#L20-L21) - looks like it only applies to the core lib at the minute, we should apply it to both!

PR #828, opened 2023-03-15
ci: Tweak actions to build before parallel cypress runs

Jakeii:
## What does this change?
Thanks to @arelra we have parallel cypress runs, this adds a small optimisation to build once for all the cypress runners


arelra (approved):
This is a nice improvement thanks! The checks will probably take roughly the same time but it should shave 1/2 mins off a retry.

It's a shame we have to install dependencies twice to get the node_modules back. Adding node_modules to the artefact I assume is not feasible due to its GB+ size and number of files.

Jakeii:
Not sure why this introduced flakiness🤔

Also something we don't think about too often is that Github action minutes = $$$!

PR #829, opened 2023-03-15
fix: Remove `NoCarrotAdsNearNewsletterSignupBlocks`

simonbyford:
## What does this change?

Removes the `NoCarrotAdsNearNewsletterSignupBlocks` experiment. The change it was built to test has been live for 100% of users [since December](https://github.com/guardian/frontend/pull/25756).

## Why?

This was never run and would provide limited insight if run today. I'm proposing we ditch it.


chrislomaxjones (approved):
Makes sense!

PR #830, opened 2023-03-16
feat: Make spacefinder ignore thumbnails

simonbyford:
## What does this change?

Ignore thumbnails when placing `inline1` ads on desktop

## Why?

construction


domlander:
I think the screenshots might be the wrong way around

domlander (approved):
Nice addition! Do we want to measure the impact of this change?

domlander:
> Richlinks are a lot more common than thumbnails and fixing https://github.com/guardian/frontend/pull/24690 led to a revenue uplift in the order of a few £1000 a month.

Gotcha. That makes sense. An AB test would likely not be worth it then.

PR #835, opened 2023-03-21
refactor: Improve spacefinder debugger

simonbyford:
## What does this change?

Makes a number of improvements and simplifications to the spacefinder debugger.

The main change is that instead of displaying information about one particular pass of spacefinder (specified by query parameter, for example `?sfdebug=1` or `?sfdebug=carrot`), it now shows information about all passes. The different passes can be toggled via a UI.

https://user-images.githubusercontent.com/7423751/226976136-2cfcdbc3-16f7-4249-82fa-3ae248621f27.mov

---

This required some changes to the architecture. Instead of highlighting paragraphs on the fly as spacefinder runs, the debugger first "annotates" the DOM with metadata, with the actual highlighting (or "rendering") of paragraphs deferred until later when all passes have completed. 

More details about the two phases:

### 1) Annotation

The debugger is called with a reference to the current `pass` (which is just an arbitrary string like `inline1` or `carrot`). The debugger then...

1) exits if the user is not in debug mode
2) adds a UI panel to the page (unless it has already been added on a previous pass)
3) adds classes and data attributes to each of the candidates with metadata about why they were excluded/chosen. For example, this paragraph:

```
<p>
   Text
</p>
```

might become:

```
<p class="candidate" data-sfdebug-inline1="tooClose">
   Text
</p>
```

4) adds hover listeners to candidates which were excluded for the `tooClose` reason
5) adds classes and data attributes to the "body" element with metadata required to render the "min above indicator" line. The body is the element specified in the [`bodySelector`](https://github.com/guardian/commercial/blob/b5df13366e0f0efd9c742aa8804a1ff4ac311528/bundle/src/projects/common/modules/spacefinder.ts#L28) in the `SpacefinderRules` object.

For example, this:

```
<div class="article-body-commercial-selector>
   <p>Text</p>
   <p>Text</p>
   <p>Text</p>
</div>
```

might become:

```
<div class="article-body-commercial-selector" id="sfdebug-body" data-sfdebug-min-above-inline1="200px">
   <p>Text</p>
   <p>Text</p>
   <p>Text</p>
</div>
```

6) adds a button to the UI to trigger a rendering for that pass

### 2) Rendering

Rendering only happens when the user toggles the appropriate pass via the UI. In practice this is always after spacefinder has completed all passes, given spacefinder runs almost instantaneously on page load.

The rendering phase simply highlights the candidates and moves the "min above indicator" line based on the annotations from the first phase.

---

This new approach has several advantages:

- removes the need for multiple different query parameters - you can just use `?sfdebug`. Furthermore it's easy to see which passes are valid for a given page. For example, articles in desktop will typically show 3 buttons (4 if `im` is enabled) while liveblogs will only show one.
- all debug logic has been removed from spacefinder. Instead, the debugger decides whether to run. Furthermore the [`'spacefiller-complete'`](https://github.com/guardian/commercial/pull/835/files#diff-248119827b80a063e163cca0514082fefa659cf8cfa2baf173a10c7fd1883ac0L9-L17) event is no longer required, which further simplifies the interface between spacefinder and the debugger
- because rendering now happens after spacefinder has finished running, there is no longer a risk of spacefinder making decisions based on elements added by the debugger in previous passes. This means we can safely remove all [`.sfdebug` selectors](https://github.com/guardian/commercial/blob/main/bundle/src/projects/commercial/modules/article-body-adverts.ts#L391)

Other changes in this PR:

- improves the appearance of the "min above indicator" line
- reimplements "explainer" elements as `::before` pseudo-elements
- centralises most styling in a `<style>` element instead of having everything inline
- fixes a bug in liveblogs where the "explainer" elements were being shown in the wrong place
- adds the [official spacefinder logo](https://github.com/guardian/commercial/pull/835/files#diff-0ef0470641c2c8b93806ab15a055deed0967776f4ccd09576bec7f5a1a96a92f)

Known regressions:

- all ads now have a red outline, not just ones from the given pass. This would be straightforward enough to reintroduce if we decide this is a useful feature to keep.
- you have to click something to see the results. This is non-trivial to fix because - without introducing a new event - there is no way to tell when all passes of spacefinder have completed, and therefore when it is safe to render. Adding a new event might be the way to go but I thought I'd keep the interface as clean as possible to begin with. **edit: could probably also just wait 500ms and it would work in 99.99% of cases**


bundle/src/projects/common/modules/spacefinder.ts
```diff
@@ -81,7 +83,7 @@ const LOADING_TIMEOUT = 5_000;
 const defaultOptions: SpacefinderOptions = {
 	waitForImages: true,
 	waitForInteractives: false,
-	debug: false,
+	pass: 'inline1',
```
simonbyford, line 86: This is questionable, `tsc` made me do it. There is no appropriate default value. But luckily this property is non-optional so the default will never actually be used.

ashishpuliyel (commented):
This is outrageously good and I'm very excited about the UI.

arelra (approved):
This is fantastic work! Spacefinder debug 2.0 ! 👏👏👏

arelra:
This PR video deserves an EPIC version 🚀 

https://user-images.githubusercontent.com/7014230/227196084-3ffd1306-993f-4750-b7b6-821cda990165.mp4

PR #851, opened 2023-04-05
fix: Skip loading Opt Out if in the elements-manager test

chrislomaxjones:
## What does this change?

When in the elements-manager variant, don't load the consentless bundle.

## Why

Fix so we can properly enter the consentless AB test.

### DIFF
--- a/bundle/src/bootstraps/standalone.commercial.ts
+++ b/bundle/src/bootstraps/standalone.commercial.ts
@@ -221,7 +221,7 @@ const chooseAdvertisingTag = async () => {
 	// Only load the Opt Out tag in TCF regions when there is no consent for Googletag
 	if (consentState.tcfv2 && !getConsentFor('googletag', consentState)) {
 		// Don't load OptOut (for now) if loading Elements Manager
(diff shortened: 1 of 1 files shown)


bundle/src/bootstraps/standalone.commercial.ts
```diff
@@ -221,7 +221,7 @@ const chooseAdvertisingTag = async () => {
 	// Only load the Opt Out tag in TCF regions when there is no consent for Googletag
 	if (consentState.tcfv2 && !getConsentFor('googletag', consentState)) {
 		// Don't load OptOut (for now) if loading Elements Manager
-		if (!isInVariantSynchronous(elementsManager, 'variant')) {
+		if (isInVariantSynchronous(elementsManager, 'variant')) {
```
arelra, line 224: Will this also skip loading the regular commercial bundle if the user is in the elements manager test?

arelra: Ah yes :)

arelra (approved):

PR #852, opened 2023-04-06
refactor: handle DuckDuckGo googletag Sentry error

emma-imber:
## What does this change?
If we detect that a user is on the DuckDuckGo browser, handle loadScript errors when trying to load googletag.

## Why?
The default behaviour of DuckDuckGo is to block requests to googletag, meaning we see a lot of errors in Sentry as a result. By handling the promise rejection, we prevent the error from being logged by Sentry, reducing noise.

We replace
Screenshot 2023-04-06 at 15 54 18

with
Screenshot 2023-04-06 at 15 45 19

### DIFF
--- a/bundle/src/projects/commercial/modules/dfp/prepare-googletag.ts
+++ b/bundle/src/projects/commercial/modules/dfp/prepare-googletag.ts
@@ -4,7 +4,7 @@ import {
 	onConsent,
 } from '@guardian/consent-management-platform';
 import type { ConsentState } from '@guardian/consent-management-platform/dist/types';
-import { loadScript } from '@guardian/libs';
+import { loadScript, log } from '@guardian/libs';
 import { init as initMeasureAdLoad } from 'commercial/modules/messenger/measure-ad-load';
 import { isInVariantSynchronous } from 'common/modules/experiments/ab';
 import { elementsManager } from 'common/modules/experiments/tests/elements-manager';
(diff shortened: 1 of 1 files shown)


bundle/src/projects/commercial/modules/dfp/prepare-googletag.ts
```diff
@@ -136,11 +136,29 @@ export const init = (): Promise<void> => {
...
+							'//www.googletagservices.com/tag/js/gpt.js',
+						{ async: false },
+					).catch(duckDuckErrorHandler);
+				} else {
+					void loadScript(
+						window.guardian.config.page.libs?.googletag ??
+							'//www.googletagservices.com/tag/js/gpt.js',
+						{ async: false },
+					);
+				}
```
Jakeii, line 161:
```suggestion
				//DuckDuckGo blocks googletag request by default, creating a lot of noise in Sentry
				//This flow allows us to handle errors originating from DuckDuckGo without spamming Sentry
				loadScript(
						window.guardian.config.page.libs?.googletag ??
							'//www.googletagservices.com/tag/js/gpt.js',
						{ async: false },
					).catch((error: Error) => {
					if (navigator.userAgent.includes('DuckDuckGo')) {
						log(
							'commercial',
							'🦆 Caught loadScript error on DuckDuckGo',
							error,
						);
					} else {
						throw error;
					}
				});

Jakeii (approved):

bundle/src/projects/commercial/modules/dfp/prepare-googletag.ts
emma-imber: Great shout! Will make the change

PR #858, opened 2023-04-19
feat: Add new `allkw` targeting key

Jakeii:
## What does this change?
Add a targeting key `allkw` that is the unique values from both `urlkw` and `k`

## Why?
Excerpt from the ticket:

> The reason for a new key-value is that there still remains a number of use cases for just using urlkw and just k. However, the new key-value will populate when either are present - which helps resolve issues (e.g. from Google, where they want to block 100s of distinct keywords regardless of it it's in the URL or editorial keywords).

### DIFF
--- a/core/src/targeting/build-page-targeting.ts
+++ b/core/src/targeting/build-page-targeting.ts
@@ -70,10 +70,6 @@ const filterValues = (pageTargets: Record<string, unknown>) => {
 	return filtered;
 };
 
-const concatUnique = (a: string[], b: string[]): string[] => [
-	...new Set([...a, ...b]),
-];
-
 type BuildPageTargetingParams = {
 	adFree: boolean;
 	clientSideParticipations: Participations;
@@ -91,6 +87,10 @@ const buildPageTargeting = ({
 
 	const adFreeTargeting: { af?: True } = adFree ? { af: 't' } : {};
 
+	const sharedAdTargeting = page.sharedAdTargeting
+		? getSharedTargeting(page.sharedAdTargeting)
+		: {};
+
 	const contentTargeting: ContentTargeting = getContentTargeting({
 		webPublicationDate: page.webPublicationDate,
 		eligibleForDCR: page.dcrCouldRender,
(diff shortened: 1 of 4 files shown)


core/src/targeting/build-page-targeting.ts
```diff
@@ -136,13 +141,18 @@ const buildPageTargeting = ({
 		youtube,
 	});
 
+	const otherTargeting = {
```
arelra, line 144: Was wondering if rather than introducing `otherTargeting` we could derive `allkw` in `contentTargeting` by shifting the order and passing it `k`?

arelra (approved):

core/src/targeting/build-page-targeting.ts
Jakeii: yeah, sounds good to me! Wasn't sure where to put it tbh

core/src/targeting/content.ts
```diff
@@ -104,6 +106,10 @@ const getUrlKeywords = (url: SharedTargeting['url']): string[] => {
 	return isString(lastSegment) ? lastSegment.split('-').filter(Boolean) : [];
 };
 
+const concatUnique = (a: string[], b: string[]): string[] => [
+	...new Set([...a, ...b]),
+];
```
chrislomaxjones, line 111: 😍

core/src/targeting/content.ts
```diff
@@ -86,6 +86,8 @@ type ContentTargeting = {
 	 * [gam]: https://admanager.google.com/59666047#inventory/custom_targeting/detail/custom_key_id=195087
 	 */
 	vl: null | (typeof videoLengths)[number];
+
+	allkw: string[];
```
chrislomaxjones, line 90: Could we add a TSDoc that links to the key-value in Ad Manager like the others above?

chrislomaxjones (approved):

core/src/targeting/content.ts
Jakeii: good idea

PR #862, opened 2023-04-25
fix: Allow Teads on liveblog server-side inserted inline ads

domlander:
## What does this change?

Adds the Outstream ad sizes to server-side inserted inline ads.

## Why?

So that we can show Outstream ads (Teads) on liveblog pages when the advert slots are inserted server-side.

### DIFF
--- a/bundle/src/projects/commercial/modules/dfp/fill-advert-slots.ts
+++ b/bundle/src/projects/commercial/modules/dfp/fill-advert-slots.ts
@@ -19,6 +19,7 @@ import { queueAdvert } from './queue-advert';
 const decideAdditionalSizes = (adSlot: HTMLElement): SizeMapping => {
 	const { contentType } = window.guardian.config.page;
 	const { name } = adSlot.dataset;
+
 	if (contentType === 'Gallery' && name?.includes('inline')) {
 		return {
 			desktop: [adSizes.billboard, createAdSize(900, 250)],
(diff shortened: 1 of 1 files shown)


chrislomaxjones (approved):
Makes sense!

bundle/src/projects/commercial/modules/dfp/fill-advert-slots.ts
```diff
@@ -30,6 +31,11 @@ const decideAdditionalSizes = (adSlot: HTMLElement): SizeMapping => {
 		return {
 			desktop: [adSizes.billboard],
 		};
+	} else if (contentType === 'LiveBlog' && name?.includes('inline')) {
```
arelra, line 34: This is just a nice to have but wondering if we could update the `contentType` type to be a string literal of the values we expect.

arelra (approved):

Jakeii (approved):

bundle/src/projects/commercial/modules/dfp/fill-advert-slots.ts
domlander: The `contentType` type is used in quite a lot of places and in multiple projects and I couldn't find anywhere that provides a definitive list of all possible values. It looks to be non-trivial to update this type to be more specific than `string` so I'll leave it as is for now.

PR #863, opened 2023-04-27
ci: Remove coveralls from ci-core action

SiAdcock:
## What does this change?

Remove coveralls from ci-core action

## Why?

Coveralls is not currently bringing us value. The CI check in GitHub often hangs with the message "Expected — Waiting for status to be reported". We have disabled the check on GitHub PRs. This change deletes the action from the CI workflow.


arelra (approved):
🧹

PR #869, opened 2023-05-10
fix: Add <ul> to spacefinder ignore list

domlander:
## What does this change?

This change removes `<ul>` elements from the list of elements that Spacefinder avoids placing ads close to. Spacefinder will now be able to place ads close to a bulleted list.

## Why?

Some elements in an article are not suitable to place ads next to or close to. For example, an advert too close to a full-width image would not look good when rendered on the page. A bulleted list is regular article content and should not be avoided when trying to place ads. This change will not insert ads immediately above a bulleted list, as adverts are only inserted above paragraph `p` tags.

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://github.com/guardian/commercial/assets/9574885/c90b1752-ab5d-497d-bd15-58f740f2052c
[after]: https://github.com/guardian/commercial/assets/9574885/2f63aaa0-f7b4-49bf-b438-2a5adcc0db07


arelra (approved):
👾

PR #873, opened 2023-05-11
feat: Add front-banner ad slot

domlander:
## What does this change?

Adds the `front-banner` ad slot type.

## Why?

So that we can target this new type of slot with banner style ads on fronts pages

## Screenshots

image

### DIFF
--- a/bundle/src/projects/commercial/modules/header-bidding/request-bids.ts
+++ b/bundle/src/projects/commercial/modules/header-bidding/request-bids.ts
@@ -8,7 +8,10 @@ const retainTopAboveNavSlotSize = (
 	advertSize: Advert['size'],
 	hbSlot: HeaderBiddingSlot,
 ): HeaderBiddingSlot[] => {
-	if (hbSlot.key !== 'top-above-nav') {
+	if (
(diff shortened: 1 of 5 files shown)


bundle/src/projects/commercial/modules/header-bidding/request-bids.ts
```diff
@@ -8,7 +8,10 @@ const retainTopAboveNavSlotSize = (
 	advertSize: Advert['size'],
 	hbSlot: HeaderBiddingSlot,
 ): HeaderBiddingSlot[] => {
-	if (hbSlot.key !== 'top-above-nav') {
+	if (
+		hbSlot.key !== 'top-above-nav' &&
+		!hbSlot.key.includes('front-banner')
```
arelra, line 13: This is a very minor suggestion but `.startsWith` might read better here

arelra (approved):

bundle/src/projects/commercial/modules/header-bidding/request-bids.ts
domlander: Good idea

PR #874, opened 2023-05-15
fix!: Do not load opt-out if the user is a digital subscriber 

arelra:
## What does this change?

Do not load opt-out if the user is a digital subscriber


bundle/src/bootstraps/standalone.commercial.ts
```diff
@@ -212,48 +211,43 @@ const bootCommercial = async (): Promise<void> => {
...
-		}
-
+	// Only load the Opt Out tag if:
+	// - in TCF region
+	// - no consent for Googletag
+	// - the user is not a subscriber
+	if (
+		consentState.tcfv2 &&
+		!getConsentFor('googletag', consentState) &&
+		!isDigitalSubscriber()
```
Jakeii, line 234: I don't think `isDigitalSubscriber` should be used directly, I think it's just a part of why someone might be ad free.

Jakeii: Actually nevermind I think I'm thinking of one of the other identity cookies

Jakeii (approved):

emma-imber (approved):
This feels much clearer now 🎉

PR #875, opened 2023-05-15
fix: Add labels to consentless merchandising slots

emma-imber:
## What does this change?

- Fixes the beta release action by using a slightly older version of node to preform the beta release
- Changes the consentless label rendering logic to add labels to merchandising slots, as these are used to display ads in consentless, instead of in house content

Before:
Screenshot 2023-05-16 at 17 50 31

After:
Screenshot 2023-05-17 at 10 09 50


arelra (approved):
Thanks for fixing beta releases too!

PR #877, opened 2023-05-18
feat: Add fronts-banner ads to Frontend UK Network Front

domlander:
## What does this change?

Add fronts-banner ads to Frontend UK Network Front.

Ads will be inserted above the following sections: `opinion`, `sport`, `around-the-world`, `lifestyle`, `in-pictures`.

This functionality is behind an [AB test](https://github.com/guardian/frontend/pull/26117) which looks to test a new ad strategy on Fronts pages. For this test, the new adverts will only be inserted in the following conditions:
- The page is the UK Network Front, i.e. https://www.theguardian.com/uk
- The page is not rendered by DCR
- The viewport size is above the desktop (980px) breakpoint.

## Why?

So that we can gather results for an [AB test](https://github.com/guardian/frontend/pull/26117) regarding the commercial impact of a new ad strategy.

## Screenshots

image

### DIFF
--- a/bundle/src/projects/commercial/modules/fronts-banner-adverts.ts
+++ b/bundle/src/projects/commercial/modules/fronts-banner-adverts.ts
@@ -61,7 +61,7 @@ export const init = async (): Promise<void> => {
 	}
 
 	const promises = sections.map(async (section, index) => {
-		await insertAdvertAboveSection(section, index + 1);
+		return insertAdvertAboveSection(section, index + 1);
 	});
 
 	await Promise.allSettled(promises);
--- a/bundle/src/projects/commercial/modules/fronts-banner-adverts.ts
+++ b/bundle/src/projects/commercial/modules/fronts-banner-adverts.ts
@@ -1,4 +1,4 @@
-import { createAdSlot, slotSizeMappings } from '@guardian/commercial-core';
+import { createAdSlot } from '@guardian/commercial-core';
 import fastdom from '../../../lib/fastdom-promise';
 import { addSlot } from './dfp/add-slot';
 
@@ -34,20 +34,14 @@ const insertAdvertAboveSection = async (section: string, advertNum: number) => {
 					sectionNode,
 				);
 			})
-			.then(async () => {
-				await addSlot(ad, false, slotSizeMappings['fronts-banner']);
+			.then(() => {
+				void addSlot(ad, false);
 			});
 	});
 };
 
 export const init = async (): Promise<void> => {
 	const { switches, tests } = window.guardian.config;
-	if (
-		!switches.frontsBannerAds ||
-		tests?.frontsBannerAdsVariant !== 'variant'
-	) {
-		return Promise.resolve();
-	}
 
 	const {
 		isDotcomRendering,
@@ -56,13 +50,19 @@ export const init = async (): Promise<void> => {
 
 	const isUkNetworkFront =
 		contentType === 'Network Front' && edition === 'UK';
-	if (!isUkNetworkFront || isDotcomRendering) {
-		return Promise.resolve();
+
+	if (
+		!switches.frontsBannerAds ||
+		tests?.frontsBannerAdsVariant !== 'variant' ||
+		!isUkNetworkFront ||
+		isDotcomRendering
+	) {
+		return;
 	}
 
-	sections.map(async (section, index) => {
+	const promises = sections.map(async (section, index) => {
 		await insertAdvertAboveSection(section, index + 1);
 	});
 
-	return Promise.resolve();
+	await Promise.allSettled(promises);
 };
--- a/bundle/src/projects/commercial/modules/fronts-banner-adverts.ts
+++ b/bundle/src/projects/commercial/modules/fronts-banner-adverts.ts
@@ -1,0 +1,68 @@
+import { createAdSlot, slotSizeMappings } from '@guardian/commercial-core';
+import fastdom from '../../../lib/fastdom-promise';
+import { addSlot } from './dfp/add-slot';
+
+/**
+ * A fronts banner advert will be inserted above each of the following sections.
+ */
+const sections = [
+	'opinion',
+	'sport',
+	'around-the-world',
+	'lifestyle',
+	'in-pictures',
+];
+
+const insertAdvertAboveSection = async (section: string, advertNum: number) => {
+	await fastdom.measure(() => {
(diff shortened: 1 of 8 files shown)


bundle/src/projects/commercial/modules/fronts-banner-adverts.ts
```diff
@@ -0,0 +1,68 @@
...
+	const {
+		isDotcomRendering,
+		page: { contentType, edition },
+	} = window.guardian.config;
+
+	const isUkNetworkFront =
+		contentType === 'Network Front' && edition === 'UK';
+	if (!isUkNetworkFront || isDotcomRendering) {
+		return Promise.resolve();
+	}
```
Jakeii, line 61: Could these conditions be combined?

bundle/src/projects/commercial/modules/fronts-banner-adverts.ts
```diff
@@ -0,0 +1,68 @@
...
+
+		void fastdom
+			.mutate(() => {
+				sectionNode.parentElement?.insertBefore(
+					adContainer,
+					sectionNode,
+				);
+			})
+			.then(async () => {
+				await addSlot(ad, false, slotSizeMappings['fronts-banner']);
```
arelra, line 38: Don't _think_ you need an await here and could replace with `void`

Jakeii, line 38: I think the third argument isn't needed, `addSlot` will look up from the `slotSizeMappings `itself.

bundle/src/projects/commercial/modules/header-bidding/types.d.ts
```diff
@@ -11,9 +11,13 @@ declare global {
 		| 'mostpop'
 		| 'right'
 		| 'top-above-nav'
+		| `fronts-banner-${number}`
```
Jakeii, line 14: could we lose the `-` before the number for consistency with `inline`?

Jakeii (commented):
Nicely done!🚀 Just a few suggestions

bundle/src/projects/commercial/modules/fronts-banner-adverts.ts
```diff
@@ -0,0 +1,68 @@
...
+	});
+};
+
+export const init = async (): Promise<void> => {
+	const { switches, tests } = window.guardian.config;
+	if (
+		!switches.frontsBannerAds ||
+		tests?.frontsBannerAdsVariant !== 'variant'
+	) {
+		return Promise.resolve();
```
Jakeii, line 49: _tiiiny_ one, you can just return here and on line `60` in an async function, I'm sure there should be a eslint rule for this.

bundle/src/projects/commercial/modules/fronts-banner-adverts.ts
domlander: Sure. My intention was an early return for those not in the test but it's probably not worth it. Plus, we're never going to just remove the test from this piece of code so it's also not a benefit to separate the conditions out for that reason either. Will amend.

bundle/src/projects/commercial/modules/header-bidding/types.d.ts
domlander: I did have a think about this, but to me I don't think that `fronts-banner1` is in itself a consistent format without the dash before the number. I don't think that it would be worth maintaining consistency with the `inline` slots. Happy to change if I'm outvoted though.

bundle/src/projects/commercial/modules/fronts-banner-adverts.ts
```diff
@@ -0,0 +1,68 @@
...
+		if (!sectionNode) return;
+
+		const ad = createAdSlot('fronts-banner', {
+			name: `fronts-banner-${advertNum}`,
+			classes: 'fronts-banner',
+		});
+
+		const adContainer = document.createElement('div');
+		adContainer.className = 'ad-slot-container fronts-banner-container';
+		adContainer.appendChild(ad);
```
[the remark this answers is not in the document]
domlander: I don't _think_ that this element is in the DOM at this point. I'm going by the last comment in the example [here](https://developer.mozilla.org/en-US/docs/Web/API/Document/createElement#basic_example). I might be wrong though so please correct me if I am.

bundle/src/projects/commercial/modules/header-bidding/types.d.ts
Jakeii: I guess I see both arguments now, I leave it to you 👍

Jakeii (approved):

bundle/src/projects/commercial/modules/fronts-banner-adverts.ts
```diff
@@ -0,0 +1,68 @@
...
+		isDotcomRendering
+	) {
+		return;
+	}
+
+	const promises = sections.map(async (section, index) => {
+		await insertAdvertAboveSection(section, index + 1);
+	});
+
+	await Promise.allSettled(promises);
```
Jakeii, line 67: Now I see you could use the return early pattern, not a sticking point though!

```suggestion
	if (
		switches.frontsBannerAds ||
		tests?.frontsBannerAdsVariant === 'variant' ||
		isUkNetworkFront ||
		!isDotcomRendering
	) {
		return Promise.allSettled(
			sections.map(
				(section, index) => insertAdvertAboveSection(section, index + 1)
			)
		);
	}
	
```

bundle/src/projects/commercial/modules/fronts-banner-adverts.ts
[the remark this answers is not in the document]
arelra: Ah yes you're right I was mixing up my append and inserts! 👍

arelra (approved):

bundle/src/projects/commercial/modules/fronts-banner-adverts.ts
```diff
@@ -0,0 +1,68 @@
...
+		!switches.frontsBannerAds ||
+		tests?.frontsBannerAdsVariant !== 'variant' ||
+		!isUkNetworkFront ||
+		isDotcomRendering
+	) {
+		return;
+	}
+
+	const promises = sections.map(async (section, index) => {
+		await insertAdvertAboveSection(section, index + 1);
```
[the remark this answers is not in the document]
domlander: Yes this makes sense! Thanks, I had been in a muddle with this.

PR #886, opened 2023-05-31
fix: build dev & prod by default

Jakeii:
## What does this change?
make `yarn build` build both the dev and prod bundles

## Why?
Frontend when running locally is requesting the bundle without a hash, simply copying the file without the hash doesn't work as the separated chunks will still be requested with the hash in the URL (prebid.js and consentless), probably because webpack works out the paths at compile time?

A quick solution is to build both dev and prod so that they both will end up in the npm package and either can be used.


chrislomaxjones (approved):
👍

PR #888, opened 2023-06-01
test: Test reliability of navigator.sendBeacon()

emma-imber:
## What does this change?
For 1% of calls to amiused, send a fetch keepalive = true POST request as well as navigator.sendBeacon() for amiused logging. Each will be labelled so that we can query the data for these specific test calls. Keeping the sampling percent low to start so we don't crowd the logs too much, and we can always increase later on if not enough data is coming through.

## Why?
This will allow us to compare the number of logs received for each send method, giving us an idea of how reliable navigator.sendBeacon() is compared to sending a fetch request. Some articles online suggest that navigator.sendBeacon() may not be as reliable as we think: https://volument.com/blog/sendbeacon-is-broken

### DIFF
--- a/.github/workflows/beta-release-on-label.yml
+++ b/.github/workflows/beta-release-on-label.yml
@@ -45,7 +45,7 @@ jobs:
         run: yarn build
 
       - name: Grab last version
-        run: yarn version --new-version $(git tag -l | grep @guardian/commercial-v\* |  sed -E 's|.*([0-9]+\.[0-9]+\.[0-9]+(-beta.[0-9]+)?)|\1|g' | tail -1) --no-git-tag-version
+        run: yarn version --new-version $(git tag -l | grep v\* |  sed -E 's|.*([0-9]+\.[0-9]+\.[0-9]+(-beta.[0-9]+)?)|\1|g' | tail -1) --no-git-tag-version
 
       - name: Bump prerelease version
         run: yarn version --prerelease --preid beta --no-git-tag-version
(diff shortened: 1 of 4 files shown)


arelra:
> 🚀 @guardian/commercial-v7.1.1-beta.3 published to npm as a beta release

Hmm the beta labelling should be v8.x-beta.x 🤔

emma-imber:
> 

Yes I noticed this too! I tried running the grep statement locally that the beta release code uses, and it seems to think the latest version is the beta release from your bundle + core merge PR and it's incrementing from that. It doesn't seem to be able to pick up the v8 release?

arelra:
> > > 🚀 @guardian/commercial-v7.1.1-beta.3 published to npm as a beta release
> > 
> > 
> > Hmm the beta labelling should be v8.x-beta.x 🤔
> 
> Yes I noticed this too! I tried running the grep statement locally that the beta release code uses, and it seems to think the latest version is the beta release from your bundle + core merge PR and it's incrementing from that. It doesn't seem to be able to pick up the v8 release?

Thanks for the fix :)

arelra (approved):
Just wondering if we are capturing or able to join browser data later as this might be useful to @guardian/open-journalism @mxdvl and their investigation of holes in CWV collection.

mxdvl:
> Just wondering if we are capturing or able to join browser data later as this might be useful to @guardian/open-journalism @mxdvl and their investigation of holes in CWV collection.

I doubt the issue is with the mechanism for sending data over the wire, [as we still get a lot information for TTFB and LCP](https://docs.google.com/spreadsheets/d/1wLf_0CA27m2XYfLQeB9kCfwtNTHErIwxQik_Cbd8qX8/edit)

PR #891, opened 2023-06-02
feat: Bake commit hash into bundle

chrislomaxjones:
## What does this change?

This PR bakes the commit SHA into the bundle and logs it to the browser console.

Screenshot 2023-06-12 at 10 46 34

## Why?

This should make it easier to tell which version of the Commercial bundle is running on CODE/Production.

### DIFF
--- a/.changeset/stale-pets-remember.md
+++ b/.changeset/stale-pets-remember.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
+Log the commit SHA of the current build to the browser console
--- a/src/standalone.commercial.ts
+++ b/src/standalone.commercial.ts
@@ -173,6 +173,12 @@ const recordCommercialMetrics = () => {
 
 const bootCommercial = async (): Promise<void> => {
 	log('commercial', '📦 standalone.commercial.ts', __webpack_public_path__);
+	if (process.env.COMMIT_SHA) {
+		log(
+			'commercial',
+			`@guardian/commercial commit ${process.env.COMMIT_SHA}`,
+		);
+	}
 
 	//adding an amiused call for a very small proportion of users to test sendBeacon vs fetch
 	//this will be removed when we have enough data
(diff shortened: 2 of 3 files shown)


arelra (approved):
🎉 

This is just a nice to have but would it be useful to have a comment at the top of `standalone.commercial.ts` which we could replace using webpack's [define plugin](https://webpack.js.org/plugins/define-plugin/#usage) with the package version and the commit hash e.g.

```
standalone.commercial.ts
// PACKAGE_VERSION
// COMMIT_SHA
```

and

```
new webpack.DefinePlugin({
  PACKAGE_VERSION: ...,
  COMMIT_SHA: ...,
});
```

I think this would only affect the built dist output so it wouldn't touch the original source files.

chrislomaxjones:
> This is just a nice to have but would it be useful to have a comment at the top of `standalone.commercial.ts` which we could replace using webpack's [define plugin](https://webpack.js.org/plugins/define-plugin/#usage) with the package version and the commit hash

Nice idea! I was actually inspired by poking around in Figma's browser console, and it looks like they do something similar. But I think just replacing with comments is a nice idea.

I'll have a think about the package version, is it easy to know the version number before you've actually created the bundles?

arelra:
> I'll have a think about the package version, is it easy to know the version number before you've actually created the bundles?

That's a good point we would need the package version set before building or somehow derive it. The commit sha is a good start though and we could follow up with the package version.

@Jakeii is looking at changesets which might make it easier aiui it will bump versions as part of the changeset.

Jakeii:
> > I'll have a think about the package version, is it easy to know the version number before you've actually created the bundles?
> 
> That's a good point we would need the package version set before building or somehow derive it. The commit sha is a good start though and we could follow up with the package version.
> 
> @Jakeii is looking at changesets which might make it easier aiui it will bump versions as part of the changeset.

The actual version does get written to package.json with changesets, so I think in the github workflow it could version using the `yarn changeset version`, store the new version in an environment variable, webpack can is now 'aware' of the version and can use it

Jakeii (approved):
Tiiiny enhancement, could the log be a link to the commit in github?

PR #900, opened 2023-06-13
Configure Prebid on fronts banner ads

domlander:
## What does this change?

- Fixes a bug that enables prebid to be run on fronts banner ads.
- Configures the following vendors to be able to identify and bid on fronts-banner-ads: Improve; Pubmatic; Appnexus; Index; Smart; Criteo.

## Why?

So that we can run header-bidding on the fronts banner ads

### DIFF
--- a/src/projects/commercial/modules/header-bidding/prebid/bid-config.ts
+++ b/src/projects/commercial/modules/header-bidding/prebid/bid-config.ts
@@ -14,6 +14,7 @@ import {
 	isInUsOrCa,
 } from '../../../../common/modules/commercial/geo-utils';
 import type {
+	BidderCode,
 	HeaderBiddingSize,
 	PrebidAdYouLikeParams,
 	PrebidAppNexusParams,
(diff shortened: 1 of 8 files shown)


src/projects/commercial/modules/header-bidding/prebid/bid-config.ts
```diff
@@ -464,8 +417,17 @@ const smartBidder: PrebidBidder = {
 // There's an IX bidder for every size that the slot can take
 const indexExchangeBidders = (
 	slotSizes: HeaderBiddingSize[],
+	isInVariant: boolean,
 ): PrebidBidder[] => {
-	const indexSiteId = getIndexSiteId();
+	let indexSiteId = getIndexSiteId();
+	if (isInVariant) {
```
emma-imber, line 423: Non-blocking, but it would be nice to have a more specific parameter name than 'isInVariant' to specify exactly which variant we're talking about, and for more consistency with the rest of the code

emma-imber (approved):
One small comment but looks good!

Jakeii (approved):

src/projects/commercial/modules/header-bidding/prebid/bid-config.ts
domlander: That's a good point. I have updated

PR #903, opened 2023-06-14
build both dev and prod to subdirectories in the dist folder

Jakeii:
## What does this change?
Move built bundle files to `prod` and `dev` subdirectories in the dist/bundle folder and have both in npm, so they can both be used, this is a breaking change as the path to the files/entrypoints will change.

## Why?
frontend when running locally wants the unhashed files to be available, webpack bakes in the hashes for chunks like prebid and consentless so we need to provide a build without them.

### DIFF
--- a/.changeset/witty-pianos-give.md
+++ b/.changeset/witty-pianos-give.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': major
+---
+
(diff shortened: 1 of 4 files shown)


domlander (approved):

chrislomaxjones (approved):
👍

PR #906, opened 2023-06-15
Enhance sendBeacon testing

emma-imber:
## What does this change?
- Set default sendBeacon test sampling rate to zero, and update the function test file to reflect that
- Call amiused in the bootCommercial function for 0.01% of cases, to test the reliability of sendBeacon vs fetch
- Send the UserAgent details in the amiused testing logs

## Why?
We will no longer use the actual amiused call (in spacefinder.ts) to test the reliability of sendBeacon vs fetch. This is because we don't know if there might be bias in the situations in which this function is called, perhaps in the browsers or different user setups, and this could lead to bias in our results. Instead, we will call amiused 0.01% of the times that the bootCommercial function in the standalone file runs. We know this script should run for all users who see google ads, and this should give us a more reliable sample of users.

We're setting the sampling for this test to 1 in 10,000 users (0.01%) to start, and will increase or decrease if we're not satisfied with the test numbers that we have coming through.

We are also going to include the UserAgent details in the test calls, so that we can spot any potential patterns between browsers and the respective reliability of sendBeacon vs fetch.

Our data comes through in BigQuery containing the user agent details and a label showing which method was used to send the data:
Screenshot 2023-06-15 at 15 59 04

### DIFF
--- a/src/projects/commercial/am-i-used.ts
+++ b/src/projects/commercial/am-i-used.ts
@@ -28,7 +28,7 @@ const amIUsed = (
 	parameters?: Partial<
 		Record<string, string> & Record<RestrictedKeys, never>
 	>,
-	sampling: number = 1 / 100,
+	sampling = 0,
 ): void => {
 	// The function will return early if the sentinelLogger switch is disabled.
 	if (!window.guardian.config.switches.sentinelLogger) return;
--- a/src/projects/commercial/am-i-used.ts
+++ b/src/projects/commercial/am-i-used.ts
@@ -56,6 +56,9 @@ const amIUsed = (
 			: properties,
 	};
 
+	//The code below is for testing the reliability of the sendBeacon method compared to fetch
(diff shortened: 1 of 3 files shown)


src/projects/commercial/am-i-used.ts
domlander, line 1: Could you add a comment above line 59 to explain a bit about why we are testing the beacon? Something like...

`We are testing the reliability of navigator.sendBeacon() compared to sending a fetch request. Some articles online suggest that navigator.sendBeacon() may not be as reliable as we think: https://volument.com/blog/sendbeacon-is-broken.`

P.S. Those words were mostly stolen from your previous PR!

domlander (approved):
👍

src/projects/commercial/am-i-used.ts
emma-imber: Great idea - have added a comment!

PR #914, opened 2023-06-21
Try updating reference to Github API to fix beta release version comments

emma-imber:
## What does this change?
Updates the reference to the Github API in the beta release workflow.

## Why?
This is so that the beta release flow can comment the version on pull requests. The previous way was deprecated and was causing errors when creating beta releases.

### DIFF
--- a/.changeset/kind-tables-build.md
+++ b/.changeset/kind-tables-build.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
+Update beta release workflow command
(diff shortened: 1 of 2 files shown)


Jakeii (approved):
JINKS! #915

emma-imber:
> #915

Haha nice! 🏎

Jakeii:
For something like this that would have been a chore/docs in semantic-release land you can leave out the changeset if you like :)

emma-imber:
> For something like this that would have been a chore/docs in semantic-release land you can leave out the changeset if you like :)

Ah good to know, thanks! Just wanted to make sure that the comment popped up on the PR

PR #915, opened 2023-06-21
fix github-script in label beta release workflow

Jakeii:
## What does this change?
This was a [breaking change](https://github.com/actions/github-script#breaking-changes-in-v5) that I didn't notice when upgrading github-script to v6


vlkerag:
[Upgrade to node 18, upgrade to ts 5.1.3, upgrade as many dependencies as possible](https://trello.com/c/N042RtTX/1042-upgrade-to-node-18-upgrade-to-ts-513-upgrade-as-many-dependencies-as-possible)

PR #921, opened 2023-06-27
Remove the non refresh line item fallback

chrislomaxjones:
- Remove usage of fetchNonRefreshableLineItems
- Ensure the final should-refresh test fails now
- Update should-refresh test and test newer method


Jakeii (approved):
Looks good to me!

PR #925, opened 2023-06-28
Prune unused exports and modules

Jakeii:
## What does this change?
Used `ts-prune` to find unused exports and modules, left a few that are sometimes used e.g. some AB testing stuff.

## Why?

Thanks to tree-shaking this doesn't really have an effect on bundle size, but still, deleting useless code is good right?


arelra (approved):
Great!

I was expecting more dead code in user-features and friends but good to see lots of deleted code.

Jakeii:
@arelra great point, there is a bunch of unused stuff in user-features, but not being picked up by this tool.

PR #936, opened 2023-07-05
Visual tests should only be manually triggered

chrislomaxjones:
## What does this change?

This PR changes the Percy visual testing Github Action to only be triggered manually from the Github UI.

## Why?

This action is not currently required for a build to pass, and often errors. This prevents the workflow from running when its not required.


chrislomaxjones:
cc @arelra

Jakeii (approved):

PR #940, opened 2023-07-10
Replace `switches.commercial` with `switches.shouldLoadGoogletag`

chrislomaxjones:
## What does this change?

This PR updates the commercial code to use the new switch introduced in https://github.com/guardian/frontend/pull/26295. We also update the commercial-features to use an equivalent name.

As a small piece of drive-by refactoring, I've edited a couple of tests to not use `config.set`, and to remove some unecessary mocks.

## Why?

This new switch should have a clearer behaviour than the previous one, and be the same across Frontend and DCR.

### DIFF
--- a/.changeset/famous-teachers-divide.md
+++ b/.changeset/famous-teachers-divide.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
+Use the new `shouldLoadGoogletag` switch.


src/lib/commercial-features.spec.ts
```diff
@@ -105,36 +106,36 @@ describe('Commercial features', () => {
 	describe('DFP advertising', () => {
 		it('Runs by default', () => {
 			const features = new CommercialFeatures();
-			expect(features.dfpAdvertising).toBe(true);
+			expect(features.shouldLoadGoogletag).toBe(true);
```
arelra, line 109: I don't mind if you want to stick with 'should' but this is a bit shorter?

```suggestion
			expect(features.loadGoogletag).toBe(true);
```

src/lib/header-bidding/prebid/prebid.spec.ts
```diff
@@ -32,7 +32,6 @@ const resetPrebid = () => {
 describe('initialise', () => {
 	beforeEach(() => {
 		resetPrebid();
-		window.guardian.config.switches.commercial = true;
```
arelra, line 35: We don't need to replace with `switches.shouldLoadGoogletag` here?

arelra (approved):
lgtm

src/lib/header-bidding/prebid/prebid.spec.ts
chrislomaxjones: It's not actually required in this test, so I removed it entirely. I think it's only necessary in prepare-prebid.spec.ts

src/lib/commercial-features.spec.ts
chrislomaxjones: I think I'll leave it as is for now, just to keep it aligned with the switch. I do appreciate it's a little long though, so maybe we can rename it across Frontend/Commercial once this has settled?

PR #947, opened 2023-07-17
Liveblog right hand ad test

domlander:
## What does this change?

- Adds the `liveblog-right` ad size
- Creates a script that will fill the new `liveblog-right` ad slots with ads, when a custom event is received 
- Remove superfluous definitions from slotSizeMappings. A set of ad-sizes defined for a breakpoint will apply to every screen size above this breakpoint, so redefining the same set of ad-sizes for tablet as well as mobile is not needed.
- Remove commitizen from list of recommended code extensions
- Rename `retainTopAboveNavSlotSize` to `retainAdSizeOnRefresh`

## Why?

- We will be testing inserting multiple ads in to the right-hand column of mobile pages, so we need a new ad slot to identify these ads. Currently, the ad in the top-right is called the `right` ad. I didn't name these new slots `right-x`, since we are only making a change to liveblog pages and it is not possible that there will be multiple right-ads on article pages (Since the ads that are currently pushed out to the right are `inline-2+`).
- It is not a team custom to use the commitizen extension, so we shouldn't explicitly recommend to new devs.

## Other PR's

- [Frontend](https://github.com/guardian/frontend/pull/26332)
- [Dotcom-rendering](https://github.com/guardian/dotcom-rendering/pull/8289)


src/lib/spacefinder/liveblog-right-adverts.ts
```diff
@@ -0,0 +1,75 @@
...
+type AdsInsertedCustomEventDetail = {
+	numAdsToInsert: number;
+	fromIndex: number;
+};
+
+type AdsInsertedCustomEvent = {
+	detail: AdsInsertedCustomEventDetail;
+};
+
+const isCustomEvent = (event: Event): event is CustomEvent => {
```
domlander, line 19: A function of type (e: CustomEvent) => void is not a valid instance of EventListener, so I had to use Event then use this type guard

https://stackoverflow.com/questions/47166369/argument-of-type-e-customevent-void-is-not-assignable-to-parameter-of-ty/47171216#47171216

src/lib/spacefinder/liveblog-right-adverts.ts
```diff
@@ -0,0 +1,75 @@
...
+		dfpEnv.advertIds[advert.id] = currentLength + index;
+	});
+
+	if (isInEagerPrebidVariant()) {
+		// Request bids for all server rendered adverts
+		await requestBidsForAds(adverts);
+	}
+
+	adverts.forEach(queueAdvert);
+	if (dfpEnv.shouldLazyLoad()) {
```
chrislomaxjones, line 56: I think you can avoid duplicating the code that directly manipulates `dfpEnv.adverts` and calls `displayAds` here by calling [`addSlot`](https://github.com/guardian/commercial-core/blob/d0ec41f494be36556ad1a25a889556e5b3cce921/src/lib/dfp/add-slot.ts#L20). I think this is the pattern taken by `article-body-adverts.ts` and `liveblog-adverts.ts`.

EDIT: It'll handle the call to `createAdvert` as well.

src/lib/spacefinder/liveblog-right-adverts.ts
```diff
@@ -0,0 +1,75 @@
...
+		displayAds();
+	}
+};
+
+// When a liveblog-right ad slot is inserted into the page, create an ad in this new slot
+export const initLiveblogRightAdverts = (): Promise<void> => {
+	document.addEventListener('liveblog-right-ads-inserted', (event: Event) => {
+		if (!isCustomEvent(event)) throw new Error('not a custom event');
+
+		const { numAdsToInsert, fromIndex } = (<AdsInsertedCustomEvent>event)
```
chrislomaxjones, line 68: This may become clearer in subsequent PRs, but by what means is this event going to be fired? Will it be from somewhere in the Commercial bundle, or is DCR going to be tracking when to insert ads?

Also, just to check my understanding, this code is only responsible for filling the slots in response to events, but some other system is going to be inserting the slots into the DOM?

src/lib/spacefinder/liveblog-right-adverts.ts
```diff
@@ -0,0 +1,75 @@
...
+
+	const adverts = [
+		...document.querySelectorAll<HTMLElement>('.ad-slot--liveblog-right'),
+	]
+		.filter(({ id }) =>
+			// exclude ad slots that have already been filled
+			newAdvertIndexes.includes(Number(id[id.length - 1])),
+		)
+		.map((adSlot) => createAdvert(adSlot))
+		.filter((advert): advert is Advert => advert !== null);
```
chrislomaxjones, line 40: There's a function [`isNonNullable`](https://github.com/guardian/csnx/blob/main/libs/%40guardian/libs/src/isNonNullable/isNonNullable.ts) in libs for this use case.

src/lib/header-bidding/request-bids.ts
```diff
@@ -4,11 +4,11 @@ import { a9 } from './a9/a9';
 import { prebid } from './prebid/prebid';
 import type { HeaderBiddingSlot } from './prebid-types';
 
-// Force the refreshed advert to be the same size as the first
-const retainTopAboveNavSlotSize = (
+const retainAdSizeOnRefresh = (
```
chrislomaxjones, line 7: Thanks for renaming this and updating the comments, it's a lot clearer.

chrislomaxjones (commented):
Nice work!

I was also curious about the AB test setup for this one, as there isn't a test definition or variant check here. Is DCR going contain the test, and will only fire this custom event on pageviews in the variant?

src/lib/spacefinder/liveblog-right-adverts.ts
domlander: Good idea, will update. 

It does feel like `addSlot` is a weird name. The function takes a slot and fills that slot with an advert, but `addSlots` sounds like we're adding a slot to the page. What do you think about changing this name to `fill-advert-slot-dynamic` or `fill-dynamic-advert-slot`, as it looks to do a similar task to https://github.com/guardian/commercial/blob/main/src/lib/dfp/fill-advert-slots.ts which is filling slots with ads?

domlander:
> I was also curious about the AB test setup for this one, as there isn't a test definition or variant check here. Is DCR going contain the test, and will only fire this custom event on pageviews in the variant?

That was the plan, as the only thing that will change following this PR is that an event listener will be created but never fired, so I didn't include the test. However, now you mention it I think it would be a good idea to create the test and wrap the event listener with it, so that we can be confident that by turning the test off the liveblog ads cannot be inserted

src/lib/spacefinder/liveblog-right-adverts.ts
chrislomaxjones: Emphatic +1 from me, we can do better than `addSlot` (it's also hard to tell if someone is saying adSlot or addSlot when speaking aloud).

`fillDynamicAdSlot` sounds good!

src/lib/experiments/tests/liveblog-right-column-ads.ts
```diff
@@ -0,0 +1,36 @@
+import type { ABTest } from '@guardian/ab-core';
+
+export const liveblogRightColumnAds: ABTest = {
+	id: 'LiveblogRightColumnAds',
+	author: '@commercial-dev',
+	start: '2023-07-15',
+	expiry: '2023-09-20',
+	audience: 0 / 100,
+	audienceOffset: 0 / 100,
+	audienceCriteria: 'Desktop users with wide (1300px+) screens only',
```
chrislomaxjones, line 10: Do you want to use the `canRun` predicate below to constrain to an eligible set of pageviews? E.g. desktop liveblogs or something like that.

chrislomaxjones (approved):
Looking good, will be interested in the results for this one. Had a couple of non-blocking comments below.

This is also non-blocking, but I'm interested in the approach of having DCR measure the space available for adverts, inserting the slots into the DOM, and firing a custom event. Is there any particular reason you chose to implement it this way? To be clear, there's no right or wrong with the approach, it's just a departure from `article-body-adverts` and `liveblog-adverts`, in which the commercial code is responsible for all measurement and slot insertion.

src/lib/spacefinder/liveblog-right-adverts.ts
chrislomaxjones: Thanks for sharing the link, I've had a quick look over this and left a separate comment below.

src/lib/experiments/tests/liveblog-right-column-ads.ts
domlander: Ah yes good spot thanks! We also don't need to check for the showcase element here, as that isn't used on liveblogs

src/lib/header-bidding/slot-config.ts
```diff
@@ -121,6 +125,17 @@ const getSlots = (): HeaderBiddingSizeMapping => {
...
+			desktop: hasShowcaseMainElement
+				? [adSizes.mpu]
+				: [adSizes.halfPage, adSizes.mpu],
+			tablet: hasShowcaseMainElement
+				? [adSizes.mpu]
+				: [adSizes.halfPage, adSizes.mpu],
+			mobile: hasShowcaseMainElement
+				? [adSizes.mpu]
+				: [adSizes.halfPage, adSizes.mpu],
+		},
```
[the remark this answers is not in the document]
domlander: Ah yes good spot thanks! We also don't need to check for the showcase element here, as that isn't used on liveblogs

src/lib/experiments/tests/liveblog-right-column-ads.ts
chrislomaxjones: Hmm it might be more effort than it's worth, and easier to just filter pageviews in the analysis.

chrislomaxjones:
Thanks for the explanation! We can definitely revisit once we've got this out there and validated the hypothesis. It may very well be a pattern we want to adopt more widely.

PR #948, opened 2023-07-17
Allow overriding the data model in E2E tests

chrislomaxjones:
## Motivation

It'd be useful to be able to run end-to-end / integration tests where we fix the data model we send to DCR.

1. When the `shouldLoadGoogletag` switch is switched on, ads load on the page.
2. When the `shouldLoadGoogletag` switch is switched off, ads do not load on the page.

Therefore, we want to ignore the production value for the `shouldLoadGoogletag` switch and override it with a value we have fixed for the test.

The test that implements that as part of this PR is located in [googletag-switch.ts](https://github.com/guardian/commercial/blob/39fd2ab93ff3feae54ee2e4c1db745c38f9e33d3/cypress/e2e/parallel-2/googletag-switch.cy.ts) as part of this PR.

## The approach

In order to achieve the above, this PR adds a _fixtures server_ to our existing webpack dev server setup. This has an endpoint that accepts GET requests of the form:

```
http://localhost:3031/renderFixture/:fixtureId/*.json
```

Note:
- The `*` captures the path that will be used to fetch the page data from production - so `/renderFixture/:fixutreId/uk.json` will get the data from https://www.theguardian.com/uk.json?dcr.
- The `fixtureId` is used to lookup an object with that key in [`fixtures.js`](https://github.com/guardian/commercial/blob/39fd2ab93ff3feae54ee2e4c1db745c38f9e33d3/scripts/fixtures/fixtures.js). This is then deeply merged into the page data from production, producing JSON that can be used to render content by DCR.

This allows us to render a page for testing that matches production data, but with a small amount of changes required to produce a desirable condition for our test (such as a switch being switched on, or a server-side participation for example). 

So, to render the UK front, overriding properties from fixture with id `foo`, we'd use the URL:

```
http://localhost:3030/Front/http://localhost:3031/renderFixture/foo/uk.json
```

This URL can be used from within a Cypress test using the `getTestUrl` utility, where you have to supply the fixture id that corresponds to a fixture in `fixtures.js`.

## Caveats

- I had to edit our CI to no longer use a Github [container service](https://docs.github.com/en/actions/using-containerized-services/about-service-containers) and use a direct `docker run ...` command. This was necessary in order to pass in the `--network host` argument ([more on this here](https://docs.docker.com/network/drivers/host/)), which allows the container to share the networking stack of the host runner. This is important to allow DCR (localhost:3030) to reach the dev server outside of the container (localhost:3031).

- Fixtures are defined in `fixtures.js` so they're available to the webpack dev server. This is quite far from the actual spec files that send them - it'd be nice if these were more tightly coupled, and perhaps TS was aware of the set of permitted fixture ids.

On the second point, I think we can try and improve the dev-ex here once we've decided to migrate our tests to a different framework.

### DIFF
--- a/.github/workflows/e2e-cypress.yml
+++ b/.github/workflows/e2e-cypress.yml
@@ -34,17 +34,17 @@ jobs:
         run: yarn install --frozen-lockfile
         working-directory: ./commercial
 
-        # We always run our commercial code against the latest main of DCR
-        # This does make our tests sensitive to changes in DCR
-        # (e.g. imagine someone removes the top-above-nav slot from DCR)
-        # This is something we accept in order to easily test our own code
-        #
-        # Note we use the containerised version of DCR, published from:
-        # https://github.com/guardian/dotcom-rendering/blob/6a6df272/.github/workflows/container.yml
-        #
-        # The argument `--network host` is crucial here, as it means the container shares the networking stack of the host
-        # This makes the commercial dev server available from inside the container
-        # Note that GHA provides a service container feature, but it does not support this argument
+      # We always run our commercial code against the latest main of DCR
+      # This does make our tests sensitive to changes in DCR
+      # (e.g. imagine someone removes the top-above-nav slot from DCR)
+      # This is something we accept in order to easily test our own code
+      #
+      # Note we use the containerised version of DCR, published from:
+      # https://github.com/guardian/dotcom-rendering/blob/6a6df272/.github/workflows/container.yml
+      #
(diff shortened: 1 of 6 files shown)


Jakeii (approved):
This is very very cool, would be great to get a dev time on this!

emma-imber (approved):
This is going to be so useful 🎉

arelra (commented):
Great work Chris! 🎉

PR #950, opened 2023-07-18
Remove fronts banner test. Remove code to create banner ads on frontend

domlander:
## What does this change?

Removes the Fronts Banner AB test and associated code to implement `front-banner` ads on frontend.

## What is the value of this and can you measure success?

In Q1, we ran an AB test that involved displaying banner ads instead of MPUs on the UK Network Front. This test has now been completed. We plan to implement fronts banner ads on DCR, but we will doing this in a different way to what we did here, so the code in `fronts-banner-adverts` is no longer required.

Note that code introduced [here](https://github.com/guardian/commercial/pull/878/files), [here](https://github.com/guardian/commercial/pull/885/files), [here](https://github.com/guardian/commercial/pull/900) and much of [here](https://github.com/guardian/commercial/pull/877/files) has been retained, as it pertains to the functionality of the `fronts-banner` slot rather than the implementation on frontend and we intend to implement fronts banner ad slots on DCR.


chrislomaxjones (approved):
Nice 👍

PR #952, opened 2023-07-19
Fix mobile sticky close button

emma-imber:
## What does this change?
The close button was being pushed off to the side of the mobile sticky ad, meaning that it was very difficult to see. The styles have been updated so that the button is now back where it should be.

Also renames the button element variable in the code that creates the close button, to be clearer about the type of element it creates.

Before:
Screenshot 2023-07-19 at 15 41 04

After:
Screenshot 2023-07-19 at 15 40 10

### DIFF
--- a/.changeset/cold-plants-pay.md
+++ b/.changeset/cold-plants-pay.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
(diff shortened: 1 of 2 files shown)


chrislomaxjones (approved):
✅ 

Only 3 stars?!

PR #954, opened 2023-07-20
Billboards in merchandising high slots

chrislomaxjones:
## What does this change?

This PR restores a previous AB test for running billboards in merchandising high slots (see https://github.com/guardian/frontend/pull/25750).

The main changes collected here are:
- Update the AB test definition to match Frontend / DCR.
- Constrain billboards to only running in `merchandising-high` and not `merchandising`.
- on Frontend, wrap the high merch in a container that centres the billboard in the slot. This is analogous to code that already exists on DCR. To achieve this, I moved the `wrapAdSlotInContainer` into the core section of our source.

Corresponding PRs:
- https://github.com/guardian/frontend/pull/26333
- https://github.com/guardian/dotcom-rendering/pull/8305

## Screenshots

| DCR Front      | Frontend Gallery      | DCR Article |
| ----------- | ---------- | ---------- |
| ![one][] | ![two][] | ![three][] |

[one]: https://github.com/guardian/commercial/assets/8000415/22d39803-1a83-4704-ac7e-236726f5e45c
[two]: https://github.com/guardian/commercial/assets/8000415/bd069a92-059c-4a92-8394-2e5c8700f3e3
[three]: https://github.com/guardian/commercial/assets/8000415/8f29825f-5114-461a-942b-e483d2cc0e60

## Why?

We're testing the impact of running programmatic advertising in the `merchandising-high` slot, a slot that has previously been reserved for our own marketing / promotions. This PR includes a 0% AB test, with later PRs making design changes and finally increasing the audience that'll participate in the test.

## Next steps

- [ ] Confirm and implement any changes we'd like to make to the design of these slots on the page.
- [ ] Investigate whether we can easily reuse the fronts banner placements to enable Prebid for merchandising high.


src/core/create-ad-slot.ts
```diff
@@ -33,10 +33,8 @@ const adSlotConfigs: AdSlotConfigs = {
 		label: false,
 		refresh: false,
 	},
-	'high-merch': {
-		label: false,
+	'merchandising-high': {
 		refresh: false,
-		name: 'merchandising-high',
 	},
```
chrislomaxjones, line 38: This diff is a little hard to read but contains two changes:

- We remove `label: false` from the definition. This is uneccessary, since when a merch lands in a slot it doesn't get a label as it's fluid.
- Rename `high-merch` to `merchandising-high` for consistency. This allows us to omit the `name: 'merchandising-high` attribute as well.

src/lib/spacefinder/article-body-adverts.ts
```diff
@@ -62,18 +60,6 @@ let insertedDynamicAds: Advert[] = [];
 const getStickyContainerClassname = (i: number) =>
 	`${adSlotContainerClass}-${i + 2}`;
 
-const wrapSlotInContainer = (
```
Jakeii, line 65: not blocking, but I think we could do this in `liveblog-adverts.ts` too?

Jakeii (approved):
Nice little tidy up with the containers!

src/lib/spacefinder/article-body-adverts.ts
chrislomaxjones: great shout, will sort that out separately.

PR #955, opened 2023-07-20
Remove unused `createAdSlot` config

chrislomaxjones:
## What does this change?

Remove unused slots from ad slot config object:
- `high-merch-lucky`: There are no places where `createAdSlot` is called with this slot name.
- `high-merch-paid`: This is the exact same object as `high-merch`, so we can just use this definition.
- `epic`: There are no places where `createAdSlot` is called with this slot name.
- `mostpop`: These slots are currently not created dynamically.

## Why?

`createAdSlot` is used to create an ad slot div for dynamically inserted ads (e.g. by Spacefinder). This function relies on some configuration, which I've cleaned up above to remove definitions we no longer use.

### DIFF
--- a/.changeset/wicked-plums-matter.md
+++ b/.changeset/wicked-plums-matter.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': patch
+---
+
+Remove unused `createAdSlot` config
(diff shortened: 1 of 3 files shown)


arelra (approved):
lgtm!

PR #957, opened 2023-07-24
Load user features in DCR for reject all readers

emma-imber:
## What does this change?
Loads user features for reject all readers on pages rendered by DCR.

## Why?
We need user features to set cookies for subscribers to make sure that they don't see adverts or paid for messaging. In the future, we should aim to move this functionality out of the commercial bundle as it affects multiple teams.

### DIFF
--- a/src/commercial.consentless.ts
+++ b/src/commercial.consentless.ts
@@ -11,7 +11,10 @@ import { initConsentless } from './lib/consentless/prepare-ootag';
 import { init as setAdTestCookie } from './lib/set-adtest-cookie';
 import { init as setAdTestInLabelsCookie } from './lib/set-adtest-in-labels-cookie';
 
-const bootConsentless = async (consentState: ConsentState): Promise<void> => {
+const bootConsentless = async (
+	consentState: ConsentState,
+	isDotcomRendering: boolean,
+): Promise<void> => {
 	/*  In the consented ad stack, we set the ad free cookie for users who
 		don't consent to targeted ads in order to hide empty ads slots.
 		We remove the cookie here so that we can show Opt Out ads.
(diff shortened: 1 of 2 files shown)


src/commercial.consentless.ts
```diff
@@ -20,15 +23,26 @@ const bootConsentless = async (consentState: ConsentState): Promise<void> => {
...
-	]);
+	];
+
+	if (isDotcomRendering) {
+		const userFeatures = await import(
+			/* webpackChunkName: "dcr" */
+			'lib/user-features'
+		);
+
+		consentlessModuleList.push(userFeatures.refresh());
```
arelra, line 42: Do we want to push the function here rather than invoking?

```suggestion
		consentlessModuleList.push(userFeatures.refresh);
```

arelra: Ah okay sorry yes its okay as you have it. We have two different methods of booting in standalone and consentless. In `commercial.standalone` we create an array of functions which we later invoke but in `commercial.consentless` we immediately execute a bunch of promises that we `promise.all` on.

arelra (approved):
Nice work. Thanks for extracting this!

chrislomaxjones (approved):
Great to see this fixed

src/commercial.consentless.ts
```diff
@@ -20,15 +23,26 @@ const bootConsentless = async (consentState: ConsentState): Promise<void> => {
...
 		initConsentless(consentState),
 		initExclusionSlot(),
 		initFixedSlots(),
 		initArticleInline(),
 		initLiveblogInline(),
-	]);
+	];
+
+	if (isDotcomRendering) {
+		const userFeatures = await import(
```
chrislomaxjones, line 37: This doesn't require any action, just noting down something I noticed:

It looks like Webpack doesn't create a new chunk for lib/user-features for this dynamic import, which is what I would've expected. It seems instead it just includes it in the main standalone/consentless bundles, I guess because we're already imported it statically elsewhere?

Screenshot 2023-07-24 at 13 17 47

johnduffell (approved):
nice, thank you for doing this work 👏

PR #959, opened 2023-07-24
Enable Prebid for the merchandising-high AB test

chrislomaxjones:


src/lib/dfp/merchandising-high-test.ts
```diff
@@ -0,0 +1,6 @@
+import { isInVariantSynchronous } from 'lib/experiments/ab';
+import { billboardsInMerchHigh } from 'lib/experiments/tests/billboards-in-merch-high';
+
+export const includeBillboardsInMerchHigh = () => {
```
chrislomaxjones, line 4: **Why wrap the variant check in this helper?**

We'll probably want to change this check to show billboards to all non-participations whilst the test is running, something like:

```typescript
!isInTest(billboardsInMerchHigh) || isInVariantSynchronous(billboardsInMerchHigh, 'variant');
```

and this will save us from having to do it in a number of different places.

src/lib/header-bidding/prebid/appnexus.ts
```diff
@@ -46,7 +47,7 @@ const getAppNexusDirectPlacementId = (
 	const defaultPlacementId = '9251752';
 	switch (getBreakpointKey()) {
 		case 'D':
-			if (isInFrontsBannerVariant) {
+			if (isInFrontsBannerVariant || includeBillboardsInMerchHigh()) {
```
domlander, line 50: Could you update the comment to indicate that the merchandising-high slot doesn't include leaderboard sized ads either

domlander (approved):

src/lib/header-bidding/prebid/appnexus.ts
chrislomaxjones: Great spot, will do!

PR #961, opened 2023-07-25
Public good slot and messenger message to refresh

Jakeii:
## What does this change?
We've added a new messenger message `passback-refresh` that is similar to the `passback` message, but simply sets the `passback` targeting and refreshes.

Added a new slot called `article-end` that will be shown at the end of article when the Epic doesn't.

Added an AB test to test the UX of the Public good slot and ads.

## Why?
We have a new vendor called "Public Good" which serves via a custom creative in GAM so that we can keep their third-party scripts in a safeframe.

The vendor will show in a new slot when the Epic doesn't appear.

However if they are not serving a campaign, or the user has selected the "Do Not Sell" CCPA consent option, they are not able so serve an ad so we refresh the slot to serve another ad or  collapse the slot.


src/lib/messenger/passback-refresh.ts
```diff
@@ -0,0 +1,30 @@
...
+const passbackRefresh = (specs: string, adSlot: HTMLElement) => {
+	const advert = getAdvertById(adSlot.id);
+	if (advert) {
+		advert.slot.setTargeting('passback', specs);
+		refreshAdvert(advert);
+	}
+};
+
+const init = (register: RegisterListener): void => {
+	register('passback-refresh', (specs, ret, iframe) => {
```
emma-imber, line 15: What is ret being used for? Could we add a comment to explain why it's declared but not used?

src/lib/messenger/passback-refresh.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+              </div>`;
+
+		expect.hasAssertions();
+	});
+
+	afterEach(() => {
+		document.body.innerHTML = '';
+	});
+	describe('passbackRefresh', () => {
+		it("should successfully define an advert and call refreshAdvert and set's targeting", () => {
```
emma-imber, line 35: Very picky but should be sets instead of set's here!

src/lib/messenger/passback-refresh.spec.ts
emma-imber, line 1: Really nice to see testing!

src/lib/messenger/passback-refresh.ts
Jakeii: It's not used but we needed access to the third argument, will change to `_`

domlander (approved):
Looks good 👍

PR #962, opened 2023-07-25
Remove unused "refresh on resize" functionality

chrislomaxjones:
## What does this change?

Remove the currently defunct logic that controls queuing up adverts for refreshing when the breakpoint of the page changes.

Since the Commercial bundle no longer supports Mediator (and DCR never will), this is currently dead code. Removing it will result in some simplification.

**In the process, we can also fix a bug where the data-refresh="false" attribute on slots was being ignored.** This is my main reason for performing this refactor.

## Why?

This code had me convinced that adding `data-refresh="false"` to an ad slot would stop it refreshing. That however is no longer the case. This PR rectifies that, removing dead code in the process.

FWIW, I think this behaviour should eventually be restored, in a way that does not use Mediator and is compatible with DCR.


src/lib/dfp/refresh-on-resize.ts
```diff
@@ -1,66 +0,0 @@
...
-	advertsToRefresh.forEach(refreshAdvert);
-};
-
-const windowResize = () => {
-	// refresh on resize
-	hasBreakpointChanged(refresh);
-};
-
-export const refreshOnResize = (): void => {
-	mediator.on('window:throttledResize', windowResize);
```
chrislomaxjones, line 65: This event is never triggered, and therefore the code that empties the `advertsToRefresh` queue is never called.

src/lib/dfp/should-refresh.ts
```diff
@@ -24,6 +24,11 @@ const shouldRefresh = (
 ): boolean => {
 	const sizeString = advert.size?.toString();
 
+	// Do not refresh ads in slots labelled data-refresh="false"
+	if (advert.node.dataset.refresh === 'false') {
+		return false;
+	}
```
chrislomaxjones, line 30: I've moved this condition that handles `data-refresh="true"` in here to ensure all of the code that decides whether a given advert should refreshed is located in a single place.

Jakeii (approved):
awesome, nice tidy up, I guess if this is revived can just use the data attribute?

PR #964, opened 2023-07-26
Ensure liveblog-right ads are inserted via custom event

domlander:
## What does this change?

- This fixes a bug in the liveblog right column adverts AB test. The liveblog-right ads were being created by `fill-advert-slots.ts`. However, we cannot be certain that all the `liveblog-right` ad slots will be on the page when this code is run. Therefore there's an extra call to `fillAdvertSlots()` before the eventListener is registered, in case the slots are already on the page.
- Moved the logic to restrict the height of the `right` ad slot from the DCR to Commercial repo.

## Why?

- We want the `liveblog-right` adverts to be created when we know the slots are ready.
- So that we can update the styles, since the `right` ad slot is rendered on the server

### DIFF
--- a/src/lib/dfp/fill-advert-slots.ts
+++ b/src/lib/dfp/fill-advert-slots.ts
@@ -85,7 +85,7 @@ const fillAdvertSlots = async (): Promise<void> => {
 			(adSlot) => !(isDCRMobile && adSlot.id === 'dfp-ad--top-above-nav'),
 		)
 		// We cannot guarantee that liveblog-right slots will be on the page when this code runs.
-		// liveblog-right slots are inserted by liveblog-right-adverts.ts, where they will be
+		// liveblog-right slots are inserted by liveblog-right-column-adverts.ts, where they will be
 		// inserted when a custom event is received.
 		.filter((adSlot) => !adSlot.id.includes('liveblog-right'))
 		.map((adSlot) => {
--- a/src/lib/dfp/fill-advert-slots.ts
+++ b/src/lib/dfp/fill-advert-slots.ts
@@ -84,6 +84,10 @@ const fillAdvertSlots = async (): Promise<void> => {
 		.filter(
 			(adSlot) => !(isDCRMobile && adSlot.id === 'dfp-ad--top-above-nav'),
 		)
+		// We cannot guarantee that liveblog-right slots will be on the page when this code runs.
+		// liveblog-right slots are inserted by liveblog-right-adverts.ts, where they will be
+		// inserted when a custom event is received.
+		.filter((adSlot) => !adSlot.id.includes('liveblog-right'))
 		.map((adSlot) => {
 			const additionalSizes = decideAdditionalSizes(adSlot);
 			return createAdvert(adSlot, additionalSizes);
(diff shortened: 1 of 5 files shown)


src/lib/dfp/fill-advert-slots.ts
```diff
@@ -84,6 +84,10 @@ const fillAdvertSlots = async (): Promise<void> => {
 		.filter(
 			(adSlot) => !(isDCRMobile && adSlot.id === 'dfp-ad--top-above-nav'),
 		)
+		// We cannot guarantee that liveblog-right slots will be on the page when this code runs.
+		// liveblog-right slots are inserted by liveblog-right-column-adverts.ts, where they will be
+		// inserted when a custom event is received.
+		.filter((adSlot) => !adSlot.id.includes('liveblog-right'))
```
chrislomaxjones, line 90: Happy to do this for this AB test, but I'd be keen to avoid any extra filtering by slot here in the future.

We discussed this briefly already, but it'd be really nice if `fill-advert-slots.ts` was as greedy as possible with the slots on the page, before mounting an event listener, allowing any future slots that are inserted on the page to be picked up. I'm not sure how that'd play with Spacefinder-inserted sol

src/lib/spacefinder/liveblog-right-column-adverts.ts
```diff
@@ -55,20 +53,54 @@ const fillAdvertSlots = async (numAdsToInsert: number, fromIndex: number) => {
...
 	});
+};
+
+const restrictRightAdContainerHeight = () => {
+	void fastdom.measure(() => {
+		const rightAdContainer =
+			document.querySelector<HTMLElement>('#top-right-ad-slot');
+		if (rightAdContainer !== null) {
+			rightAdContainer.style.maxHeight = '1059px';
+		}
```
chrislomaxjones, line 81: Does this part need to be in a `fastdom.mutate` call rather than measure?

src/lib/spacefinder/liveblog-right-column-adverts.ts
```diff
@@ -55,20 +53,54 @@ const fillAdvertSlots = async (numAdsToInsert: number, fromIndex: number) => {
...
 
-// When a liveblog-right ad slot is inserted into the page, create an ad in this new slot
-export const initLiveblogRightAdverts = (): Promise<void> => {
-	if (isInVariantSynchronous(liveblogRightColumnAds, 'multiple-adverts')) {
-		return Promise.resolve();
-	}
+const createLiveblogRightAdverts = () => {
+	void fillAdvertSlots();
 
 	document.addEventListener('liveblog-right-ads-inserted', (event: Event) => {
```
chrislomaxjones, line 59: I haven't thought too hard about whether this would be simpler, but could we avoid tracking the number of adverts inserted on the Commercial side by having DCR include the id of each slot it wants filled when it fires the custom event?

Feel free to disregard this though, I don't want to make unecessary work!

src/lib/spacefinder/liveblog-right-column-adverts.ts
domlander: Yes, good spot

src/lib/spacefinder/liveblog-right-column-adverts.ts
domlander: We could do, but I don't know if it would be simpler so will leave for now

Jakeii (approved):

PR #967, opened 2023-07-28
Record DCR performance measures

mxdvl:
## What does this change?

Start recording DCR performance measures.

## Why?

So we can see what is fast and what is slow for our users.

### DIFF
--- a/src/core/send-commercial-metrics.ts
+++ b/src/core/send-commercial-metrics.ts
@@ -1,6 +1,7 @@
 import { onConsent } from '@guardian/consent-management-platform';
 import type { ConsentState } from '@guardian/consent-management-platform/dist/types';
-import { log } from '@guardian/libs';
+import type { TeamName } from '@guardian/libs';
+import { getMeasures, isNonNullable, log } from '@guardian/libs';
 import { EventTimer } from './event-timer';
 import type { ConnectionType } from './types';
 
@@ -143,6 +144,17 @@ const getOfflineCount = (): Metric[] =>
 		  ]
 		: [];
 
+/**
+ * Measures added with @guardian/libs’s `startPerformanceMeasure`
+ *
+ * Allows for more granular monitoring of web page performance.
+ */
+const getPerformanceMeasures = (...teams: TeamName[]): Metric[] =>
+	getMeasures(teams).map(({ detail: { team, name, action }, duration }) => ({
+		name: [team, name, action].filter(isNonNullable).join('_'),
+		value: duration,
+	}));
+
 function gatherMetricsOnPageUnload(): void {
 	// Assemble commercial properties and metrics
 	const eventTimer = EventTimer.get();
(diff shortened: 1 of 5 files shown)


src/core/send-commercial-metrics.ts
```diff
@@ -143,6 +143,21 @@ const getOfflineCount = (): Metric[] =>
...
+/**
+ * Measures added with @guardian/libs’s startPerformanceMeasure API
+ * that are assigned to the 'dotcom' team.
+ *
+ * Allows for more granular monitoring of web page performance.
+ */
+const getDotcomPerformanceMeasures = (): Metric[] =>
+	window.performance
+		.getEntriesByType('measure')
+		.filter(({ name }) => name.startsWith('doctom:'))
```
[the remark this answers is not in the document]
mxdvl: Hello I work for Doctor Tom.

src/core/send-commercial-metrics.ts
```diff
@@ -143,6 +143,21 @@ const getOfflineCount = (): Metric[] =>
...
+		.map(({ name, duration }) => ({
+			name: name.replace(':', '_'),
```
arelra, line 157: We should probably export this separator from libs

src/core/send-commercial-metrics.ts
arelra, line 155: Before we start sending data to the lake a PR to consider:

https://github.com/guardian/csnx/pull/756

src/core/send-commercial-metrics.ts
mxdvl: I thought about this some more, and I’m wondering if we should use the “detail” key of performance measures for this, actually.

As mentioned in https://github.com/guardian/csnx/pull/756#pullrequestreview-1554250708, it would probably be best to have a deserialiser live in `@guardian/libs`… in the future

src/core/send-commercial-metrics.ts
mxdvl: Thinking about https://github.com/guardian/csnx/pull/760, actually!

arelra (commented):
Also worth considering:

https://w3c.github.io/beacon/#return-value

> The user agent imposes limits on the amount of data that can be sent via this API: this helps ensure that such requests are delivered successfully and with minimal impact on other user and browser activity. If the amount of data to be queued exceeds the user agent limit as defined in HTTP-network-or-cache fetch, this method returns false

A 64kB limit seems to be a figure that most browsers adopt:

https://chromium.googlesource.com/chromium/src/+/c1a211aa583ade023aca652e9493c01603623aa3/third_party/WebKit/LayoutTests/http/tests/sendbeacon/beacon-allowance-limit.html

I'm sure we're nowhere near this limit but with the increase in metrics that we will now send from DCR it would be good to have a rough idea of the future payload size.

mxdvl:
> it would be good to have a rough idea of the future payload size.

About 3.6kB for the UK front, and 2.8kB for a news article.

I used the script below to get these values:

```js
JSON.stringify(
  window.performance.getEntriesByType("measure")
    .filter(({name}) => name.startsWith("dotcom:"))
    .map(({name, duration}) => ({name, duration}))
).length
```

emma-imber (approved):

PR #970, opened 2023-07-31
External ads

arelra:
## What does this change?

Enable ads to be placed on the page via an external route e.g. thrashers

I'm not completely convinced about the naming but going with 'external' for now

Mostly a generic implementation of #947, specifically [liveblog-right-adverts.ts](https://github.com/guardian/commercial/blob/53e203c77b8a7c9f36300a4dfba81c5b404baa43/src/lib/spacefinder/liveblog-right-adverts.ts) - thanks @domlander !

The extra logic in addition to dynamic slots covers when:
- the commercial runtime loads after the slot so we query any external slots and fill them
- the commercial runtime loads before the slot so we listen for custom events and fill them e.g.
    ```js
    document.dispatchEvent(
        new CustomEvent('gu.commercial.slot.fill', {
            detail: { slotId: 'dfp-ad--external1' },
        }),
    );
    ```

### DIFF
--- a/src/lib/dfp/fill-external-slots.ts
+++ b/src/lib/dfp/fill-external-slots.ts
@@ -16,7 +16,7 @@
 
 import { fillDynamicAdSlot } from './fill-dynamic-advert-slot';
 
-const EXTERNAL_SLOT_PREFIX = 'dfp-ad--external-';
+const EXTERNAL_SLOT_PREFIX = 'dfp-ad--external';
 
 type ExternalSlotCustomEventDetail = {
 	slotId: string;
--- a/src/lib/dfp/fill-external-slots.ts
+++ b/src/lib/dfp/fill-external-slots.ts
@@ -1,0 +1,60 @@
+/**
+ * External slots are a special type of dynamic slot.
+ *
+ * They are not fixed (aka SSR) or dynamic (aka injected by spacefinder).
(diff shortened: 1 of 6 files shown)


src/lib/dfp/fill-external-slots.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const fillExternalSlots = () => {
+	const externalSlots = document.querySelectorAll<HTMLElement>(
+		`[id^=${EXTERNAL_SLOT_PREFIX}]`,
+	);
+	return [...externalSlots].map((slot) => fillDynamicAdSlot(slot, false));
+};
+
+const createSlotFillListener = () => {
+	document.addEventListener('gu.commercial.slot.fill', (event: Event) => {
```
SiAdcock, line 41: Do we want this to be listening specifically for external slot fill events? (i.e. should we give it a more specific name?)

```suggestion
	document.addEventListener('gu.commercial.external.slot.fill', (event: Event) => {
```

arelra: I thought it might be useful to make this generic so that if we extend this in future any kind of slot could be picked up and fulfilled. Does that seem okay?

emma-imber (approved):

PR #974, opened 2023-08-02
Fix consent subscriber expire test

arelra:
## What does this change?

Fixes a flakey test.

We have two issues:

- repeated calls to the `fakeLogin` intercept were not isolated from each other. This is fixed by using `times: 1`
- cookies are set async in Cypress so we need to wait for them to be set and also for the commercial runtime to delete the ad-free cookie on page load

I've stuck waits in for now as a rough placeholder. We should wait for cookies to be set in a more idiomatic fashion.


emma-imber (approved):
This is great, thanks so much for your help!

PR #987, opened 2023-08-04
Don't apply label height to min-height when there is no label

arelra:
## What does this change?

Don't apply min height with label height when `data-label=false`

h/t to @domlander and @emma-imber for their help on this.

## Why?

This causes ads to shift when we add min-height even though `data-label=false`

Before:

https://github.com/guardian/commercial/assets/7014230/5f00eafe-954b-4252-9daf-481033d15f23

After:

https://github.com/guardian/commercial/assets/7014230/933b2023-54b5-46e3-a912-19b2cfac3392

### DIFF
--- a/src/lib/dfp/on-slot-viewable.ts
+++ b/src/lib/dfp/on-slot-viewable.ts
@@ -38,10 +38,16 @@ const setAdSlotMinHeight = (advert: Advert): void => {
 
 	const isStandardAdSize = !size.isProxy();
 	if (isStandardAdSize) {
-		const adSlotHeight = size.height + AD_LABEL_HEIGHT;
-		void fastdom.mutate(() => {
-			node.style.minHeight = `${adSlotHeight}px`;
-		});
+		void fastdom
+			.measure(() => node.getAttribute('data-label') === 'true')
+			.then((hasLabel) => {
+				if (hasLabel) {
+					void fastdom.mutate(() => {
+						const adSlotHeight = size.height + AD_LABEL_HEIGHT;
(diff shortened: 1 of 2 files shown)


Jakeii (approved):
Typo in the changeset message FYI

emma-imber (approved):
LGTM!

arelra:
> If `data-label` is false, then the `min-height` will no longer be set. I assume that's OK?

Good point. I believe the min-height is to prevent CLS on refresh. If they don't have a label then we probably still want min-height to keep the slot height while it refreshes. I think for SSR's slots we have min-height set but I'll update in case there are loopholes.

src/lib/dfp/on-slot-viewable.ts
```diff
@@ -38,10 +38,17 @@ const setAdSlotMinHeight = (advert: Advert): void => {
...
 	if (isStandardAdSize) {
-		const adSlotHeight = size.height + AD_LABEL_HEIGHT;
-		void fastdom.mutate(() => {
-			node.style.minHeight = `${adSlotHeight}px`;
-		});
+		void fastdom
+			.measure(() => node.getAttribute('data-label') === 'true')
+			.then((hasLabel) => {
+				const labelHeight = hasLabel ? AD_LABEL_HEIGHT : 0;
+				if (hasLabel) {
```
domlander, line 45: Can this `if` condition be removed now?

arelra: Just missed this - addressed here:
https://github.com/guardian/commercial/pull/989

PR #990, opened 2023-08-04
Update cypress tests following reduction of text line height

domlander:
## What does this change?

- Fixes a failing cypress test. 
- Replaces the liveblog test page with another that contains a more typical number of embeds. The previous Liveblog test page had a lot more Twitter embeds than usual.
- Split expected slot tests into mobile and desktop.

## Why?

- The reduction in line height for text content on liveblog meant that the page became shorter and commercial have less space to insert ads. On the test liveblog page, we were now only inserting 3 ads instead of 4 at some breakpoints.
- There is no need for our test liveblog page to have so many twitter embeds. It was a red herring when investigating the cause of this Cypress test failure.
- Desktop and mobile screen sizes typically render a different number of ad slots

### DIFF
--- a/.changeset/funny-dragons-deliver.md
+++ b/.changeset/funny-dragons-deliver.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': patch
+---
+
+Fix cypress tests following text line height reduction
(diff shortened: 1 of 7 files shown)


Jakeii:
Nice, finally have that last green tick back!😍 

I've been thinking, do we actually want to keep track of the count of inlines in cypress? Perhaps a canary would be better suited?

arelra (approved):
🎉

arelra:
> Nice, finally have that last green tick back!😍
> 
> I've been thinking, do we actually want to keep track of the count of inlines in cypress? Perhaps a canary would be better suited?

I think the idea was we'd catch any regressions in our placement / spacefinder behaviour early

PR #992, opened 2023-08-07
Add cypress test for pageskins

Jakeii:
## What does this change?
Add a test for pageskins, testing if the pageskin image is applied on desktop and that Single Request Mode is used, also check the opposite on tablet/mobile.

Refactored `getTestUrl` slightly, to remove DCR option, we're always on DCR now and in the future.

## Why?
We wan't to be confident that the logic is working correctly, especially single request mode, it's not easy to notice if it's not behaving as we'd like.

Bonus: [Puppies Test Pageskin](https://www.theguardian.com/uk?adtest=puppies-pageskin)

### DIFF
--- a/cypress/e2e/parallel-4/pageskin.cy.ts
+++ b/cypress/e2e/parallel-4/pageskin.cy.ts
@@ -24,10 +24,13 @@ describe('pageskin on uk front', () => {
 			cy.intercept(gamUrl, (req) => {
 				req.continue((res) => {
 					expect(res.headers).to.have.property('google-lineitem-id');
-					expect(
-						(res.headers['google-lineitem-id'] as string).split(',')
-							.length,
-					).to.be.greaterThan(1);
+
+					// In Single Request Mode, the google-lineitem-id header will contain multiple line item ids
+					const lineItemIds = (
+						res.headers['google-lineitem-id'] as string
+					).split(',');
+
+					expect(lineItemIds).to.be.greaterThan(1);
 				});
 			}).as('gamCall');
 
@@ -52,10 +55,11 @@ describe('pageskin on uk front', () => {
 			cy.intercept(gamUrl, (req) => {
 				req.continue((res) => {
 					expect(res.headers).to.have.property('google-lineitem-id');
-					expect(
-						(res.headers['google-lineitem-id'] as string).split(',')
-							.length,
-					).to.equal(1);
+					// In Non Single Request Mode, the google-lineitem-id header will contain a single line item id
+					const lineItemIds = (
+						res.headers['google-lineitem-id'] as string
+					).split(',');
+					expect(lineItemIds).to.have.lengthOf(1);
 				});
 			}).as('gamCall');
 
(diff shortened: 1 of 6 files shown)


emma-imber (approved):

cypress/fixtures/pages/liveblogs.ts
```diff
@@ -7,8 +7,8 @@ const liveblogs: Page[] = [
 	{
 		path: getTestUrl(
 			stage,
-			'/politics/live/2022/jan/31/uk-politics-live-omicron-nhs-workers-coronavirus-vaccines-no-10-sue-gray-report?live=true',
-			{ isDcr: true },
+			'/politics/live/2022/jan/31/uk-politics-live-omicron-nhs-workers-coronavirus-vaccines-no-10-sue-gray-report',
```
domlander, line 10: Removing `?live=true` causes the other Cypress-3 test to fail. I am fixing the other test in my PR, which will fix this one too, so happy for you to leave it

cypress/e2e/parallel-4/pageskin.cy.ts
```diff
@@ -0,0 +1,75 @@
...
+		it(`Test ${frontWithPageSkin.path} on ${breakpoint.breakpoint} should display the pageskin background and use single request mode`, () => {
+			cy.viewport(breakpoint.width, breakpoint.height);
+
+			cy.intercept(gamUrl, (req) => {
+				req.continue((res) => {
+					expect(res.headers).to.have.property('google-lineitem-id');
+					expect(
+						(res.headers['google-lineitem-id'] as string).split(',')
+							.length,
+					).to.be.greaterThan(1);
```
domlander, line 30: Does this `expect` test that we're using `single request mode`. If so, why?

domlander:
> we're always on DCR now and in the future

Bold to say we'll always be on DCR in the future 😛

domlander:
Nice refactor and great to have pageskin tests

cypress/fixtures/pages/liveblogs.ts
Jakeii: It's still there just added here now: https://github.com/guardian/commercial/pull/992/files#diff-188823ef01a17b9942962471319b2f75645e1bea6bd18c38eafb078ae0daf56aR29

cypress/e2e/parallel-4/pageskin.cy.ts
Jakeii: Tom Duffet showed us the `google-lineitem-id` header during the incident, in single request mode it'll have all the ads on the page, in "normal" mode it'll just have a single lineitem.

Will clear it up in the code

cypress/fixtures/pages/liveblogs.ts
Jakeii: Ah the urls weren't quite correct for liveblogs when running locally, fixed.

domlander (approved):

PR #994, opened 2023-08-08
Migrate commercial to use identity-auth

emma-imber:
## What does this change?
Migrates the commercial repository to use the identity-auth repository to handle identity requests. We have followed the pull requests from frontend to try and align our code with the changes made there. Changes in this PR include:

- Add identity-auth package to the repository
- Add an Okta file containing types and helper functions
- Add AuthStatus types and GetAuthStatus function
- Migrate shouldHideSupportMessaging to Okta
- Migrate requestNewData to Okta
- Migrate isRecurringContributor to Okta
- Migrate isPayingMember to Okta
- Migrate fetchUserIdentifiers to return the googleTagID from Okta
- Use isUserLoggedInOktaRefactor in place of isUserLoggedIn

We still need to make more changes, but these will come in future pull requests as this one is already pretty big!

Future changes include:
- Migrating build-page-targeting to use the new Okta isUserLoggedIn function instead of checking the GU_U cookie
- Migrate Cypress tests to align with Okta migration - they currently set a GU_U cookie to fake logging in, and soon this will not be valid

## Why?
WebEx has been leading a migration to using identity-auth for our identity requests, and the commercial repository needs to align with these.


src/lib/dfp/prepare-googletag.ts
```diff
@@ -73,14 +73,12 @@ const setPageTargeting = (consentState: ConsentState) =>
...
-	getUserIdentifiersFromApi((userIdentifiers) => {
-		if (userIdentifiers?.googleTagId) {
-			window.googletag
-				.pubads()
-				.setPublisherProvidedId(userIdentifiers.googleTagId);
-		}
-	});
+const setPublisherProvidedId = async (): Promise<void> => {
+	const googleTagId = await getGoogleTagId();
+	if (!(googleTagId === null)) {
```
AshCorr, line 78:
```suggestion
	if (googleTagId !== null) {
```

src/lib/user-features.ts
```diff
@@ -318,10 +332,10 @@ const shouldNotBeShownSupportMessaging = (): boolean =>
     which this function is dependent on.
 */
 
-const shouldHideSupportMessaging = (): boolean =>
+const shouldHideSupportMessagingOkta = async (): Promise<boolean> =>
```
AshCorr, line 335: In a lot of the cases in Frontend we created `Okta` variants of methods when we needed access to the old pre-okta method aswell as the new Okta method, but if you're able to just refactor the old method to work for Okta I don't think theres any need to suffix methods with `Okta`

src/lib/user-features.ts
```diff
@@ -142,25 +146,35 @@ const deleteOldData = (): void => {
...
+			)
+				.then((response) => {
+					if (!validateResponse(response)) {
+						throw new Error('invalid response');
+					}
+					return response;
+				})
+				.then(persistResponse)
+				.catch(noop);
+		});
```
AshCorr, line 176:
```suggestion
		.then((signedInAuthStatus) => fetchJson(
		        `${
			        window.guardian.config.page.userAttributesApiUrl ??
			        '/USER_ATTRIBUTE_API_NOT_FOUND'
		        }/me`,
		        {
			        mode: 'cors',
			        ...getOptionsHeadersWithOkta(signedInAuthStatus),
		        },
	        ))
		.then((response) => {
			if (!validateResponse(response)) {
				throw new Error('invalid response');
			}
			return response;
		})
		.then(persistResponse)
		.catch(noop);
```

You could make use of [promise chaining](https://blog.greenroots.info/javascript-promise-chain-the-art-of-handling-promises#heading-promise-chain-rule-2) here if you wanted to reduce verbosity here!

src/lib/identity/okta.ts
```diff
@@ -0,0 +1,79 @@
...
+
+let identityAuth:
+	| IdentityAuth<AccessTokenClaims, CustomIdTokenClaims>
+	| undefined;
+
+function getIdentityAuth() {
+	if (identityAuth === undefined) {
+		const stage: Stage = getStage();
+
+		identityAuth = new IdentityAuth<AccessTokenClaims, CustomIdTokenClaims>(
```
AshCorr, line 48: Curious as to what happens if there are 2 instances of IdentityAuth on a page (1 from Commercial, 1 from Frontend/DCR) 🤔 Will they both be trying to refresh the token? Do we maybe need to expose a single instance of IdentityAuth on the window? Maybe a question for Identity...

src/lib/identity/api.ts
```diff
@@ -74,41 +91,126 @@ const getUserFromCookie = (): IdentityUserFromCache => {
...
+			return {
+				kind: 'SignedOutWithCookies',
+			};
 		}
-	},
-);
+	}
+};
+
+const isUserLoggedInOktaRefactor = (): Promise<boolean> =>
```
AshCorr, line 148: We've renamed this to `isUserLoggedIn` in Frontend - we only had a seperate method whilst we were still migrating so we could run okta and non okta in parallel, but if everything already uses the Okta method you mayaswell remove the non-okta one and rename the Okta one.

src/lib/identity/okta.ts
coldlink: We'd want to share one instance of IdentityAuth between all js on the page! This will cause issues especially around token refreshing, as they'll all try to refresh at the same time.

We only have a limited number of API calls we can make to Okta per minute, so anything that will reduce this will be helpful.

src/lib/user-features.ts
emma-imber: This is a line for line copy of the same function from frontend - so I'd probably prefer to keep the two aligned with one another!

src/lib/dfp/prepare-googletag.ts
```diff
@@ -94,7 +92,7 @@ export const init = (): Promise<void> => {
 			let canRun = true;
 
 			if (consentState.canTarget) {
-				window.googletag.cmd.push(setPublisherProvidedId);
+				window.googletag.cmd.push(() => setPublisherProvidedId);
```
Jakeii, line 95: `setPublisherProvidedId` won't be called here!

Also even if it was the promise won't be waited on before execution here (I don't think), so there could be a chance that the id isn't set before googletag requests are made. 

Maybe something like:
```js
await new Promise((resolve) => {
	window.googletag.cmd.push(() => setPublisherProvidedId().then(resolve));
});
```

Or there may be a better solution

src/lib/identity/okta.ts
Jakeii: How is it shared between DCR and Frontend? can we use the same method to share it?

emma-imber: After discussion with @AshCorr we're going to use the logic from this PR: https://github.com/guardian/frontend/pull/26469

Either commercial or frontend will be able to initialise identityAuth and add it to the window when needed. The repo who looks for the identityAuth code second will then be able to use the object on the window, preventing multiple instances of identityAuth running at the same time.

We'll be looking to abstract this logic out into a separate library so that we don't need to worry about manually keeping frontend and commercial in sync with each other.

src/lib/identity/api.ts
```diff
@@ -74,37 +91,122 @@ const getUserFromCookie = (): IdentityUserFromCache => {
...
+		kind: 'SignedOutWithCookies',
+	};
+};
+
+const isUserLoggedInOktaRefactor = (): Promise<boolean> =>
+	getAuthStatus().then((authStatus) =>
+		authStatus.kind === 'SignedInWithCookies' ||
+		authStatus.kind === 'SignedInWithOkta'
+			? true
+			: false,
```
domlander, line 149:
```suggestion
		authStatus.kind === 'SignedInWithCookies' ||
		authStatus.kind === 'SignedInWithOkta'
```

src/lib/identity/okta.ts
arelra: Would we need this instantiation method on DCR too? aiui the instantiation of auth state on Frontend would only set it for Frontend rendered pages.

AshCorr: We would need it for DCR too

src/lib/identity/api.ts
```diff
@@ -57,8 +75,7 @@ const getUserFromCookie = (): IdentityUserFromCache => {
 		if (userData && cookieData) {
 			const displayName = decodeURIComponent(userData[2]);
 			userFromCookieCache = {
-				id: parseInt(userData[0], 10),
-				primaryEmailAddress: userData[1], // not sure where this is stored now - not in the cookie any more
+				id: parseInt(userData[0], 10), // not sure where this is stored now - not in the cookie any more
```
georgeblahblah, line 78: I think this comment applied to the `primaryEmailAddress` field!

```suggestion
				id: parseInt(userData[0], 10),
```

src/lib/identity/api.ts
```diff
@@ -74,37 +91,122 @@ const getUserFromCookie = (): IdentityUserFromCache => {
 	return userFromCookieCache;
 };
 
-const isUserLoggedIn = (): boolean => getUserFromCookie() !== null;
-
-const fetchUserIdentifiers = () => {
-	const url = `${idApiRoot}/user/me/identifiers`;
-	return fetch(url, {
+/**
+ * Fetch the logged in user's Google Tga ID from IDAPI
```
georgeblahblah, line 95:
```suggestion
 * Fetch the logged in user's Google Tag ID from IDAPI
```

georgeblahblah (approved):
This looks great from an Okta/WebX side, but I can't comment much on the commercial logic!

src/lib/identity/api.ts
emma-imber: Good spot!

AshCorr (approved):

src/types/global.ts
```diff
@@ -441,6 +454,7 @@ declare global {
 			notificationEventHistory?: HeaderNotification[][];
 			commercialTimer?: EventTimer;
 			offlineCount?: number;
+			identityAuth?: IdentityAuth<AccessTokenClaims, CustomIdTokenClaims>;
```
coldlink, line 457:
```suggestion
			identityAuth?: IdentityAuth<never, CustomIdTokenClaims>;
```

You only need to pass on any additional custom claims to extend the type, the default access token claims are always provided

Jakeii (approved):
Awesome to get this over the line!

arelra (commented):
Great work again @emma-imber ! 🛂

PR #995, opened 2023-08-08
Split US ozone placement IDs by slot and device

Jakeii:
## What does this change?
Adds more placement ID's for the US on desktop depending on the size

## Why?
The new placement IDs allow TripleLift to serve through Ozone

### DIFF
--- a/src/lib/header-bidding/prebid/bid-config.ts
+++ b/src/lib/header-bidding/prebid/bid-config.ts
@@ -26,6 +26,7 @@ import type {
 	PrebidXaxisParams,
 } from '../prebid-types';
 import {
+	containsBillboard,
 	containsBillboardNotLeaderboard,
 	containsLeaderboard,
 	containsLeaderboardOrBillboard,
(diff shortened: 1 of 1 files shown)


src/lib/header-bidding/prebid/bid-config.ts
```diff
@@ -223,17 +224,35 @@ const openxClientSideBidder: (pageTargeting: PageTargeting) => PrebidBidder = (
...
+};
 
 const ozoneClientSideBidder: (pageTargeting: PageTargeting) => PrebidBidder = (
 	pageTargeting: PageTargeting,
 ) => ({
 	name: 'ozone',
 	switchName: 'prebidOzone',
-	bidParams: (): PrebidOzoneParams => ({
+	bidParams: (
+		_slotId: string,
```
emma-imber, line 250: Is this slotId actually referenced at any point?

Jakeii: Nope, but it needs to be there to access the second argument, the _ prefix tells typescript/vscode to ignore the unused variable.

emma-imber (approved):
LGTM!

domlander:
Code looks good. I assume that tablet devices should continue using the current ID? Also, do all relevant desktop slots have either a billboard or and MPU size? If not, should they continue using the old ID?

domlander (approved):

Jakeii:
> Code looks good. I assume that tablet devices should continue using the current ID? 
> Also, do all relevant desktop slots have either a billboard or and MPU size? If not, should they continue using the old ID?
Yeah I noticed there would be no placement ID for all other sizes/breakpoints, I asked Alex if perhaps we should use the old ID as a fallback and he agreed.

src/lib/header-bidding/prebid/bid-config.ts
```diff
@@ -223,17 +224,35 @@ const openxClientSideBidder: (pageTargeting: PageTargeting) => PrebidBidder = (
...
+			}
+
+			if (containsMpu(sizes)) {
+				return '3500010911';
+			}
+		}
+		return '1420436308';
+	} else {
+		return '0420420500';
+	}
```
arelra, line 241: Do we need this else? If `0420420500` is a default option we could just return it?

Jakeii: Good point, this rule might be handy! https://eslint.org/docs/latest/rules/no-else-return

PR #998, opened 2023-08-10
Add eslint no-else-return

Jakeii:
## What does this change?
Add [this rule](https://eslint.org/docs/latest/rules/no-else-return)

## Why?
Tidier code


arelra (approved):
Thanks for adding this 📈

domlander (approved):

PR #999, opened 2023-08-10
Update @types/googletag to v3.0.3

Jakeii:
## What does this change?
Update @types/googletag v3.0.3

## Why?
Some of the types have been updated, and `googletag.setNonPersonalizedAds` was deprecated, requiring some code change, so giving this dependency update it's own PR

The deprecation is in the types PR https://github.com/DefinitelyTyped/DefinitelyTyped/pull/61241, but missing from the release notes https://developers.google.com/publisher-tag/release-notes which was a little confusing.

### DIFF
--- a/src/lib/dfp/prepare-googletag.ts
+++ b/src/lib/dfp/prepare-googletag.ts
@@ -110,7 +110,6 @@ export const init = (): Promise<void> => {
 				canRun = getConsentFor('googletag', consentState);
 			} else if (consentState.aus) {
 				// AUS mode
-				// canRun stays true, set NPA flag
 				const nonPersonalizedAds = !getConsentFor(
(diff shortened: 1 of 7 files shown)


src/lib/dfp/prepare-googletag.ts
```diff
@@ -111,11 +111,14 @@ export const init = (): Promise<void> => {
 			} else if (consentState.aus) {
 				// AUS mode
 				// canRun stays true, set NPA flag
```
domlander, line 113: We can remove this comment

Jakeii: We're still setting the flag, just in a different way but the comment does seem a bit redundant!

PR #1003, opened 2023-08-14
YouTube targeting uses `Consentstate.canTarget`

arelra:
## What does this change?

YouTube targeting uses `Consentstate.canTarget`

Since [this PR](https://github.com/guardian/consent-management-platform/pull/597) we now determine `canTarget` as part of the CMP process [here](https://github.com/guardian/consent-management-platform/blob/e768b95da95ded9c6320247cf5ddebb9beeb19a4/src/onConsentChange.ts#L50).

### DIFF
--- a/src/core/targeting/youtube.ts
+++ b/src/core/targeting/youtube.ts
@@ -46,19 +46,20 @@ const buildAdsConfig = (
 		} satisfies AdsConfigCCPAorAus;
 	}
 
-	if (cmpConsent.tcfv2) {
+	if (cmpConsent.framework === 'tcfv2') {
 		const tcfData = cmpConsent.tcfv2;
-		const canTarget = Object.values(tcfData.consents).every(Boolean);
-		const mergedAdTagParameters = {
-			...defaultAdsConfig.adTagParameters,
-			cmpGdpr: tcfData.gdprApplies ? 1 : 0,
-			cmpGvcd: tcfData.addtlConsent,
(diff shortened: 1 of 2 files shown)


src/core/targeting/youtube.ts
```diff
@@ -39,35 +39,27 @@ const buildAdsConfig = (
 		},
 	};
 
-	if (cmpConsent.ccpa) {
-		const canTarget = !cmpConsent.ccpa.doNotSell;
+	if (cmpConsent.framework === 'ccpa' || cmpConsent.framework === 'aus') {
```
chrislomaxjones, line 42: Non blocking as I'm sure this works, but just for my understanding, what's the difference between this here and doing `cmpConsent.ccpa || cmpConsent.aus`?

chrislomaxjones (approved):
Looks good! Nice to see `satisfies` in the wild as well.

src/core/targeting/youtube.ts
arelra: No functional difference. 'framework' was added to consentstate to make checking the framework slightly more readable and particular use cases simpler such as switch case statements and when you just want to pass the active framework along rather than having to check the existence of each property for example [here](https://github.com/guardian/frontend/pull/24498/commits/1daffb798617cb6824dd937b0e05b111de0ac83d).

chrislomaxjones: Makes sense, thank you

PR #1005, opened 2023-08-14
Add kargo as a prebid bidder in the US if in 0% AB test

Jakeii:
## What does this change?
Adds `kargo` as a bidder in the USA, 

Frontend PR where switch and test was added: https://github.com/guardian/frontend/pull/26475

Enabling kargo bid adapter in our prebid.js fork https://github.com/guardian/Prebid.js/pull/134
which has been updated in this PR

The relevant line-items have been generated in this PR https://github.com/guardian/commercial-tools/pull/165 where you can see the price bucketing and line items priority (same as prebid main)

The placement IDs are as follows:
Platform | Type | Sizes | Placement ID
-- | -- | -- | --
Mobile | Middle In-Article | 300x250 | _y9LINEsbfh
Mobile | Adhesion | 320x50 | _odszPLn2hK
Desktop | Middle In-Article | 300x250 | _qDBbBXYtzA
Desktop | Right Rail | 300x250, 300x600, 160x600 | _zOpeEAyfiz
Desktop | Above The Fold | 728x90 | _y6zKwuoPpX
Desktop | Adhesion | 970x250 | _yflg9S7c2x

## Why?
We would like to add `kargo` as a prebid bidder in the USA behind a 0% test

### DIFF
--- a/src/lib/header-bidding/prebid-types.ts
+++ b/src/lib/header-bidding/prebid-types.ts
@@ -111,6 +111,10 @@ export type PrebidSmartParams = {
 	formatId: number;
 };
 
+export type PrebidKargoParams = {
+	placementId: string;
+};
+
 export type BidderCode =
 	| 'adyoulike'
 	| 'and'
@@ -125,7 +129,8 @@ export type BidderCode =
 	| 'sonobi'
 	| 'triplelift'
 	| 'trustx'
-	| 'xhb';
+	| 'xhb'
+	| 'kargo';
 
 export type PrebidParams =
 	| PrebidAdYouLikeParams
@@ -140,7 +145,8 @@ export type PrebidParams =
 	| PrebidSonobiParams
 	| PrebidTripleLiftParams
 	| PrebidTrustXParams
-	| PrebidXaxisParams;
+	| PrebidXaxisParams
+	| PrebidKargoParams;
 
 export type PrebidBidder = {
 	name: BidderCode;
--- a/src/lib/header-bidding/prebid-types.ts
+++ b/src/lib/header-bidding/prebid-types.ts
@@ -122,6 +122,7 @@ export type BidderCode =
 	| 'criteo'
 	| 'improvedigital'
 	| 'ix'
+	| 'kargo'
 	| 'oxd'
 	| 'ozone'
 	| 'pubmatic'
@@ -129,8 +130,7 @@ export type BidderCode =
 	| 'sonobi'
 	| 'triplelift'
 	| 'trustx'
-	| 'xhb'
-	| 'kargo';
+	| 'xhb';
 
 export type PrebidParams =
 	| PrebidAdYouLikeParams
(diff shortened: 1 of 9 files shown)


src/lib/header-bidding/prebid/prebid.ts
```diff
@@ -427,6 +428,12 @@ const initialise = (window: Window, framework: Framework = 'tcfv2'): void => {
 		};
 	}
 
+	if (window.guardian.config.switches.prebidKargo) {
+		window.pbjs.bidderSettings.kargo = {
+			storageAllowed: true,
```
Jakeii, line 433: This is fine as we get consent for prebid

src/lib/header-bidding/prebid-types.ts
```diff
@@ -125,7 +129,8 @@ export type BidderCode =
 	| 'sonobi'
 	| 'triplelift'
 	| 'trustx'
-	| 'xhb';
+	| 'xhb'
+	| 'kargo';
```
chrislomaxjones, line 133: Super nitpicky this one, but could we keep these codes in alphabetical order? Same with the prebid params below as well.

src/lib/header-bidding/utils.ts
```diff
@@ -84,6 +87,9 @@ export const containsLeaderboard = (sizes: HeaderBiddingSize[]): boolean =>
 export const containsBillboard = (sizes: HeaderBiddingSize[]): boolean =>
 	contains(sizes, createAdSize(970, 250));
 
+export const containsSkyscraper = (sizes: HeaderBiddingSize[]): boolean =>
```
chrislomaxjones, line 90: Is this function used anywhere? Does it need to be used in `getKargoPlacementId`?

src/lib/header-bidding/prebid-types.ts
Jakeii: how do we feel about enabling various alphabetical eslint rules?
https://typescript-eslint.io/rules/sort-type-constituents/

For another PR I think though!

src/lib/header-bidding/utils.ts
chrislomaxjones: Nice one, thanks!

src/lib/header-bidding/prebid-types.ts
chrislomaxjones: > how do we feel about enabling various alphabetical eslint rules?

I'd defer to the rest of the team on this one. I tend to prefer alphabetical ordering in these situation when there's no obvious precedence / order of importance, however when there is a more obvious ordering then I prefer keeping it explicit.

chrislomaxjones (approved):
Great stuff 👍

src/lib/header-bidding/prebid-types.ts
domlander: I agree with Chris. I like alphabetical ordering _except_ when another ordering makes more sense. I wouldn't want to have to disable the eslint rule when another order makes more sense.

PR #1010, opened 2023-08-15
Make `@guardian` dependencies peer dependencies

arelra:
## What does this change?

Move all `@guardian` dependencies to peer dependencies so that clients of commercial (primarily DCR) with strict checks on peer dependency alignment do not have to raise versions in lockstep.

Yarn and npm < v7 do not install peer dependencies so we have to add them to dev dependencies for local development and bundling.

The commercial repo publishes two things:
1) the commercial bundle which requires deps at bundle time
2) utility functions for consumption by consuming applications

aiui release bundling still works with @guardian deps as peers deps because we install all dependencies (i.e. not using install --production) and then this get cached and restored by all the workflow jobs.

### DIFF
--- a/.changeset/heavy-berries-sit.md
+++ b/.changeset/heavy-berries-sit.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': major
+---
+
+Move guardian deps to peer deps
--- a/package.json
+++ b/package.json
@@ -52,8 +52,14 @@
 		"@babel/register": "^7.11.5",
 		"@babel/runtime": "^7.11.2",
 		"@cypress/webpack-preprocessor": "^5.17.1",
+		"@guardian/ab-core": "^5.0.0",
+		"@guardian/consent-management-platform": "^13.6.1",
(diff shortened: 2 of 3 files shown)


domlander (approved):
We'll need to remember to update peerDependencies when updating devDependencies

arelra:
> We'll need to remember to update peerDependencies when updating devDependencies

Good point @domlander !

We have recommendations here:
https://github.com/guardian/recommendations/blob/main/npm-packages.md#peerdependencies

Our case is slightly easier than those described above as our clients are internal and within our control i.e. we can always do a major bump and bump our clients versions if we want to change peer deps significantly.

domlander:
> We have recommendations here: https://github.com/guardian/recommendations/blob/main/npm-packages.md#peerdependencies

Those examples in there are really helpful! I understand peer dependencies a bit better now.

> Our case is slightly easier than those described above as our clients are internal and within our control i.e. we can always do a major bump and bump our clients versions if we want to change peer deps significantly.

Gotcha. This makes sense

PR #1014, opened 2023-08-16
Use Guardian browserslist to govern browser support

emma-imber:
## What does this change?
Uses the "@guardian/browserslist-config" package so that we build the bundle to be compatible with the browsers defined here: https://github.com/guardian/csnx/tree/main/libs/%40guardian/browserslist-config#included-browsers

I've also added the deprecated addListener function as a fallback for addEventListener in detect-breakpoint. In versions of Safari before 14, the MediaQueryList does not inherit the addEventListener function, and this was preventing ads from loading as the function could not be found.

This means Safari versions 11, 12, and 13 are now supported and will show ads.

## Why?
Using a standardised list is more robust than just building for the last x versions of a browser, and makes sure that we build our bundle to be compatible with the browsers that we know our readers use. This should mean our ads reach more users.

## How does this affect the bundle size?
Before adding the browserslist, the parsed bundle size according to webpack-bundle-analyzer was 626.74 kB

After adding browserslist, the parsed bundle size is now 653.8 kB

This is approximately a 4% increase in bundle size for the additional browser support.

### Before
Safari 13.1 on Catalina
Screenshot 2023-08-16 at 17 36 39

### After
Safari 11.1 on High Sierra
Screenshot 2023-08-18 at 17 20 03

### DIFF
--- a/src/lib/detect/detect-breakpoint.ts
+++ b/src/lib/detect/detect-breakpoint.ts
@@ -119,11 +119,11 @@ const initMediaQueryListeners = () => {
 			const listener = (mql: MediaQueryListEvent | MediaQueryList) =>
 				mql.matches && updateBreakpoint(bp);
 
-			try {
+			//addListener is a deprecated method but Safari 13 and earlier versions do not support the new
+			//method, so we need it here as a fallback to load ads on those versions of Safari
+			if (typeof mql.addEventListener !== 'undefined') {
 				mql.addEventListener('change', listener);
-			} catch {
-				//addListener is a deprecated method but Safari 13 and earlier versions do not support the new
-				//method, so we need it here as a fallback to load ads on those versions of Safari
+			} else if (typeof mql.addListener !== 'undefined') {
 				mql.addListener(listener);
 			}
 
--- a/src/lib/detect/detect-breakpoint.ts
+++ b/src/lib/detect/detect-breakpoint.ts
@@ -119,7 +119,13 @@ const initMediaQueryListeners = () => {
 			const listener = (mql: MediaQueryListEvent | MediaQueryList) =>
 				mql.matches && updateBreakpoint(bp);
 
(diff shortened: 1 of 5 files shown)


arelra:
I'm _very_ surprised that this hasn't affected the bundle size which makes me suspect that  browserslist hasn't been applied.

emma-imber:
> I'm _very_ surprised this hasn't affected the bundle size which makes me suspect that browserslist hasn't been applied.

Yep it wasn't being applied - turned out I just needed to add preset-env to pull the browserslist in

src/lib/detect/detect-breakpoint.ts
```diff
@@ -119,7 +119,13 @@ const initMediaQueryListeners = () => {
...
 				mql.matches && updateBreakpoint(bp);
 
-			mql.addEventListener('change', listener);
+			try {
+				mql.addEventListener('change', listener);
+			} catch {
+				//addListener is a deprecated method but Safari 13 and earlier versions do not support the new
+				//method, so we need it here as a fallback to load ads on those versions of Safari
+				mql.addListener(listener);
+			}
```
arelra, line 128: Wondering if it is preferable to check for the existence of `addEventListener` rather than catching the exception? I think a catch would catch any exception occurring from the listener whereas a check of whether the method exists or not is slightly more explicit. Something like this (but please verify!): 

```js
if (typeof(mql.addEventListener) !== "undefined") {
  mql.addEventListener('change', listener);
} else if (typeof(mql.addListener) !== "undefined") {
  mql.addListener(listener);
}
```

arelra (approved):
Great work Emma! 📈

src/lib/detect/detect-breakpoint.ts
emma-imber: Great point! I like that way of writing it out, will give it a go

PR #1015, opened 2023-08-17
Upgrade Prebid.js to v7.54.4

arelra:
## What does this change?

Upgrade Prebid.js to v7.54.4 as per https://github.com/guardian/Prebid.js/pull/132

Note:

The prebid config object has changed to not include any private fields (`_*`) when calling:

`window.pbjs.getConfig()`

I think this is okay as all the fields have public equivalents and it does not seem to affect the functioning of prebid.

Have tested on CODE

### DIFF
--- a/src/lib/header-bidding/prebid/prebid.spec.ts
+++ b/src/lib/header-bidding/prebid/prebid.spec.ts
@@ -43,35 +43,6 @@ describe('initialise', () => {
 	test('should generate correct Prebid config when all switches on', () => {
 		prebid.initialise(window, 'tcfv2');
 		expect(window.pbjs?.getConfig()).toEqual({
-			_auctionOptions: {},
-			_bidderSequence: 'random',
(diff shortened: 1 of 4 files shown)


src/lib/header-bidding/prebid/prebid.spec.ts
```diff
@@ -43,35 +43,6 @@ describe('initialise', () => {
 	test('should generate correct Prebid config when all switches on', () => {
 		prebid.initialise(window, 'tcfv2');
 		expect(window.pbjs?.getConfig()).toEqual({
-			_auctionOptions: {},
```
arelra, line 46: Private fields are no longer returned by `getConfig()`

chrislomaxjones (approved):
Looks good 👍

domlander (approved):

arelra:
> Looks good 👍

Thanks for checking the price granularity @chrislomaxjones

PR #1017, opened 2023-08-18
Remove switch check for ias

Jakeii:
## Why?
We're removing this switches as part of our switch audit, so we don't need to check it any more.

Once it was made to be always on the change started breaking unit tests, as it was attempting to run IAS in test. To resolve this I moved the IAS code out of `define-slot.ts` to a separate module so it can be mocked, and then itself tested in isolation (as a bonus more thoroughly!).


src/lib/dfp/init-slot-ias.ts
```diff
@@ -0,0 +1,128 @@
...
+	ms: number,
+): Promise<T | void> => {
+	let timeoutId: number;
+	return Promise.race([
+		promise,
+		new Promise<void>((resolve) => {
+			timeoutId = window.setTimeout(resolve, ms);
+		}),
+	]).then((result) => {
+		window.clearTimeout(timeoutId);
```
Jakeii, line 25: is it actually worth clearing the timeout? I don't think it's necessary?

src/core/third-party-tags/ias.ts
```diff
@@ -5,8 +5,8 @@ import type { GetThirdPartyTag } from '../types';
  * https://integralads.com/uk/
  * @param  {} {shouldRun}
  */
-export const ias: GetThirdPartyTag = ({ shouldRun }) => ({
-	shouldRun,
+export const ias: GetThirdPartyTag = () => ({
```
chrislomaxjones, line 8: It looks like, because of the signature of `GetThirdPartyTag`, even if we don't use the argument we still pass it in `third-party-tags.ts`.

```typescript
ias({ shouldRun: true })
```

Wonder if it's somehow worth changing the type of `GetThirdPartyTag`, or leaving this as-is and just passing in `true` for the value of `shouldRun`?

src/lib/dfp/init-slot-ias.ts
```diff
@@ -0,0 +1,128 @@
...
+		return result;
+	});
+};
+
+const initSlotIas = (id: string, slot: googletag.Slot) =>
+	timeout(
+		new Promise<void>((resolve) => {
+			/*
```
chrislomaxjones, line 33: The formatting of the multi line comments in this file looks a tad off, does Prettier not mind?

src/core/third-party-tags/ias.ts
Jakeii: I realised ias doesn't need to be a function if `shouldrun` is always true

chrislomaxjones: Good shout.

PR #1023, opened 2023-08-23
Fix bug when adding additional sizes to empty size mapping.

domlander:
## What does this change?

- Fixes a bug where additional sizes could be added to a breakpoint which has no existing size definition. When adding additional sizes, instead of appending the additional sizes to the existing ad sizes from the next smallest breakpoint with ad sizes, it would overwrite these sizes.

## Why?

- Inline slots on liveblog pages at phablet & tablet viewport sizes were only using the additional sizes and not the sizes defined for mobile viewport with the additional sizes.

## Screenshots

| Before | After |
| - | - |
| ![Before] | ![After] |

[Before]: https://github.com/guardian/commercial/assets/9574885/283a6bdb-bcd4-43a1-9cb1-45304743a8f5
[After]: https://github.com/guardian/commercial/assets/9574885/ca2bfb84-0f80-442e-8bee-55fec6c1818c

### DIFF
--- a/src/core/ad-sizes.ts
+++ b/src/core/ad-sizes.ts
@@ -397,7 +397,9 @@ const findAppliedSizesForBreakpoint = (
 		return [];
 	}
 
-	let breakpointIndex = breakpoints.findIndex((b) => b === breakpoint);
+	let breakpointIndex: 0 | 1 | 2 | 3 = breakpoints.findIndex(
+		(b) => b === breakpoint,
+	) as 0 | 1 | 2 | 3;
 
 	while (breakpointIndex >= 0) {
 		const breakpointToTry = breakpoints[breakpointIndex];
--- a/src/core/ad-sizes.ts
+++ b/src/core/ad-sizes.ts
@@ -1,3 +1,4 @@
+import { breakpoints, isBreakpoint } from './lib/breakpoint';
 import type { Breakpoint } from './lib/breakpoint';
 
 type AdSizeString = 'fluid' | `${number},${number}`;
@@ -178,6 +179,16 @@ const adSizes: Record<SizeKeys, AdSize> = {
  * mark: 432b3a46-90c1-4573-90d3-2400b51af8d0
  * Some of these may or may not need to be synced for with the sizes in ./create-ad-slot.ts
  * these were originally from DCR, create-ad-slot.ts ones were in frontend.
+ *
+ * Note:
+ * If a breakpoint is not defined in a size mapping for a slot, that breakpoint will use the sizes
+ * of the next breakpoint down that has a size mapping. For example, if only "mobile" and "phablet" sizes
+ * are defined for a slot, all breakpoints larger than "phablet" will use the mapping for "phablet".
+ *
+ * In another example, if a slot has only "tablet" as a size mapping defined,
+ * then "desktop" will use the size mapping for "tablet". "mobile" and "phablet"
+ * will have no size mapping. This type of example may be used in cases where
+ * we only want the slot to appear on the "tablet" size or greater.
  **/
 const slotSizeMappings: SlotSizeMappings = {
 	inline: {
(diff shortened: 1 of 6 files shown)


src/core/ad-sizes.ts
```diff
@@ -371,6 +382,38 @@ const slotSizeMappings: SlotSizeMappings = {
...
+ */
+const findAppliedSizesForBreakpoint = (
+	sizeMappings: SizeMapping,
+	breakpoint: Breakpoint,
+): AdSize[] => {
+	if (!isBreakpoint(breakpoint)) {
+		return [];
+	}
+
+	let breakpointIndex: 0 | 1 | 2 | 3 = breakpoints.findIndex(
```
domlander, line 400: Really not happy with this type. If we add another breakpoint to the `breakpoints` array we will need to update this type here. I couldn't find another way to get around the `noUncheckedIndexedAccess` typescript rule, so please shout if you have an idea!

Run `yarn tsc --project ./tsconfig.core.json` locally to test.

Jakeii:
Rather than adding the complexity of `findAppliedSizesForBreakpoint` could we instead just have a typeguard to ensure that additionalSizes are only added for breakpoints that exist in the adsizes?

domlander:
> Rather than adding the complexity of `findAppliedSizesForBreakpoint` could we instead just have a typeguard to ensure that additionalSizes are only added for breakpoints that exist in the adsizes? and restore that breakpoint that's missing and causing the bug.

We could use a typeguard. I don't see a way we could introduce a bug without a typeguard catching it. 

The reason I thought the function would be a good idea is so that if we want to add more additional sizes in the future, it will just work without having to consider the original definitions. If we had a type guard instead, we would have to duplicate a breakpoints size mappings whenever we wanted to add additional sizes. We'd also have to remember how the slot size mappings work (I've added a bunch of comments to help with this). That's why I thought a function would be better here.

src/core/ad-sizes.ts
```diff
@@ -371,6 +385,38 @@ const slotSizeMappings: SlotSizeMappings = {
...
 const getAdSize = (size: SizeKeys): AdSize => adSizes[size];
 
+/**
+ * Finds the ad sizes that will be used for a breakpoint given a size mapping
+ *
+ * If ad sizes are defined in the size mapping for the specified breakpoint, we use that.
+ * If no sizes are defined for the breakpoint, use the next smallest breakpoint with defined ad sizes.
+ * If no smaller breakpoints have defined ad sizes, return an empty array
+ */
+const findAppliedSizesForBreakpoint = (
```
Jakeii, line 395: Could you maybe throw a couple of examples in the comment?

Jakeii:
> > Rather than adding the complexity of `findAppliedSizesForBreakpoint` could we instead just have a typeguard to ensure that additionalSizes are only added for breakpoints that exist in the adsizes? and restore that breakpoint that's missing and causing the bug.
> 
> We could use a typeguard. I don't see a way we could introduce a bug without a typeguard catching it.
> 
> The reason I thought the function would be a good idea is so that if we want to add more additional sizes in the future, it will just work without having to consider the original definitions. If we had a type guard instead, we would have to duplicate a breakpoints size mappings whenever we wanted to add additional sizes. We'd also have to remember how the slot size mappings work (I've added a bunch of comments to help with this). That's why I thought a function would be better here.

I guess I was envisioning where we'd have a much stricter type of `slotSizeMapping` using `as const satisfies` or something. Allowing at compile time to know that you're trying to concat to a breakpoint (or slot) that does or doesn't exists.

But I've just had a quick experiment I think it'll probably require a pretty big refactor of the sizeMappings at best and might be impossible at worst!😅

src/core/ad-sizes.ts
domlander: Will do. There's also more examples in the tests for this function

Jakeii: sweet

Jakeii (approved):

Jakeii:
Great documentation improvements!

PR #1032, opened 2023-08-29
Add merch-high ad size to fronts-banner ads

domlander:
## What does this change?

- Adds the merch-high ad size (88x87) to fronts-banner ad slots

## Why?

- So we can display adverts targeting the 88x87 size in these slots

## Screenshots

Screenshot 2023-08-29 at 15 29 14


chrislomaxjones (approved):
👍

PR #1035, opened 2023-08-30
Record Cypress videos in CI

chrislomaxjones:
## What does this change?

- Record Cypress videos when running in CI, detected via the [`CI`](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) environment variable.
- Only upload the captured videos as artifacts when the E2E test suite fails.

## Why?

This should make it slightly easier to debug any issues with failing tests running in CI.

I've opted to switch to only upload artifacts on failure to minimise the amount of videos stored that we'll likely never watch.


Jakeii (approved):
Can't see a downside to this 👍

PR #1036, opened 2023-08-30
Sentry sampling for Okta test participants

emma-imber:
## What does this change?
Stop bypassing the Sentry sampling rate for Okta test participants.

## Why?
When we migrated to Okta we wanted to see all Sentry errors from Okta test participants to keep an eye out for any new errors. Now that the test is going to 10%, we want to reintroduce sampling so that we don't flood Sentry with error messages.


chrislomaxjones (approved):
👍

PR #1039, opened 2023-08-31
Make Okta a feature switch, not an experiment

georgeblahblah:
As we roll Okta out to 100% of users, we will no longer use an experiment to determine whether someone should use Okta. Instead, we need to check whether the Okta switch is on or off.

One side effect of this is the Cypress tests. Because the switch value is fetched from prod Frontend, the Cypress tests try to use Okta for signed in features. Whilst this is the desired end goal, we want to continue using cookie-based auth in Cypress tests for now. This change adds a "Standard" article fixture, which we use as the request body to DCR's `POST /Article` endpoint. This enables us to set the `okta` and `idCookieRefresh` switches to false.

Part of https://github.com/guardian/frontend/issues/26551.


arelra (approved):
lgtm - just let us know when you want to release this @georgeblahblah as we'll need to bump commercial in Frontend

emma-imber:
Is it worth waiting for this to be merged and then we can use the same code pattern for fixtures? https://github.com/guardian/commercial/pull/948

arelra:
> Is it worth waiting for this to be merged and then we can use the same code pattern for fixtures? #948

It's a good shout but there's a bit of time pressure on this @emma-imber so I agreed with @georgeblahblah that we could retro fit this test.

PR #1051, opened 2023-09-11
Add publisher ID for improve digital

emma-imber:
## What does this change?
Add the publisher ID to the improveDigitalBidder. I've kept it as an optional parameter for now as we don't need to send it for the improve pageskin bidder.

Screenshot 2023-09-11 at 16 46 52

## Why?
To address an incoming request from Improve Digital


chrislomaxjones (approved):
👍 

No action required, but if we remove the Improve skin bidder in the future we should make `publisherId` a required parameter.

PR #1053, opened 2023-09-13
A new size for sponsor logos

chrislomaxjones:
## What does this change?

This PR adds a new `3x3` _proxy_ ad size for sponsor logos.

Building off the work that began in #970, this size is designed to address issues where high priority MPUs can fill the ad slots placed inside special sponsored thrashers (such as the US soccer thrasher). In order to avoid this, we create a new 3x3 ad size `sponsorLogo` that no other line items will use, ensuring these thrashers are only filled with specific lines for sponsors.

### What about external ads?

We add a new slot name `sponsor-logo` for use with these thrashers, and assume there'll be at most one sponsor logo on a page. We also mark the `external` slot as deprecated, with a view to moving over the existing Soccer thrasher to use the new slot.

This PR also includes an end-to-end test that uses our new fixtures server to insert a thrasher container one of these logos onto a front, which we then test is correctly filled.

## Why?

This new size will prevent high priority MPUs (e.g. from home page takeovers) from filling the slots designed for sponsor logos.

## Screenhots

Note the small "Sponso" (excuse my dodgy cropping) logo advert in the bottom right of the thrasher.

Screenshot 2023-09-14 at 15 09 08

## Next steps

- [ ] Migrate over the existing US Soccer thrasher to use this new slot.
- [ ] Ensure that any existing line items then use the new `3x3` size.


cypress/e2e/parallel-2/sponsor-logo.cy.ts
```diff
@@ -0,0 +1,28 @@
+import { getStage, getTestUrl } from '../../lib/util';
+
+describe('sponsorshipLogo', () => {
+	beforeEach(() => {
+		cy.useConsentedSession('sponsorship-logo');
+	});
+
+	it('sponsor logo ad is correctly filled in thrasher fixture', () => {
+		// Construct a path that uses a fixture where a thrasher containers a sponsor logo
```
emma-imber, line 9: Tiny typo:
```suggestion
		// Construct a path that uses a fixture where a thrasher contains a sponsor logo
```

emma-imber (approved):
Nice!

arelra (approved):
🖼️

cypress/e2e/parallel-2/sponsor-logo.cy.ts
chrislomaxjones: Thanks!

PR #1054, opened 2023-09-14
Use `satisfies` when defining ad sizes and size mappings

chrislomaxjones:
## What does this change?

Use the new TypeScript [`satisfies`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator) operator when defining ad sizes and size mappings. I've combined this with `as const` in order to infer the narrowest possible types, allowing things like in Intellisense:

Screenshot 2023-09-14 at 10 05 43

Doing this required size mappings use readonly arrays of ad-sizes (and some fixing up in a place or two to satisfy this). I think this is probably a good thing!

## Why?

This allows us to infer a narrower type in each case, whilst still asserting that ad sizes and size mappings have a certain shape:

- In the case of ad sizes, this doesn't have a material effect, since the type inferred is the same as the one that is satisfied.
- In the case of size mappings, it means we can now access properties deeply.

This currently only has limited utility, but could be extended in the future (it'll allow certain types of per-property TSDoc to be surfaced, or for us to use a narrower ad size type).

### DIFF
--- a/.changeset/forty-rice-attend.md
+++ b/.changeset/forty-rice-attend.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': patch
+---
+
+Use satsifies operator when defining ad sizes and slot mappings config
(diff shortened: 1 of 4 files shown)


emma-imber (approved):
Nice!

PR #1057, opened 2023-09-14
AB test to run prebid eagerly

Jakeii:
## What does this change?
The is a follow up to #811 

This removes the previous test code to run prebid on page load.

Adds new code to run prebid when 50% of the viewport away from an ad slot, with 2 variants to try this with the current lazy load margin and a reduced lazy load margin.

<table>
<tr>
 <td><strong>variant
 <td><strong>prebid margin
 <td><strong>display margin
<tr>
 <td>control
 <td>N/A
 <td>20%
<tr>
 <td>variant-1
 <td>50%
 <td>20%
<tr>
 <td>variant-2
 <td>50%
 <td>10%
</table>

These values are have been picked pretty arbitrarily, open to suggestions!

My thinking is that there's not much point testing multiple variants of the prebid margin, as long as it's sufficiently earlier than the display margin.

## Why?

Running prebid ahead of time should allow the ad to load faster, leading to higher viewability/ad ratio and revanue, while also reducing layout shift.


src/lib/spacefinder/liveblog-adverts.ts
```diff
@@ -119,17 +114,15 @@ const insertAdAtPara = (para: Node): Promise<void> => {
...
-		});
+			}),
+		);
 };
 
 const insertAds: SpacefinderWriter = async (paras) => {
 	const fastdomPromises = [];
 	for (let i = 0; i < paras.length && AD_COUNTER < MAX_ADS; i += 1) {
 		const para = paras[i];
+		// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- false positive
```
chrislomaxjones, line 125: You might be able to remove these comments if you rebase off the latest main, which could require passing `--no-verify` to `git push` if the validation is failing. Happy for you to leave them if that's too much of a fuss though.

src/lib/dfp/lazy-load.ts
```diff
@@ -35,37 +61,53 @@ const onIntersect = (
...
+		});
 
-const getObserver = once(() => {
-	return Promise.resolve(
-		new window.IntersectionObserver(onIntersect, {
-			rootMargin: decideLazyLoadMargin(),
-		}),
+	dfpEnv.advertsToLoad = dfpEnv.advertsToLoad.filter(
+		(advert) => !advertIds.includes(advert.id),
 	);
```
chrislomaxjones, line 85: Just for my understanding, but here we're effectively marking these adverts as loaded at the point we run header bidding? Is it important we do this here, rather than doing it solely in the display ad observer?

chrislomaxjones (approved):
Looks good, this will make for an interesting experiment.

src/lib/dfp/lazy-load.ts
Jakeii: Oh interesting question, I didn't notice this was in both observers. I don't think will have any negative consequences as I don't think the `advertsToLoad` queue is ever used again after the page and ads have loaded.

I'll remove this duplicate though as it's just confusing!

PR #1058, opened 2023-09-15
Add ix to acBidders in realTimeData

emma-imber:
## What does this change?
Adds 'ix', the bidder name for indexSSP, to acBidders in the realTimeData config.

## Why?
This will allow them to access our Permutive audience segments so that they have audience data for targeting deals.

**Before:**

Screenshot 2023-09-15 at 10 55 57

**After:**

Screenshot 2023-09-14 at 17 21 56


chrislomaxjones (approved):
👍

PR #1060, opened 2023-09-18
Fix some noUncheckedIndexedAccess rule errors

emma-imber:
## What does this change?
Fixes a few of the more simple errors that result from switching on the noUncheckedIndexedAccess rule for the bundle code.

## Why?
Reduces the number of errors left to tackle:
Before:
Screenshot 2023-09-15 at 15 27 47
After:
Screenshot 2023-09-15 at 16 51 47


arelra (approved):
Nice! ✨

PR #1064, opened 2023-09-19
Fix noUncheckedIndexedAccess errors for files in the header bidding folder

emma-imber:
## What does this change?
Fixes noUncheckedIndexedAccess errors for files in the `header-bidding` folder

## Why?
Reduces the number of these errors by 16.

### DIFF
--- a/src/lib/header-bidding/prebid/bid-config.spec.ts
+++ b/src/lib/header-bidding/prebid/bid-config.spec.ts
@@ -540,11 +540,13 @@ describe('bids', () => {
 			[createAdSize(728, 90)],
 			mockPageTargeting,
 		)[2];
-		expect(openXBid.params).toEqual({
-			customParams: 'someAppNexusTargetingObject',
-			delDomain: 'guardian-d.openx.net',
-			unit: '540279541',
-		});
+		if (typeof openXBid !== 'undefined') {
+			expect(openXBid.params).toEqual({
+				customParams: 'someAppNexusTargetingObject',
+				delDomain: 'guardian-d.openx.net',
+				unit: '540279541',
+			});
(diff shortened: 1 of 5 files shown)


src/lib/header-bidding/prebid/bid-config.spec.ts
```diff
@@ -540,11 +540,13 @@ describe('bids', () => {
...
-			delDomain: 'guardian-d.openx.net',
-			unit: '540279541',
-		});
+		if (typeof openXBid !== 'undefined') {
+			expect(openXBid.params).toEqual({
+				customParams: 'someAppNexusTargetingObject',
+				delDomain: 'guardian-d.openx.net',
+				unit: '540279541',
+			});
+		}
```
chrislomaxjones, line 549: You could make this a touch more concise with optional chaining

```suggestion
	expect(openXBid?.params).toEqual({
		customParams: 'someAppNexusTargetingObject',
		delDomain: 'guardian-d.openx.net',
		unit: '540279541',
	});
```

domlander (approved):
🧹

src/lib/header-bidding/prebid/bid-config.spec.ts
emma-imber: Yeah fair point! I initially tried this but it triggered the `no-unnecessary-condition` rule, and I preferred to minimise the number of times I had to disable eslint rules. But still did it anyway in another file 😅

chrislomaxjones: Yeah introducing `no-unnecessary-condition` errors once we turn the rule off is a tricky one actually

PR #1067, opened 2023-09-20
remove logic related to expiry-ad-free-cookie and its switch

dskamiotis:
## What does this change?
This PR removes any related logic to the `AdFreeStrictExpiryEnforcement` switch.
This was also removed from the related PR in `frontend` [here](https://github.com/guardian/frontend/pull/26587)
## Why?

Currently we have a switch `AdFreeStrictExpiryEnforcement` 
When this switch is ON and the website loads, it checks to see if the user's ad-free cookie has expired. If so, it will refetch data to update this cookie. This switch has been OFF for a long time and is no longer needed.
We are not concerned with how old the cookie is so therefore we are removing any logic which checks for date or expiry for that cookie.

so the `ad-free-strict-expiry-enforcement` the logic related to this along with the tests have been removed.

### DIFF
--- a/.changeset/rude-lemons-eat.md
+++ b/.changeset/rude-lemons-eat.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': patch
+---
+
+removing ad free expiry switch and logic:
--- a/cypress/fixtures/pages/articles.ts
+++ b/cypress/fixtures/pages/articles.ts
@@ -1497,7 +1497,6 @@ export const Standard = {
 			adaptiveSite: true,
 			prebidPermutiveAudience: true,
 			compareVariantDecision: false,
-			adFreeStrictExpiryEnforcement: false,
 			enableSentryReporting: true,
 			lazyLoadContainers: true,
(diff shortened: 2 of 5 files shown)


domlander (approved):
I would change this line in the PR
`When the website loads it checks to see if the user is eligible for a ad-free experience.`
to
`When this switch is ON and the website loads, it checks to see if the user's ad-free cookie has expired. If so, it will refetch data to update this cookie. This switch has been OFF for a long time and is no longer needed.`
for clarity.

I'd recommend testing this on CODE before release, but I think you may have done that already!
``

dskamiotis:
> I would change this line in the PR `When the website loads it checks to see if the user is eligible for a ad-free experience.` to `When this switch is ON and the website loads, it checks to see if the user's ad-free cookie has expired. If so, it will refetch data to update this cookie. This switch has been OFF for a long time and is no longer needed.` for clarity.
> 
> I'd recommend testing this on CODE before release, but I think you may have done that already! ``

Thanks Tested on CODE via Frontend

PR #1069, opened 2023-09-20
Unify methods to fill advert slots

chrislomaxjones:
## What does this change?

This PR is a refactor that merges the two different ways we can fill advert slots on the page.

Previously, non-dynamic (static, and what we previously called _external_) adverts could be filled either via:
- `fill-advert-slots.ts`: this is where our bundle selects all of the slots from the DOM and attempts to fill them. We refer to the adverts picked up in this phase as _static_ ad slots. We assume they're server side rendered, to guarantee presence in the DOM, regardless of when the commercial JS runs.
- `fill-external-slots.ts`: this is where we select all of the external ads (marked by a prefix in their ids) from the DOM, but also add an event listener so external code can fire an event that notifies us to fill a slot. This is useful for when we can't guarantee the slot will be in the DOM before the commercial JS runs.

This PR unifies the two methods above into one by first renaming `fill-advert-slots.ts` to `fill-static-advert-slots.ts`, and removing the external path. The main observation here is that the external slots already present in the DOM will be filled as if they were static.

For any slots that we miss, we add an event listener as before so we can be notified that they require filling. We add this listener **before** we attempt to fill advert slots, so that we don't miss any events. **Note this requires all adverts that aren't server-side rendered must fire an event, in case the commercial JS hasn't run before the slot is present in the DOM**.

Finally, we add an E2E test with a thrasher that adds a slot with a couple of seconds of delay, before firing the event. This will exercise this new code path.

## Why?

This should simplify the number of ways through which we can fill a slot on the page.


src/lib/dfp/fill-static-advert-slots.ts
```diff
@@ -39,15 +39,15 @@ const decideAdditionalSizes = (adSlot: HTMLElement): SizeMapping => {
...
  * For dynamic ad slots that are created at js-runtime, see:
  *  - article-aside-adverts
  *  - article-body-adverts
  *  - liveblog-adverts
  *  - high-merch
  */
-const fillAdvertSlots = async (): Promise<void> => {
+const fillStaticAdvertSlots = async (): Promise<void> => {
 	// This module has the following strict dependencies. These dependencies must be
 	// fulfilled before fillAdvertSlots can execute reliably. The bootstrap
```
domlander, line 52: This comment refers to the old name of the function

chrislomaxjones:
> Is it possible that a gu.commercial.slot.fill event may be fired between the eventListener being created and the static ad slots on the page are collected? If so, this could mean displayAd() is called twice for an ad slot

I believe it is possible, although I think there's guards in place to prevent displayAd being called for a slot that's already been filled. It's just important we don't start listener some time after collecting up the slots, as this leaves a window in which we can miss eventd.

I'll leave it as it is for now, but I'll have a think about a way to make the guard logic more clear (it shouldn't matter if you fire an event for a slot that's already been filled, or vice versa).

> What do you think to moving the call to createSlotFillListener(); to line 72 of fill-static-advert-slots.ts? That way, we don't set the listener if the user is adFree and we don't setup Prebid between the creation of the eventListener and the collection of the static ad slots.

I'm going to remove that ad free check as it's superfluous - we perform that check much [earlier](https://github.com/guardian/commercial-core/blob/3c72ace3e1350ef0e695b66d3135d0ead2542af5/src/standalone.commercial.ts#L86).

I've never seen issues with dynamic ad slots being filled before Prebid is ready, but it might be worth checking that separately.

PR #1071, opened 2023-09-21
Various `noUncheckedIndexedAccess` fixes

chrislomaxjones:
## What does this change?

Fixe instances of no unchecked indexed access errors in:
- src/lib/consentless/dynamic/liveblog-inline.ts
- src/lib/dfp/init-slot-ias.spec.ts
- src/lib/dfp/prepare-permutive.spec.ts
- src/lib/spacefinder/article-aside-adverts.ts
- src/lib/spacefinder/liveblog-adverts.ts
- src/lib/third-party-tags.ts
- src/lib/utils/geolocation.ts

I've also switched the `no-unnecessary-condition` rule to `warn` whilst we squash these errors. We can re-enable it once this work is complete.

## Why?

Better type safety once the rule is enabled.

### DIFF
--- a/src/lib/utils/geolocation.ts
+++ b/src/lib/utils/geolocation.ts
@@ -1,13 +1,14 @@
 import { getCookie, isString, storage } from '@guardian/libs';
 import type { CountryCode } from '@guardian/libs';
+import type { Edition } from 'types/global';
 
-const editionToGeolocationMap: Record<string, CountryCode> = {
+const editionToGeolocationMap: Record<Edition, CountryCode> = {
 	UK: 'GB',
 	US: 'US',
 	AU: 'AU',
 };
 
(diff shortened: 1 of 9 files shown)


src/lib/utils/geolocation.ts
```diff
@@ -1,13 +1,14 @@
 import { getCookie, isString, storage } from '@guardian/libs';
 import type { CountryCode } from '@guardian/libs';
+import type { Edition } from 'types/global';
 
-const editionToGeolocationMap: Record<string, CountryCode> = {
+const editionToGeolocationMap: Record<Edition, CountryCode> = {
 	UK: 'GB',
 	US: 'US',
 	AU: 'AU',
 };
```
chrislomaxjones, line 9: We should update the `Edition` type as well.

emma-imber (approved):

PR #1072, opened 2023-09-21
Fix noUncheckedIndexedAccess errors for messenger files

emma-imber:
## What does this change?
Fixes the noUncheckedIndexedAccess errors in messenger files.

## Why?
Reduces the total error count by 5.

### DIFF
--- a/src/lib/messenger/get-stylesheet.ts
+++ b/src/lib/messenger/get-stylesheet.ts
@@ -14,7 +14,6 @@ const getStyles = (
 ): string[] => {
 	const result: string[] = [];
 	for (let i = 0; i < styleSheets.length; i += 1) {
-		// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- Fixes noUncheckedIndexedAccess error
 		const ownerNode = styleSheets[i]?.ownerNode;
 
 		if (
--- a/src/lib/messenger/get-stylesheet.ts
+++ b/src/lib/messenger/get-stylesheet.ts
@@ -14,7 +14,8 @@ const getStyles = (
 ): string[] => {
 	const result: string[] = [];
 	for (let i = 0; i < styleSheets.length; i += 1) {
-		const ownerNode = styleSheets[i].ownerNode;
+		// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- Fixes noUncheckedIndexedAccess error
+		const ownerNode = styleSheets[i]?.ownerNode;
 
 		if (
 			ownerNode instanceof HTMLStyleElement &&
--- a/src/lib/messenger/passback.ts
+++ b/src/lib/messenger/passback.ts
@@ -198,8 +198,8 @@ const init = (register: RegisterListener): void => {
 							const slotId = event.slot.getSlotElementId();
 							if (slotId === passbackElementId) {
 								const size = event.size;
-								if (Array.isArray(size)) {
-									const adHeight = Number(size[1]);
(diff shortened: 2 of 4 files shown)


src/lib/messenger/get-stylesheet.ts
```diff
@@ -14,7 +14,8 @@ const getStyles = (
 ): string[] => {
 	const result: string[] = [];
 	for (let i = 0; i < styleSheets.length; i += 1) {
-		const ownerNode = styleSheets[i].ownerNode;
+		// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- Fixes noUncheckedIndexedAccess error
```
domlander, line 17: I don't think we'll need these comments after Chris' PR, as he's temporarily set the `no-unnecessary-condition` errors to be warnings until we've finished this work: 

https://github.com/guardian/commercial/pull/1071/files

src/lib/messenger/passback.ts
```diff
@@ -199,7 +199,7 @@ const init = (register: RegisterListener): void => {
 							if (slotId === passbackElementId) {
 								const size = event.size;
 								if (Array.isArray(size)) {
-									const adHeight = size[1];
+									const adHeight = Number(size[1]);
```
chrislomaxjones, line 202: I think there might be two options here which would avoid the case where you call `Number(undefined)` at runtime (which may or may not be possible, depending on the assumptions below):

1) Assume that googletag may return improperly sized arrays and perform the check in the if statement:

```typescript
if (Array.isArray(size) && size[1]) {
  // ...
}
```

2) Assume googletag always returns an array of the correct size, and either ignore the rule or use a non-null assertion:

```typescript
const adHeight = size[1]!;
```

Note this will generate an eslint warning, which could be ignored / disabled with a comment.

src/lib/messenger/scroll.ts
```diff
@@ -38,13 +38,14 @@ const domRectToRect = (rect: DOMRect) => ({
...
+	iframes[iframeId]?.respond(null, domRectToRect(domRect));
 };
 
-const getDimensions = (id: string): [string, DOMRect] => [
-	id,
-	iframes[id].node.getBoundingClientRect(),
-];
+const getDimensions = (id: string): [string, DOMRect] => {
+	// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion  -- Fixes a type error caused by the optional chain
+	const node = <HTMLIFrameElement>iframes[id]?.node;
```
chrislomaxjones, line 46: Just to confirm my understanding, are we assuming that `iframes[id]` is always defined here, as we're casting to tell TypeScript it'll never be undefined?

src/lib/messenger/passback.ts
emma-imber: Great shout, I'll implement option 1

src/lib/messenger/get-stylesheet.ts
emma-imber: Yep good shout! I've rebased now so that I can remove the comments

src/lib/messenger/scroll.ts
emma-imber: Yes we are - I'm not thrilled about handling it this way, but allowing the function to return undefined causes more type errors downstream where getDimensions is used. Very open to nicer ideas of how we can handle it!

chrislomaxjones: No it's all good, just wanted to check!

PR #1074, opened 2023-09-25
Stop recording Cypress videos in CI

chrislomaxjones:
Reverts guardian/commercial#1035.

I've found I get lots of errors when compressing the video artefacts in CI. Since we're moving away from Cypress, I figure we may as well revert this.

### DIFF
--- a/.github/workflows/e2e-cypress.yml
+++ b/.github/workflows/e2e-cypress.yml
@@ -62,11 +62,8 @@ jobs:
         run: yarn cypress run --spec cypress/e2e/parallel-${{ matrix.group }}/*
         working-directory: ./commercial
 
-      # Only capture videos and screenshots on failure
-      # In order to minimize unused artifacts
-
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-cypress.yml
```diff
@@ -62,11 +62,8 @@ jobs:
         run: yarn cypress run --spec cypress/e2e/parallel-${{ matrix.group }}/*
         working-directory: ./commercial
 
-      # Only capture videos and screenshots on failure
-      # In order to minimize unused artifacts
-
       - uses: actions/upload-artifact@v2
-        if: failure()
+        if: always()
```
chrislomaxjones, line 66: I'll leave this here in case anyone wants to override `video: true` in the Cypress config of their branch to do debugging.

Jakeii (approved):
Good shout 👍

PR #1075, opened 2023-09-26
fix `noUncheckedIndexedAccess` from dfp directory

dskamiotis:
## What does this change?
Fixes instances of no unchecked indexed access errors in following Dir:

- src/lib/dfp/display-ads.ts:24
- src/lib/dfp/get-advert-by-id.ts:5
- src/lib/dfp/on-slot-render.ts:40

## Why?
Better type safety once the rule is enabled.

### DIFF
--- a/src/lib/dfp/display-ads.ts
+++ b/src/lib/dfp/display-ads.ts
@@ -1,4 +1,5 @@
 import { pageSkin } from '../creatives/page-skin';
+import type { Advert } from './Advert';
 import { dfpEnv } from './dfp-env';
 import { loadAdvert } from './load-advert';
 
@@ -21,8 +22,12 @@ const displayAds = (): void => {
 	window.googletag.pubads().collapseEmptyDivs();
 	window.googletag.enableServices();
 
-	loadAdvert(dfpEnv.advertsToLoad[0]);
-	dfpEnv.advertsToLoad.length = 0;
+	const firstAdvertToLoad: Advert | undefined = dfpEnv.advertsToLoad[0];
+
+	if (firstAdvertToLoad) {
+		loadAdvert(firstAdvertToLoad);
+		dfpEnv.advertsToLoad.length = 0;
+	}
 	pageSkin();
 };
 
--- a/src/lib/dfp/display-ads.ts
+++ b/src/lib/dfp/display-ads.ts
@@ -28,7 +28,7 @@ const displayAds = (): void => {
 
 	if (firstAdvertToLoad) {
 		loadAdvert(firstAdvertToLoad);
-		dfpEnv.advertsToLoad.length = 0;
+		dfpEnv.advertsToLoad = [];
 	}
 	pageSkin();
 };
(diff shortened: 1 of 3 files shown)


src/lib/dfp/display-ads.ts
```diff
@@ -21,8 +22,14 @@ const displayAds = (): void => {
...
 
-	loadAdvert(dfpEnv.advertsToLoad[0]);
-	dfpEnv.advertsToLoad.length = 0;
+	const firstAdvertToLoad: Advert | undefined = dfpEnv.advertsToLoad.length
+		? dfpEnv.advertsToLoad[0]
+		: undefined;
+
+	if (firstAdvertToLoad) {
+		loadAdvert(firstAdvertToLoad);
+		dfpEnv.advertsToLoad.length = 0;
```
SiAdcock, line 31: Might not be the best place to raise this, but it's driving me a bit loopy. Can we update to:

```suggestion
		dfpEnv.advertsToLoad = [];
```

As far as I can tell, [we don't keep references to the original array](https://github.com/search?q=repo%3Aguardian%2Fcommercial%20dfpEnv.advertsToLoad&type=code) anywhere (which would be horrible to keep track of) and in [other](https://github.com/guardian/commercial/blob/de2bb5039bf670bc123560a613c7a57e6bc4870f/src/lib/dfp/lazy-load.ts#L33) [places](https://github.com/guardian/commercial/blob/de2bb5039bf670bc123560a613c7a57e6bc4870f/src/lib/dfp/display-lazy-ads.ts#L18) we totally create new arrays anyway.

Originally implemented in [frontend 13289](https://github.com/guardian/frontend/blame/4e9da3bde887ff0762bc43f117e55de4eac0e7b3/static/src/javascripts/projects/common/modules/commercial/dfp/dfp-api.js#L304)

dskamiotis: I think a valid suggestion @SiAdcock , and perhaps makes more logical sense looking at it. 
I would suggest amending it in this PR?

SiAdcock: Okay, go for it 👍 I'll take responsibility.

Thanks @dskamiotis

src/lib/dfp/display-ads.ts
```diff
@@ -21,8 +22,14 @@ const displayAds = (): void => {
 	window.googletag.pubads().collapseEmptyDivs();
 	window.googletag.enableServices();
 
-	loadAdvert(dfpEnv.advertsToLoad[0]);
-	dfpEnv.advertsToLoad.length = 0;
+	const firstAdvertToLoad: Advert | undefined = dfpEnv.advertsToLoad.length
+		? dfpEnv.advertsToLoad[0]
+		: undefined;
```
Jakeii, line 27: I think shortening this to
```suggestion
	const firstAdvertToLoad = dfpEnv.advertsToLoad[0];
```
Should still work

Jakeii (approved):

PR #1078, opened 2023-09-27
Add `fullwidth` messenger message

Jakeii:
## What does this change?

Add a messenger message to add a class that will set the slot to `width: 100%`

The class is added to DCR in this PR https://github.com/guardian/dotcom-rendering/pull/8917

## Why?
Usually front banner slots are restricted to the content area width (1300px), we would like our in house ad team and some ad partners to be able to have full-width ads in these slots.

Example usage from an ad:
```js
top.window.postMessage(JSON.stringify(
    {
        id: 'bf724866-723c-6b0a-e5d7-ad61535f98b7',
        type: 'full-width',
        value: true,
        source: 'clipcentric'

    }
), '*');
//send postMessage to commercial bundle
  top.window.postMessage(JSON.stringify(
    {
        id: 'bf724866-723c-6b0a-e5d7-ad61535f98b7',
        type: 'resize',
        value: {
        	width: '100vw',
         	height: '600px',
        },
      	source: 'clipcentric'
    }
), '*');
 
```

If the ad wanted to restrict to the column width again another message can be sent
```js
top.window.postMessage(JSON.stringify(
    {
        id: 'bf724866-723c-6b0a-e5d7-ad61535f98b7',
        type: 'full-width',
        value: false,
        source: 'clipcentric'

    }
), '*');

### DIFF
--- a/src/core/messenger.ts
+++ b/src/core/messenger.ts
@@ -13,6 +13,7 @@ type MessageType =
 	| 'measure-ad-load'
 	| 'passback'
 	| 'resize'
+	| 'fullwidth'
 	| 'set-ad-height'
 	| 'scroll'
 	| 'type'
--- a/src/core/messenger.ts
+++ b/src/core/messenger.ts
@@ -13,7 +13,7 @@ type MessageType =
 	| 'measure-ad-load'
 	| 'passback'
 	| 'resize'
-	| 'fullwidth'
+	| 'full-width'
 	| 'set-ad-height'
 	| 'scroll'
 	| 'type'
--- a/src/lib/messenger/fullwidth.ts
+++ b/src/lib/messenger/fullwidth.ts
@@ -9,12 +9,12 @@ interface FullwidthSpecs {
 const isFullwidthSpecs = (specs: unknown): specs is FullwidthSpecs =>
 	!!specs && isObject(specs) && 'fullwidth' in specs;
 
-const fullwidth = (specs: FullwidthSpecs, slotContainer: HTMLElement) =>
+const fullwidth = (specs: FullwidthSpecs, slot: HTMLElement) =>
 	fastdom.mutate(() => {
 		if (specs.fullwidth) {
-			slotContainer.classList.add('ad-slot-container--fullwidth');
+			slot.classList.add('ad-slot--full-width');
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: src/lib/messenger/fullwidth.ts [hunk 1]
```diff
@@ -0,0 +1,44 @@
+import { isObject } from '@guardian/libs';
+import fastdom from 'fastdom';
+import type { RegisterListener } from 'core/messenger';
+
+interface FullwidthSpecs {
+	fullwidth: boolean;
+}
+
+const isFullwidthSpecs = (specs: unknown): specs is FullwidthSpecs =>
+	!!specs && isObject(specs) && 'fullwidth' in specs;
```
chrislomaxjones, line 10: guardian/libs has got a `isBoolean` function if you wanted to validate the type of the `fullwidth` property as well

chrislomaxjones (approved):
Looks good

src/lib/messenger/fullwidth.ts
```diff
@@ -0,0 +1,44 @@
...
+			if (!isFullwidthSpecs(specs)) {
+				return;
+			}
+			const adSlot =
+				iframe.closest<HTMLElement>('.js-ad-slot') ?? undefined;
+
+			const name = adSlot?.dataset.name;
+
+			// only allow for banner ads
+			if (!name?.startsWith('fronts-banner') || !adSlot) {
```
domlander, line 33: if `adslot` is falsy then `name` will be undefined, so I don't think you need this second check

domlander (approved):
👍

src/core/messenger.ts
```diff
@@ -13,6 +13,7 @@ type MessageType =
 	| 'measure-ad-load'
 	| 'passback'
 	| 'resize'
+	| 'fullwidth'
```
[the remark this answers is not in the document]
Jakeii: ah thanks, I started changing it part way through, missed the message itself

src/lib/messenger/fullwidth.ts
Jakeii: I agree it's a bit pointless but typescript complains without it when calling `fullwidth`

PR #1079, opened 2023-09-27
Add video option to interscroller

emma-imber:
## What does this change?
Adds the functionality to messenger to create a video inside the interscroller ad if a video source has been specified.

## Why?
This allows us to support a video interscroller. The template will also be updated to add the videoSource parameter. This is an optional parameter that we check for, and if it's been added we create a video html element inside the interscroller ad.

This has been tested on CODE, demo of the video interscroller is below (video is supposed to be cropped, it's just a cropped version of another video):

https://github.com/guardian/commercial/assets/108270776/98a6a905-8e49-47ba-8fba-716a20196332

### DIFF
--- a/.changeset/empty-bananas-work.md
+++ b/.changeset/empty-bananas-work.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
+Add a video version of the interscroller template to messenger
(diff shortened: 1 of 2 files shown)


Jakeii (approved):
Nice work!

PR #1080, opened 2023-09-27
Use a `Map` to track adverts on the page

chrislomaxjones:
## What does this change?

- Remove the `advertIds` property from `dfpEnv` and instead store adverts as a `Map` from their string ids to the adverts themselves. This allows us to store the same information in a single structure, with clearer semantics around getting/setting elements.

- Remove the unused `creativeIds` property. It appears this array is only ever pushed to, and the values never retrieved from it.

## Why?

Should be clearer when managing which adverts are being tracked in the `dfpEnv`.

In the browser console, for `window.guardian.commercial.dfpEnv.adverts`:

Screenshot 2023-10-25 at 12 17 24

### DIFF
--- a/src/lib/dfp/dfp-env.ts
+++ b/src/lib/dfp/dfp-env.ts
@@ -9,7 +9,6 @@ interface DfpEnv {
 	adSlotSelector: string;
 	lazyLoadEnabled: boolean;
 	lazyLoadObserve: boolean;
-	creativeIDs: string[];
 	advertsToLoad: Advert[];
 	adverts: Map<Advert['id'], Advert>;
 	shouldLazyLoad: () => boolean;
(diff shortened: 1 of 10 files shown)


src/lib/dfp/dfp-env.ts
```diff
@@ -9,10 +9,8 @@ interface DfpEnv {
 	adSlotSelector: string;
 	lazyLoadEnabled: boolean;
 	lazyLoadObserve: boolean;
-	creativeIDs: string[];
-	advertIds: Record<string, number>;
 	advertsToLoad: Advert[];
-	adverts: Advert[];
+	adverts: Map<Advert['id'], Advert>;
```
domlander, line 13: Should this be renamed to `advertsMap` or similar to reflect the type of object is represents?

domlander (approved):
Nice refactor! Also a reminder that I should use Map more.

Good find on the unused creativeIDs 👍

src/lib/dfp/dfp-env.ts
chrislomaxjones: I personally think `adverts` is enough here, since the map is implied by the type. Happy to change it in the future if it proves confusing though.

chrislomaxjones:
> Nice refactor! Following #1121 we'll just have to update this line here too:
> 
> https://github.com/guardian/commercial/blob/c1dd4641048c86b96b054f00fba52e8e3d86c0a4/src/lib/header-bidding/prebid/bid-config.ts#L271

Thanks for letting me know!

PR #1081, opened 2023-09-28
fix more noUncheckedIndexedAccess

Jakeii:
## What does this change?
Fixes instances of no unchecked indexed access errors in following files
```
src/lib/identity/api.ts
src/lib/utils/$$.ts
src/lib/user-prefs.ts
src/lib/experiments/ab.spec.ts
```

## Why?

Better type safety

### DIFF
--- a/src/lib/user-prefs.ts
+++ b/src/lib/user-prefs.ts
@@ -68,29 +68,31 @@ const setPrefs = (loc: Location): void => {
 	let i: number;
 	let j: number;
 	for (i = 0, j = qs.length; i < j; i += 1) {
-		const m = /^gu\.prefs\.(.*)=(.*)$/.exec(qs[i]);
-		if (m) {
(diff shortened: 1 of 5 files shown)


src/lib/user-prefs.ts
```diff
@@ -68,29 +68,31 @@ const setPrefs = (loc: Location): void => {
...
+							set(key, +val);
+						} else if (isBoolean(val)) {
+							// String(val) === "true" converts a string to bool
+							set(key, String(val).toLowerCase() === 'true');
+						} else {
+							set(key, val);
+						}
 					} else {
-						v = val;
+						remove(key);
```
Jakeii, line 93: The original code allowed setting a storage value to undefined but that seems unintentional? Pretty confident this code is not used by us anyway

emma-imber (approved):

PR #1083, opened 2023-09-29
Fix remaining noUncheckedIndexedAccess errors

emma-imber:
## What does this change?
Fixes the final 3 errors we get when we enable the noUncheckedIndexedAccess rule.

## Why?
Gets us to 0 errors 📉

Screenshot 2023-09-29 at 15 24 45

### DIFF
--- a/src/lib/spacefinder/article-body-adverts.ts
+++ b/src/lib/spacefinder/article-body-adverts.ts
@@ -383,20 +383,20 @@ const attemptToAddInlineMerchAd = (): Promise<boolean> => {
 	};
 
 	const insertAds: SpacefinderWriter = (paras) => {
-		if (typeof paras[0] !== 'undefined') {
-			return insertAdAtPara(
-				paras[0],
-				'im',
-				'im',
-				'',
-				{},
-				{
-					className: 'ad-slot-container--im',
-				},
+		if (typeof paras[0] === 'undefined') {
+			throw new Error(
+				'Trying to insert inline merch before a node that does not exist',
 			);
 		}
-		throw new Error(
-			'Trying to insert inline merch before a node that does not exist',
+		return insertAdAtPara(
+			paras[0],
+			'im',
+			'im',
+			'',
+			{},
+			{
+				className: 'ad-slot-container--im',
+			},
 		);
 	};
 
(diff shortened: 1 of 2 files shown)


src/lib/spacefinder/article-body-adverts.ts
```diff
@@ -384,7 +384,7 @@ const attemptToAddInlineMerchAd = (): Promise<boolean> => {
 
 	const insertAds: SpacefinderWriter = (paras) =>
 		insertAdAtPara(
-			paras[0],
+			paras[0] as Node,
```
arelra, line 387: Wonder if we can type the first parameter to `insertAdAtPara` as a HTMLElement to avoid this cast?

emma-imber: The cast was to avoid dealing with undefined instead of there being a type conflict between HTMLElement and Node - happy to rethink how to do this without a cast though!

src/lib/third-party-tags/imr-worldwide.ts
```diff
@@ -67,22 +67,24 @@ const onLoad = (): void => {
 			? sectionFromMeta
 			: 'The Guardian - brand only';
 
-	const nolggGlobalParams = {
-		sfcode: 'dcr',
-		apid: subBrandApId,
-		apn: 'theguardian',
-	};
+	if (typeof subBrandApId === 'string') {
```
chrislomaxjones, line 70: I think you can avoid checking the type of `subBrandApId` here, since

```typescript
guMetadata[sectionFromMeta] || guMetadata['brand-only']
```

should always be defined (if we can't lookup the section in the object, we fall back to using the `brand-only` key, which is always present).

I think it'd be simpler to simply assert this yourself (e.g. add an `as string` to the declaration of `subBrandApId`).

 Or, possibly more complicated, to try make this explicit in the typeof `guMetadata`. I haven't tested this but something like:

```typescript
const guMetadata: {
	'brand-only': string;
	[key: string]: string | undefined;
} = {
  // ... the other values
  'brand-only': 'P0EE0F4F4-8D7C-4082-A2A4-82C84728DC59',
}
```

emma-imber: Will do 👍

chrislomaxjones: Nice thanks!

chrislomaxjones (approved):
🥳

arelra (commented):
Great work all we got there! 🥳

PR #1087, opened 2023-10-02
Catch errors from Okta

georgeblahblah:
I think uncaught exceptions from the Identity Auth library might be contributing to the spike in dropped errors in Sentry.

In DCR, we catch errors and print them to console. I think it is sensible to do the same here.

https://github.com/guardian/dotcom-rendering/blob/7fe3c9adf5cddc28830d433421a69085a2415fb2/dotcom-rendering/src/lib/useAuthStatus.ts#L88-L102

Part of https://github.com/guardian/dotcom-rendering/issues/8820

See [this chat thread](https://chat.google.com/room/AAAAG9rU0m0/jgCCywOVkm8) for more info.

### DIFF
--- a/src/lib/identity/api.ts
+++ b/src/lib/identity/api.ts
@@ -118,15 +118,20 @@ const isUserLoggedIn = (): boolean => getUserFromCookie() !== null;
 
 const getAuthStatus = async (): Promise<AuthStatus> => {
 	if (useOkta) {
-		const { isSignedInWithOktaAuthState } = await import('./okta');
-		const authState = await isSignedInWithOktaAuthState();
-		if (authState.isAuthenticated) {
-			return {
-				kind: 'SignedInWithOkta',
-				accessToken: authState.accessToken,
-				idToken: authState.idToken,
-			};
+		try {
+			const { isSignedInWithOktaAuthState } = await import('./okta');
+			const authState = await isSignedInWithOktaAuthState();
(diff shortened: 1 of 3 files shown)


chrislomaxjones (approved):
This makes sense to me, thanks!

src/lib/identity/api.ts
```diff
@@ -118,15 +118,20 @@ const isUserLoggedIn = (): boolean => getUserFromCookie() !== null;
...
+			const authState = await isSignedInWithOktaAuthState();
+			if (authState.isAuthenticated) {
+				return {
+					kind: 'SignedInWithOkta',
+					accessToken: authState.accessToken,
+					idToken: authState.idToken,
+				};
+			}
+		} catch (e) {
+			console.error(e);
```
arelra, line 132: Non-blocking. Would it be useful to copy the reasons from the chat thread?

```suggestion
			// Errors:
			// the user's token is invalid/expired
			// the request to refresh the token timed out
			// the user's clock skew is too large (> 5 minutes)
			console.error(e);
```

arelra: Just missed the merge. No worries we can add. Thanks!

georgeblahblah: Oh that's a good suggestion, sorry I missed it! it also looks like this change has not addressed the original issue

PR #1094, opened 2023-10-06
Playwright fixes: consent and article inline tests

domlander:
## What does this change?

- Adds a couple of script which should make it easier for new users to be able to run Playwright

### Fixes consent tests
 
- Added basic "Accept cookies -> ad slots fulfilled test"
- Reordered tests
- Removed skipped "cookie expiry" tests. Commercial only check the existence of cookies (ad free and digital subscriber) when checking consent, so these tests were attempting to test implementation details outside of our domain. Therefore they are not needed.

### Fixes article inline slots tests

- We were doing a count() to get the expected number of ad slots which resolve immediately. This was waiting for the first inline slot to appear before running. However, Spacefinder runs through it's logic twice when inserting inline ad slots on article pages. This change ensures we wait until Spacefinder has completely finished inserting inline ad slots before we take a count.

### Skipped liveblog live update tests

- These are the only tests still not working 100% of the time, so they are skipped for now.

### Pageskin tests not amended

- Pageskin tests pass locally but fail on CI. These will be subject to a future PR.

### DIFF
--- a/.vscode/settings.json.recommended
+++ b/.vscode/settings.json.recommended
@@ -1,5 +1,5 @@
 // Recommended VS Code settings for this project.
-// Copy these to .vscode/settings.json to apply them to your editor – but don't commit it!
+// Copy these to .vscode/settings.json to apply them to your editor
 // https://github.com/guardian/recommendations/blob/main/VSCode.md#do-not-commit-vscode
 {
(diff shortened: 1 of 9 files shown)


arelra (approved):
Great work @domlander on these fixes!

PR #1096, opened 2023-10-06
Skip failing playwright test

domlander:
## What does this change?

- Skip the playwright test that always fails on CI

## Why?

- So all Playwright tests pass on CI. There will be a follow up PR to fix the failing test

### DIFF
--- a/playwright/tests/parallel-4/pageskin.spec.ts
+++ b/playwright/tests/parallel-4/pageskin.spec.ts
@@ -52,12 +52,15 @@ test.describe('pageskin on uk front', () => {
 		});
 	});
 
-	/**
-	 * TODO e2e flakey test
-	 * The test for tablet always passes locallly but always fails on CI
-	 */
 	small.forEach(({ breakpoint, width, height }) => {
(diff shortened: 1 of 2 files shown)


playwright/tests/parallel-4/pageskin.spec.ts
```diff
@@ -57,7 +57,7 @@ test.describe('pageskin on uk front', () => {
 	 * The test for tablet always passes locallly but always fails on CI
 	 */
 	small.forEach(({ breakpoint, width, height }) => {
-		test(`Test pageskin front on ${breakpoint} should NOT display the pageskin and NOT use single request mode`, async ({
+		test.skip(`Test pageskin front on ${breakpoint} should NOT display the pageskin and NOT use single request mode`, async ({
```
arelra, line 60: Could we just skip tablet if that's the problem one. No worries if just simpler to skip both

arelra (approved):

playwright/tests/parallel-4/pageskin.spec.ts
domlander: Yep, absolutely. I thought it was both tablet and mobile but you are correct. I will return early with that comment above for tablet

PR #1099, opened 2023-10-12
Fix liveblog Playwright tests

emma-imber:
## What does this change?

- Makes the liveblog-live-update tests run in order instead of concurrently, as this resolves the flakiness
- Adds an extra console log line to the liveblog-live-update test for easier local testing and debugging
- Forces the code to wait until newly inserted blocks are visible before counting them in the liveblog-ad-limit test
- Separates out the liveblog-ad-limit tests into individual tests on a shared web page context, so that we're only testing one thing at a time instead of having three test conditions in one test

## Why?
These changes improve the flakiness in the liveblog-live-update and liveblog-ad-limit tests.

## How reliable are the tests now?

I ran these tests a large number of times locally to get these percentage pass rates, to give us a rough idea of how flaky the tests are:

Liveblog-live-update test: 98% pass rate
Liveblog-ad-limit test: 84% pass rate

So these tests aren't perfect, but they are definitely much more reliable than they were before.

### DIFF
--- a/playwright/tests/parallel-3/liveblog-ad-limit.spec.ts
+++ b/playwright/tests/parallel-3/liveblog-ad-limit.spec.ts
@@ -14,9 +14,6 @@ const desktopBreakpoint = breakpoints.filter(
 
 const maxAdSlots = 8;
 
-const countInlineSlots = (page: Page) =>
-	page.locator('#liveblog-body .ad-slot--liveblog-inline').count();
-
 const addAndAwaitNewBlocks = async (page: Page, blockContent: string) => {
 	// scroll to the top so we get a toast to click on
 	await page.evaluate(() => window.scrollTo(0, 0));
@@ -45,6 +42,9 @@ const addAndAwaitNewBlocks = async (page: Page, blockContent: string) => {
 	const newBlock = page.getByText(blockContent).first();
 	await expect(newBlock).toBeVisible();
 };
+
+const countInlineSlots = (page: Page) =>
+	page.locator('#liveblog-body .ad-slot--liveblog-inline').count();
 
 test.describe.serial('Ad slot limits', () => {
 	pages.forEach(({ path, expectedMinInlineSlotsOnDesktop }) => {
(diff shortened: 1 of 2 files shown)


playwright/tests/parallel-3/liveblog-ad-limit.spec.ts
```diff
@@ -12,70 +12,89 @@ const desktopBreakpoint = breakpoints.filter(
...
+	// Here, we force the test to wait until the blocks are visible on the page before counting the ad slots
+	const newBlock = page.getByText(blockContent).first();
+	await expect(newBlock).toBeVisible();
 };
 
 const countInlineSlots = (page: Page) =>
 	page.locator('#liveblog-body .ad-slot--liveblog-inline').count();
 
-test.describe('Ad slot limits', () => {
+test.describe.serial('Ad slot limits', () => {
```
arelra, line 48: Nice detective work @emma-imber ! I'm surprised that we needed to add serial as each test should already be isolated ?: https://playwright.dev/docs/browser-contexts

emma-imber: Yeah I thought the same! But when I remove it the tests don't pass. Not 100% sure what's going on but it seems we need it for the tests to work

playwright/tests/parallel-3/liveblog-live-update.spec.ts
```diff
@@ -12,10 +12,10 @@ import { loadPage } from '../../lib/load-page';
```
domlander, line 12: Can the comment on lines 7-11 be removed now?

emma-imber: Good spot, now deleted!

arelra (approved):
Great another test fixed!

I'd like to see if we can run in parallel eventually so could we add a TODO and/or a ticket in the backlog to investigate that separately pls? Thanks.

emma-imber:
Just adding a note here that I've been testing this pretty thoroughly locally today and the tests still aren't completely reliable.  I've added details into the PR description, they now have pass rates of 98% and 84%. Before this refactor the pass rates at a rough guess were below 50%. While not perfect, I'm happy that this is enough of an improvement to merge these changes.

PR #1100, opened 2023-10-13
Make eager prebid a zero percent test

emma-imber:
## What does this change?
Sets the eager prebid test to 0%

## Why?
Depending on the outcome of the test, we may want to test again with different numbers or keep the logic in the code, so it makes sense to preserve it behind a 0% test instead of deleting it while we wait for the test results.


arelra (approved):

dskamiotis (commented):
This looks good to me as well as the minor update to commercial 👍

PR #1118, opened 2023-10-24
Unit tests for shouldLazyLoad

emma-imber:
## What does this change?
- Moves the existing two tests for shouldLazyLoad to a new file
- Adds a mocked version for getCurrentBreakpoint to test the behaviour of shouldLazyLoad at different breakpoints
- Adds a test to cover when `dll = 1`

## Why?
We had an incident earlier this year where shouldLazyLoad was not returning the value we expected on mobile view. We wanted to add more rigorous unit testing to help catch this before it becomes a problem in production.


arelra (approved):
Thanks for improving the coverage!

PR #1121, opened 2023-10-25
Add Ozone testgroup to Ad slot

deedeeh:
## What does this change?

This PR adds Ozone testgroup to the GAM request. We add it to ad slot object `DefineSlot` in order to get the same value of the testgroup in prebid. We create the testgroup in `DefineSlot` which is then stored in the Advert class. We read this value back in the `OzoneClientSideBidder` object. 

## Why?

We’d like to compare Ozone data with Ad Manager data which it wasn't the case before when we had just key-value in the Ozone prebid configuration but not in Google Ad Manager.

### DIFF
--- a/src/lib/dfp/Advert.ts
+++ b/src/lib/dfp/Advert.ts
@@ -100,7 +100,7 @@ class Advert {
 	hasPrebidSize = false;
 	headerBiddingBidRequest: Promise<unknown> | null = null;
 	lineItemId: number | null = null;
-	testgroup: string | undefined;
+	testgroup: string | undefined; //Ozone testgroup property
 
 	constructor(
 		adSlotNode: HTMLElement,
--- a/src/lib/dfp/Advert.ts
+++ b/src/lib/dfp/Advert.ts
@@ -119,7 +119,7 @@ class Advert {
 
 		this.slot = slotDefinition.slot;
 		this.whenSlotReady = slotDefinition.slotReady;
-		this.testgroup = slotDefinition.slot.getTargeting('testgroup');
+		this.testgroup = slotDefinition.slot.getTargeting('testgroup')[0];
 	}
 
 	/**
--- a/src/lib/dfp/Advert.ts
+++ b/src/lib/dfp/Advert.ts
@@ -100,6 +100,7 @@ class Advert {
 	hasPrebidSize = false;
 	headerBiddingBidRequest: Promise<unknown> | null = null;
 	lineItemId: number | null = null;
+	testgroup: string;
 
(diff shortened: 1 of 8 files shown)


src/lib/header-bidding/prebid/bid-config.ts
```diff
@@ -266,22 +267,27 @@ const ozoneClientSideBidder: (pageTargeting: PageTargeting) => PrebidBidder = (
...
-		placementId: getOzonePlacementId(sizes),
-		customData: [
-			{
-				settings: {},
-				targeting: {
-					// Assigns a random integer between 0 and 99
-					testgroup: String(Math.floor(100 * Math.random())),
-					...buildAppNexusTargetingObject(pageTargeting),
+	): PrebidOzoneParams => {
+		const advert = dfpEnv.adverts.find((ad) => ad.id === _slotId);
```
arelra, line 271: @chrislomaxjones depending on who merges first I think this will affect/be affected by #1080

arelra (approved):
Nicely done! ✨

src/lib/dfp/Advert.ts
```diff
@@ -100,6 +100,7 @@ class Advert {
 	hasPrebidSize = false;
 	headerBiddingBidRequest: Promise<unknown> | null = null;
 	lineItemId: number | null = null;
+	testgroup: string | undefined;
```
domlander, line 103: I think this might benefit from a comment to indicate that this is an Ozone testgroup. It isn't obvious that this property is fo

domlander (approved):
Congrats on your first commercial PR 🎉

src/lib/dfp/Advert.ts
deedeeh: I totally agree with you! Done 130c68de8204e39f925964f88b5b66601b58c446

src/lib/header-bidding/prebid/bid-config.ts
deedeeh: Who should merge first? @arelra @chrislomaxjones

deedeeh:
> Congrats on your first commercial PR 🎉

Thank you Dom 🎉🚀

src/lib/header-bidding/prebid/bid-config.ts
arelra: I think lets get this one in first @deedeeh as it is more urgent. Thanks!

deedeeh:
> Nicely done! ✨

Thank you Ravi for pairing 🙏🏻 It was very useful!

PR #1122, opened 2023-10-26
Handle undefined testgroups 

arelra:
## What does this change?

Handle undefined testgroups and add a changeset for #1121


deedeeh (approved):
Looks good 👍🏼

PR #1128, opened 2023-11-01
Added E2E tests for `fronts-banner` ads

domlander:
## What does this change?

- Added E2E tests for fronts-banner ads
- Since fronts pages may change their layout, this test doesn't intend to ensure the correct amount of `fronts-banner` ad slots are present, since the removal of collections may lead to fewer ad slots being inserted. Instead, this test merely ensures that the first `fronts-banner` ad slot is being inserted.

## Why?

- Testing coverage

### DIFF
--- a/playwright/lib/util.ts
+++ b/playwright/lib/util.ts
@@ -3,7 +3,7 @@ import type { UserFeaturesResponse } from '../../src/types/membership';
 
 type Stage = 'code' | 'prod' | 'dev';
 
-type ContentType = 'article' | 'liveblog' | 'front' | 'taggedFront';
+type ContentType = 'article' | 'liveblog' | 'front' | 'tagFront';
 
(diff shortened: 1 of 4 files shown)


playwright/lib/util.ts
```diff
@@ -27,41 +29,51 @@ const getHost = (stage?: Stage | undefined) => {
...
 
+const getDcrContentType = (
+	type: ContentType,
+): 'Article' | 'Front' | 'TagFront' => {
+	switch (type) {
+		case 'front':
+			return 'Front';
+
+		case 'taggedFront':
+			return 'TagFront';
```
emma-imber, line 40: Minor, but if the DCR type is TagFront should we be using _tagFront_ instead of _taggedFront_ in the ContentType type for consistency?

emma-imber (approved):

PR #1129, opened 2023-11-01
Roll out eager Prebid for desktop

chrislomaxjones:
## What does this change?

This PR removes the AB test and rolls out the new behaviour for eager Prebid, first implemented in this [this PR](https://github.com/guardian/commercial/pull/1057).

Note we're just rolling out the behaviour for Desktop breakpoints and above, and retaining the original non-eager behaviour for mobile/tablet. When we do run Prebid eagerly, we adjust the display margin to 10%, matching `variant-2` in the original AB test.

Hence the new behaviour is:
- `mobile` and `tablet`: A single intersection observer that fires when an ad slot is **20%** of the viewport away.
- `desktop` and `wide`: An intersection observer that fires when an ad slot is **10%** of the viewport away, and initiates a Prebid auction when ad slot is **50%** of the viewport away.

This is captured by a basic unit test, but we should also run an end-to-end test for this some time in the future.

## Why?

This test concluded with favourable results for desktop screen sizes, so we're rolling it out.


deedeeh (approved):
It is great that you added a new test for the new changes and like you said at some point adding e2e would be even better 👍🏼

Jakeii:
Good point about adding an e2e test @deedeeh, will followup!

PR #1130, opened 2023-11-02
removed references to 'user-features' module booting commercial bu…

dskamiotis:
## What does this change?
This pull request removes all references to the 'user-features' module from the project. The 'user-features' module was previously used for managing user-related data and features but is no longer needed in this codebase.

**Important Note**
Some functions from 'user-features' are still required for specific features like `adblock-ask.ts` and `commercial-features.ts`, which rely on functions like shouldHideSupportMessaging() and isAdFreeUser(). To accommodate this, a partial user-features.ts file has been retained for these functions.

This may not be a desirable long term solution but allows us to successfully test the user-features being set in one place (dotcom-rendering)
A discussion could be had regarding a centralised place for these functions eg. https://github.com/guardian/csnx/tree/main/libs/%40guardian

## Why?
The 'user-features' module is no longer needed in this codebase as its functionality has been ported over to the `dotcom-rendering` repository. This change prevents redundant setting and retrieval of cookies, as user state management is now handled by 'dotcom-rendering'.

This revised version maintains the essential information while being somewhat more concise and to the point. It still explains the purpose and the reason for the change.

## Testing
- The  [beta] @guardian/commercial label has been added to this PR, releasing a beta version of the bundle to NPM.
- Frontend deployed to CODE in https://github.com/guardian/frontend/pull/26673 to boot this version commercial.
- 

| **Cookies in browser** |
|:------------------|
|   Screenshot 2023-11-03 at 09 52 51     |

### DIFF
--- a/src/commercial.consentless.ts
+++ b/src/commercial.consentless.ts
@@ -7,10 +7,7 @@ import { initConsentless } from './lib/consentless/prepare-ootag';
 import { init as setAdTestCookie } from './lib/set-adtest-cookie';
 import { init as setAdTestInLabelsCookie } from './lib/set-adtest-in-labels-cookie';
 
-const bootConsentless = async (
-	consentState: ConsentState,
-	isDotcomRendering: boolean,
-): Promise<void> => {
+const bootConsentless = async (consentState: ConsentState): Promise<void> => {
 	const consentlessModuleList = [
 		setAdTestCookie(),
 		setAdTestInLabelsCookie(),
@@ -22,15 +19,6 @@ const bootConsentless = async (
 	];
 
 	//this is added so that we can load the subscriber cookie for DCR pages and correctly hide ads
-
-	if (isDotcomRendering) {
-		const userFeatures = await import(
-			/* webpackChunkName: "dcr" */
-			'lib/user-features'
-		);
-
-		consentlessModuleList.push(userFeatures.refresh());
-	}
 
 	await Promise.all(consentlessModuleList);
 
--- a/src/lib/experiments/ab.spec.ts
+++ b/src/lib/experiments/ab.spec.ts
@@ -22,11 +22,11 @@ const concurrentTests = _concurrentTests as typeof concurrentTestsMock;
 
 // This is required as loading these seems to cause an error locally (and in CI)
 // because of some implicit dependency evil that I haven't been able to figure out.
-jest.mock('lib/user-features', () => ({
-	getLastOneOffContributionDate: () => null,
-	isRecurringContributor: () => false,
-	shouldNotBeShownSupportMessaging: () => false,
-}));
+// jest.mock('lib/user-features', () => ({
+// 	getLastOneOffContributionDate: () => null,
+// 	isRecurringContributor: () => false,
+// 	shouldNotBeShownSupportMessaging: () => false,
+// }));
 
 function emptyFunction() {
 	// do nothing
--- a/src/lib/experiments/ab.spec.ts
+++ b/src/lib/experiments/ab.spec.ts
@@ -20,14 +20,6 @@ const { overwriteMvtCookie } = _;
 
 const concurrentTests = _concurrentTests as typeof concurrentTestsMock;
 
-// This is required as loading these seems to cause an error locally (and in CI)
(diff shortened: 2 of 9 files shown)


src/lib/user-features.ts
```diff
@@ -321,105 +96,35 @@ const isRecurringContributor = async (): Promise<boolean> =>
...
-	try {
-		purchaseInfo = JSON.parse(
-			decodeURIComponent(purchaseInfoRaw),
-		) as PurchaseInfo;
-	} catch {} // eslint-disable-line no-empty -- silently handle error
-
-	return purchaseInfo;
-};
-
 export {
```
domlander, line 107: Can we delete all these instead of commenting them out

src/commercial.consentless.ts
```diff
@@ -23,15 +20,6 @@ const bootConsentless = async (
 
 	//this is added so that we can load the subscriber cookie for DCR pages and correctly hide ads
```
domlander, line 21: Can this comment be deleted too?

src/standalone.commercial.ts
```diff
@@ -113,17 +112,6 @@ if (!commercialFeatures.adFree) {
  * Not sure if this is needed. Currently no separate chunk is created
```
domlander, line 112: This comment can also be deleted

src/lib/experiments/ab.spec.ts
```diff
@@ -22,11 +22,11 @@ const concurrentTests = _concurrentTests as typeof concurrentTestsMock;
 
 // This is required as loading these seems to cause an error locally (and in CI)
```
domlander, line 23: Can this function and comment be deleted instead of commented out?

src/lib/user-features.ts
```diff
@@ -39,174 +21,17 @@ const SUPPORT_RECURRING_CONTRIBUTOR_ANNUAL_COOKIE =
...
 
-/**
- * Does our _existing_ data say the user is a paying member?
- * This data may be stale; we do not wait for userFeatures.refresh()
- */
+// /**
+//  * Does our _existing_ data say the user is a paying member?
+//  * This data may be stale; we do not wait for userFeatures.refresh()
+//  */
 const isPayingMember = async (): Promise<boolean> =>
```
domlander, line 35: Can IsPayingMember be deleted? I only see it mocked it tests. I don't see anywhere where its functionality is used

src/lib/user-features.spec.ts
```diff
@@ -1,28 +1,26 @@
...
 } from './user-features';
 
+const isUserLoggedInOktaRefactor =
+	isUserLoggedInOktaRefactor_ as jest.MockedFunction<
+		typeof isUserLoggedInOktaRefactor_
+	>;
+
+jest.mock('lib/identity/api', () => ({
+	isUserLoggedInOktaRefactor: jest.fn(),
+	getAuthStatus: jest.fn(),
```
domlander, line 21: `getAuthStatus` no longer needs to be exported from  `src/lib/identity/api.ts` so we don't need to mock it any more

domlander (commented):
Nice work! My comments are mostly other things that look like they can be deleted

src/commercial.consentless.ts
dskamiotis: These were remnant comments from the original user-features and though it would be useful to keep track to explain the various functions?

src/lib/user-features.ts
dskamiotis: This is one of the few remaining functions which are needed in testing for `commercial-features.spec.ts` in `isPayingMember.mockResolvedValue(false);`. It can definitely be removed at a later date?

src/lib/user-features.spec.ts
dskamiotis: do you mean the whole mock? or just getAuthStatus?

dskamiotis: Also this getAuthStatus is being used in `src/core/targeting/build-page-targeting.spec.ts`

domlander:
> This PR still needed @dskamiotis?

I believe it is, but will need a few changes, i.e deleting user-features, and a rebase on main

arelra:
I think this can be closed @dskamiotis ?

PR #1133, opened 2023-11-08
Provide the full hash for prebid in dependencies

sndrs:
## What does this change?

add the full hash for the link to the prebid commit

## Why?

yarn 4 won't stand for the short hash


chrislomaxjones (approved):
Thanks for this.

PR #1135, opened 2023-11-08
New ad sizes for merchandising slots

SiAdcock:
## What does this change?

- adds MPU size to `merchandising-high` slot
- adds MPU and billboard sizes to `merchandising` slot
- enables header bidding for these slots and sizes

## Why?

Commercial have identified a number of Black Friday opportunities, one of which is to show ads in `merchandising-high` and `merchandising` slots on fronts and articles.

This change ensures that ads of Billboard and MPU sizes can be added to these slots.

There was an AB test added in guardian/commercial#954 to support showing Billboard ads in `merchandising-high`. This change extends the scope of this test to include new sizes and the `merchandising` slot. 

By extending the scope of the AB test, I thought it appropriate to update the name of the test. Corresponding name changes will also be needed in DCR and frontend.

This change will not be rolled out as an AB test, it will go live to all users. However I decided not to delete the AB test logic at this stage: by using the `#ab-AdsInMerch ` URL hash, Ad Ops may want to test the changes in PROD before we set it live.

## Related PRs

- guardian/frontend#26682
- guardian/dotcom-rendering#9454

### DIFF
--- a/src/lib/dfp/fill-static-advert-slots.ts
+++ b/src/lib/dfp/fill-static-advert-slots.ts
@@ -4,11 +4,11 @@ import { adSizes, createAdSize } from 'core/ad-sizes';
 import { commercialFeatures } from 'lib/commercial-features';
 import { getCurrentBreakpoint } from 'lib/detect/detect-breakpoint';
 import { removeDisabledSlots } from '../remove-slots';
+import { includeAdsInMerch } from './ads-in-merchandising-test';
 import { createAdvert } from './create-advert';
 import { dfpEnv } from './dfp-env';
 import { displayAds } from './display-ads';
 import { displayLazyAds } from './display-lazy-ads';
-import { includeBillboardsInMerchHigh } from './merchandising-high-test';
 import { setupPrebidOnce } from './prepare-prebid';
 import { queueAdvert } from './queue-advert';
 
@@ -22,7 +22,7 @@ const decideAdditionalSizes = (adSlot: HTMLElement): SizeMapping => {
 		};
 	} else if (
 		(name === 'merchandising-high' || name === 'merchandising') &&
-		includeBillboardsInMerchHigh()
+		includeAdsInMerch()
 	) {
 		return {
 			mobile: [adSizes.mpu],
--- a/src/lib/dfp/fill-static-advert-slots.ts
+++ b/src/lib/dfp/fill-static-advert-slots.ts
@@ -21,10 +21,11 @@ const decideAdditionalSizes = (adSlot: HTMLElement): SizeMapping => {
 			desktop: [adSizes.billboard, createAdSize(900, 250)],
 		};
 	} else if (
-		name === 'merchandising-high' &&
+		(name === 'merchandising-high' || name === 'merchandising') &&
 		includeBillboardsInMerchHigh()
 	) {
 		return {
+			mobile: [adSizes.mpu],
 			desktop: [adSizes.billboard],
 		};
 	} else if (contentType === 'LiveBlog' && name?.includes('inline')) {
--- a/src/lib/experiments/tests/ads-in-merch.ts
+++ b/src/lib/experiments/tests/ads-in-merch.ts
@@ -1,7 +1,7 @@
 import type { ABTest } from '@guardian/ab-core';
 
-export const billboardsInMerchHigh: ABTest = {
-	id: 'BillboardsInMerchHigh',
+export const adsInMerch: ABTest = {
+	id: 'AdsInMerch',
 	author: '@commercial-dev',
 	start: '2022-12-07',
 	expiry: '2023-11-30',
--- a/.changeset/warm-seas-grow.md
+++ b/.changeset/warm-seas-grow.md
@@ -1,0 +1,5 @@
+---
+"@guardian/commercial": minor
+---
+
+Show MPUs and Billboard ads in merchandising and merchandising-high slots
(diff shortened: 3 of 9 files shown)


src/lib/experiments/tests/ads-in-merch.ts
```diff
@@ -1,7 +1,7 @@
 import type { ABTest } from '@guardian/ab-core';
 
-export const billboardsInMerchHigh: ABTest = {
-	id: 'BillboardsInMerchHigh',
+export const adsInMerch: ABTest = {
+	id: 'AdsInMerch',
```
chrislomaxjones, line 4: You'll want to make sure to update the [switch in Frontend](https://github.com/guardian/frontend/blob/main/common/app/conf/switches/ABTestSwitches.scala#L58-L66) to match this new name.

chrislomaxjones (approved):
Looks good!

You'll want to update the test definitions in Frontend (+ the switch) and DCR to match this new one.

There may also be some extra styling required [in DCR](https://github.com/guardian/dotcom-rendering/pull/6720/files) (to match this PR for the merch-high), but since this is a zero-percent test it's not required before this goes out.

src/lib/experiments/tests/ads-in-merch.ts
```diff
@@ -1,7 +1,7 @@
...
 	author: '@commercial-dev',
 	start: '2022-12-07',
 	expiry: '2023-11-30',
```
chrislomaxjones, line 7: Are we happy with this for the expiry?

domlander:
> adds MPU size to merchandising-high slot
> adds MPU and billbaord sizes to merchandising slot

Are these temporary or permanent changes? Will we want this behaviour after Black Friday?

src/lib/dfp/fill-static-advert-slots.ts
```diff
@@ -21,10 +21,11 @@ const decideAdditionalSizes = (adSlot: HTMLElement): SizeMapping => {
 			desktop: [adSizes.billboard, createAdSize(900, 250)],
 		};
 	} else if (
-		name === 'merchandising-high' &&
-		includeBillboardsInMerchHigh()
+		(name === 'merchandising-high' || name === 'merchandising') &&
+		includeAdsInMerch()
 	) {
 		return {
+			mobile: [adSizes.mpu],
```
domlander, line 28: This will cause all devices mobile and larger to have the MPU size. To exclude MPUs from phablet and above, a new size mapping will need to be created for phablet in `ad-sizes.ts`.

chrislomaxjones: I hadn't really ever considered this before, do we do something like this for top-above-nav?

domlander: For top-above-nav, we define sizes for the tablet viewport in `ad-sizes.ts`. This means that if we do define additional sizes for top-above-nav at the mobile viewport anywhere, they won't be picked up by tablet or above.

SiAdcock: Oh yes thanks @domlander I can see now how my implementation is totally wrong! I don't have many gaps today, so let me have a quick go at this:

Currently, our `merchandising-high` ad sizes object looks like:

```js
{
  mobile: [
    adSizes.outOfPage,
    adSizes.empty,
    adSizes.merchandisingHigh,
    adSizes.fluid,
  }
}
```

I've updated it dynamically so that it now looks like this:

```js
{
  mobile: [adSizes.mpu],
  desktop: [adSizes.billboard]
}
```

Which is totally wrong 😄 We actually want it to look like this:

```js
{
  mobile: [
    adSizes.outOfPage,
    adSizes.empty,
    adSizes.merchandisingHigh,
    adSizes.fluid,
    adSizes.mpu
  },
  phablet: [
    adSizes.outOfPage,
    adSizes.empty,
    adSizes.merchandisingHigh,
    adSizes.fluid
  }
  desktop: [
    adSizes.outOfPage,
    adSizes.empty,
    adSizes.merchandisingHigh,
    adSizes.fluid,
    adSizes: billboard
  }
}
```

Does that sound about right?

src/lib/experiments/tests/ads-in-merch.ts
SiAdcock: I think this is okay. I fully intend to tear out all the AB test stuff before we go live, as we won't actually be running an AB test for Black Friday. It'll just be how merchandising slots work going forward.

src/lib/dfp/fill-static-advert-slots.ts
domlander: That last one looks right to me! Although thinking about it, `phablet` can probably display MPU's. `tablet` is maybe the smallest breakpoint that should not display MPUs.

The above can be achieved by adding the following to `ad-sizes.ts`. I think a comment is necessary to explain why we are duplicating the mapping from mobile to tablet.

```
'merchandising-high': {
		mobile: [
			...
		],
                 // The mobile mapping is given an additional size when filling the slot. We do not want this extra size applied to tablet, so we provide a mapping for tablet here.
                 tablet: [
			adSizes.outOfPage,
			adSizes.empty,
			adSizes.merchandisingHigh,
			adSizes.fluid,
		],
	},
```

I think that later when we remove the AB test, we will be able to specify the sizes in `ad-sizes.ts` directly, instead of having to add the sizes later in `fill-static-advert-slots.ts`.

SiAdcock: Makes perfect sense, thanks @domlander 🙏

PR #1136, opened 2023-11-08
Remove fronts banner test

domlander:
## What does this change?

Removes the logic checking whether the fronts banner test is running.

## Why?

The fronts banner test has been removed a while ago and this was missed.

### DIFF
--- a/src/lib/header-bidding/prebid/appnexus.ts
+++ b/src/lib/header-bidding/prebid/appnexus.ts
@@ -1,6 +1,5 @@
 import type { PageTargeting } from 'core/targeting/build-page-targeting';
 import { buildAppNexusTargetingObject } from 'lib/build-page-targeting';
-import { includeBillboardsInMerchHigh } from 'lib/dfp/merchandising-high-test';
 import { isInAuOrNz } from 'lib/utils/geo-utils';
 import type { HeaderBiddingSize } from '../prebid-types';
 import {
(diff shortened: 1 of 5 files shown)


src/lib/header-bidding/prebid/appnexus.ts
```diff
@@ -36,23 +35,18 @@ const getAppNexusInvCode = (sizes: HeaderBiddingSize[]): string | undefined => {
...
-): string => {
+const getAppNexusDirectPlacementId = (sizes: HeaderBiddingSize[]): string => {
 	if (isInAuOrNz()) {
 		return '11016434';
 	}
 
 	const defaultPlacementId = '9251752';
 	switch (getBreakpointKey()) {
 		case 'D':
-			if (isInFrontsBannerVariant || includeBillboardsInMerchHigh()) {
```
domlander, line 50: Note that `includeBillboardsInMerchHigh()` and `containsBillboardNotLeaderboard()` will always resolve to the same result. They are both true if and only if billboard is in `sizes`.

src/lib/header-bidding/prebid/bid-config.spec.ts
```diff
@@ -145,26 +142,11 @@ describe('getImprovePlacementId', () => {
...
 	test('should return -1 if no cases match', () => {
-		expect(getImprovePlacementId([createAdSize(1, 2)], false)).toBe(-1);
-	});
-
-	test('should give the expected values when in the fronts banner test in uk on desktop', () => {
-		isInUk.mockReturnValue(true);
-		getBreakpointKey.mockReturnValue('D');
-		containsMpuOrDmpu.mockReturnValue(false);
-		containsLeaderboardOrBillboard.mockReturnValueOnce(true);
-		containsBillboardNotLeaderboard.mockReturnValue(true);
```
emma-imber, line 160: I think it would be good to still have a test to cover the containsBillboardNotLeaderboard case for Improve - with these changes we don't test for the 22987847 return value any more

domlander: It would be good (but a proper chore!) to improve the readability of these tests at some point.

emma-imber: Completely agree! I'm very reliant on searching for specific values to navigate these tests

emma-imber (approved):

PR #1137, opened 2023-11-08
Adds helper function to find smallest size for slot

domlander:
## What does this change?

Adds a helper function that finds the smallest known (i.e. not fluid, fabric, or other ad size of varying width/height) ad size that can be selected to fill a slot.

## Why?

This is useful when attempting to minimise CLS. Our strategy in this regard is to reserve space on the page for the height of smallest ad size that can be inserted into a slot, then expand the slot if a larger ad size is selected.

Currently, when we find the smallest ad size for a particular slot, it is [hard-coded](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/components/ArticleContainer.tsx#L146), rather than derived from the list of possible ad-sizes. This means it is susceptible to be incorrect in the future if the list of ad-sizes for a slot changes. For example, the smallest ad size on mobile articles in the first inline slot is the Outstream size at 197px tall. If this ad size was removed, we could instead reserve space for the next smallest ad size, the MPU at 250px.

### DIFF
--- a/src/dfp/Advert.spec.ts
+++ b/src/dfp/Advert.spec.ts
@@ -1,6 +1,5 @@
 import type * as AdSizesType from 'core/ad-sizes';
 import { slotSizeMappings as slotSizeMappings_ } from 'core/ad-sizes';
-import type { Breakpoint } from 'core/lib/breakpoint';
 import { _, Advert, findSmallestAdHeightForSlot } from './Advert';
 
 const { getSlotSizeMapping } = _;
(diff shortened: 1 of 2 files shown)


src/lib/dfp/Advert.spec.ts
```diff
@@ -157,3 +158,32 @@ describe('getAdSizeMapping', () => {
...
+		['right', 'mobile', 250],
+		['right', 'tablet', 250],
+		['right', 'desktop', 250],
+	])(
+		'should find the smallest size for %s slot at %s breakpoint',
+		(slot, breakpoint, expected) => {
+			expect(
+				findSmallestAdHeightForSlot(
+					slot as AdSizesType.SlotName,
+					breakpoint as Breakpoint,
```
chrislomaxjones, line 184: Take or leave this one, but you could probably avoid asserting these here types here by asserting `as const` on the array of test cases.

chrislomaxjones (approved):
Neat!

PR #1138, opened 2023-11-09
Add Index site id for `320x50` size ads

chrislomaxjones:
## What does this change?

This PR adds a new `siteId` to be used with the mobile-sticky ads by Index. We add this site id to Index's bid params when we detect the slot supports the mobile-sticky size.

In order to accommodate this, I did a small amount of refactoring to break out the way we pick a site id into two separate functions.

Screenshot 2023-11-15 at 12 01 59

## Why?

So that Index can bid on this new inventory.

### DIFF
--- a/src/lib/header-bidding/prebid/bid-config.ts
+++ b/src/lib/header-bidding/prebid/bid-config.ts
@@ -117,11 +117,35 @@ const getTrustXAdUnitId = (
 	}
 };
 
-const getIndexSiteId = (): string => {
+/**
+ * We store a mapping of sections to Index site ids server-side, where each is
+ * split out by breakpoint. These are transferred to the client via the window,
+ * and read here
+ *
+ * This appears to be an old method of assigning site ids, with the newer method
+ * being to assign them according to ad size (@see getIndexSiteId)
+ */
+const getIndexSiteIdFromConfig = (): string => {
 	const site = window.guardian.config.page.pbIndexSites.find(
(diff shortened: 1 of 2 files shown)


src/lib/header-bidding/prebid/bid-config.ts
```diff
@@ -117,13 +117,37 @@ const getTrustXAdUnitId = (
...
+	// The only prebid compatible size for fronts-banner-ads and the merchandising-high is the billboard (970x250)
+	// This check is to distinguish from the top-above-nav slot, which includes a leaderboard
+	if (containsBillboardNotLeaderboard(slotSizes)) {
+		return '983842';
+	}
+
+	// Return a specific site id for the mobile sticky slot
+	if (containsMobileSticky(slotSizes)) {
+		return '1047869';
+	}
```
chrislomaxjones, line 145: This isn't super robust (it relies on the mobile sticky being the only slot that serves this size), but follows the existing patterns in this file. I think we might want to revisit how we assign placement-style ids in the future.

dskamiotis, line 145: Splitting into 2 separate functions seems sensible, clearly shows the logic of getting the site id. I suspect much easier to unpick and further ad to in future

dskamiotis (approved):
This seems like a handy piece of data for this scenario, tested locally and all passed 👍

PR #1144, opened 2023-11-16
Remove ads-in-merch AB test

SiAdcock:
## What does this change?

Removes the AB test that controls whether ads appear in merch slots

## Why?

Over Black Friday, all users will see this functionality. It will be controlled by the `adsInMerch` feature switch. 

## Related

- guardian/frontend#26693
- guardian/dotcom-rendering#9536


chrislomaxjones (approved):
Makes sense!

PR #1145, opened 2023-11-16
Unify centring slots in container to a single class

Jakeii:
## What does this change?
Add `ad-slot-container--centre-slot` class to replace the `top-above-nav-ad-rendered` attribute. Which will be removed later so this can release across DCR and Frontend gracefully.

Move the code out of the `render-advert-label` module as it's not to do with labels, it was just there to benefit from the check for fluid.

## Why?
So we can reuse the centring style semantically for other slots like merchandising.

Also changing to a class name, as this is just used to apply a css style.

This is to fix the centring issues when ads are displayed in merch slots ([details and screenshots in the Frontend PR](https://github.com/guardian/frontend/pull/26694))

### DIFF
--- a/src/lib/dfp/render-advert.ts
+++ b/src/lib/dfp/render-advert.ts
@@ -185,9 +185,35 @@ const renderAdvert = (
 					  })
 					: Promise.resolve();
 
+			const centredAdSlots = [
+				'dfp-ad--top-above-nav',
+				'dfp-ad--merchandising-high',
+				'dfp-ad--merchandising',
+			];
+
+			/* Center certain slots in their containers, this is class is added dynamically to avoid rendering quirks with the ad label, and variable width ads.
+			 */
+			const addContainerCenterClass = () => {
+				if (
+					isRendered &&
+					advert.node.parentElement?.classList.contains(
+						'ad-slot-container',
+					) &&
+					centredAdSlots.includes(advert.node.id)
+				) {
+					return fastdom.mutate(() => {
+						advert.node.parentElement?.classList.add(
+							'ad-slot-container--centred-slot',
+						);
+					});
+				}
+				return Promise.resolve();
+			};
+
 			return callSizeCallback()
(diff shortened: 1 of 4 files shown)


src/lib/dfp/render-advert.ts
```diff
@@ -146,6 +146,38 @@ const addContentClass = (adSlotNode: HTMLElement) => {
...
+				isRendered &&
+				!adSlotNode.classList.contains('ad-slot--fluid') &&
+				adSlotNode.parentElement?.classList.contains(
+					'ad-slot-container',
+				) &&
+				centreAdSlots.includes(adSlotNode.id)
+			) {
+				return true;
+			}
+			return false;
```
chrislomaxjones, line 168: The if statement is redundant here?
```suggestion
			return (
				isRendered &&
				!adSlotNode.classList.contains('ad-slot--fluid') &&
				adSlotNode.parentElement?.classList.contains(
					'ad-slot-container',
				) &&
				centreAdSlots.includes(adSlotNode.id)
			);
```

src/lib/dfp/render-advert.ts
```diff
@@ -146,6 +146,38 @@ const addContentClass = (adSlotNode: HTMLElement) => {
 	}
 };
 
+/* Centre certain slots in their containers, this is class is added dynamically to avoid rendering quirks with the ad label, and variable width ads. */
```
chrislomaxjones, line 149:
```suggestion
/* Centre certain slots in their containers, this class is added dynamically to avoid rendering quirks with the ad label, and variable width ads. */
```

chrislomaxjones (approved):
A nice elegant solution!

I know this is somewhat time-sensitive, so feel free to follow-up suggestions later.

src/lib/dfp/render-advert.ts
Jakeii: Nice spot, thanks 👍

PR #1150, opened 2023-11-20
Change order of adding ad centering/rendered classes

Jakeii:
## Why?
This should prevent a momentary flicker of the uncentered ad


chrislomaxjones (approved):
👍 Nice spot

PR #1152, opened 2023-11-21
Move `src/lib/*` to src

arelra:
## What does this change?

We currently have the following directory structure:
```
- src
  - core (lib type modules for export)
  - lib (containing most of the codebase)
    - dfp
    - messenger
    - ...
```
This PR updates the structure to the following:
```
- src
  - core (lib type modules for export)
  - dfp
  - messenger
  - ...
```

I've checked the build output and it is exactly the same.

## Why?

The current structure is a hangover from:

1. the migration of the commercial code from `frontend` to this repo `commercial`
2. the merging of `commercial-core` into `commercial`

Having all directories flattened under `src/` is more standard.

The directory structure however is still largely carried over from Frontend which we can improve but this is the first step in improving how the codebase is organised.


chrislomaxjones (approved):
This is a great improvement! Can't say I reviewed every file, but if all the checks have passed...

PR #1153, opened 2023-11-21
Exlcude US locale from ads in merch behaviour

SiAdcock:
## What does this change?

- Exclude the US locale from behaviour that allows ads to display in merch slots

## Why?

There's a possibility that US will need to do some work in GAM + training to ensure they can support ads in merch slots. If so, we would still like to roll out the behaviour in other regions. 

This change allows us to roll out to all regions except US.


SiAdcock:
I think this PR is ready for review, although we shouldn't merge until we get approval from @Amouzle / US stakeholders. A corresponding version bump of `@guardian/commercial` will also be needed in frontend.

chrislomaxjones (approved):
👍

ashishpuliyel:
We won't merge this. The US don't need an exclusion.

PR #1155, opened 2023-11-22
Enable inserting new inline ad slots when new blocks are created on liveblogs

domlander:
## What does this change?

- On liveblog pages where inline ads are inserted server-side, this enables the functionality to insert more ads as new blocks are pushed to the page. This functionality currently exists when inserting inline ads client-side, but we are moving to a server-side strategy.

## Why?

- So that we can replicate the functionality of client-side inserted ad slots, allowing to make the switch to server-side inserted inline ad slots

### DIFF
--- a/playwright/fixtures/pages/blogs.ts
+++ b/playwright/fixtures/pages/blogs.ts
@@ -15,11 +15,11 @@ const blogs: GuPage[] = [
 	{
 		path: getTestUrl({
 			stage,
-			path: '/football/live/2023/aug/08/carabao-cup-first-round-wrexham-v-wigan-gillingham-v-southampton-live',
+			path: '/business/live/2023/nov/22/sam-altman-openai-return-jeremy-hunt-autumn-statement-tax-cuts-business-live',
 			type: 'liveblog',
 		}),
 		name: 'ad-limit',
-		expectedMinInlineSlotsOnDesktop: 5,
+		expectedMinInlineSlotsOnDesktop: 4,
 	},
 	{
 		path: getTestUrl({
--- a/playwright/tests/parallel-3/liveblog-ad-limit.spec.ts
+++ b/playwright/tests/parallel-3/liveblog-ad-limit.spec.ts
@@ -5,19 +5,19 @@ import { breakpoints } from '../../fixtures/breakpoints';
 import { blogs } from '../../fixtures/pages';
 import { cmpAcceptAll } from '../../lib/cmp';
 import { loadPage } from '../../lib/load-page';
+import { countLiveblogInlineSlots } from '../../lib/util';
 
 /**
  * TODO serial e2e tests
  * - It would be good to see if these tests could be run in parallel in the future
  */
-
 const pages = blogs.filter(({ name }) => name === 'ad-limit');
 
 const desktopBreakpoint = breakpoints.filter(
 	({ breakpoint }) => breakpoint === 'desktop',
 )[0] as unknown as BreakpointSizes;
 
-const maxAdSlots = 8;
+const MAX_AD_SLOTS = 8;
 
 const addAndAwaitNewBlocks = async (page: Page, blockContent: string) => {
 	// scroll to the top so we get a toast to click on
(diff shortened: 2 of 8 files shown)


playwright/tests/parallel-3/liveblog-ad-limit.spec.ts
```diff
@@ -26,81 +26,73 @@ const addAndAwaitNewBlocks = async (page: Page, blockContent: string) => {
...
 	// Here, we force the test to wait until the blocks are visible on the page before counting the ad slots
 	const newBlock = page.getByText(blockContent).first();
 	await expect(newBlock).toBeVisible();
 };
 
-const countInlineSlots = (page: Page) =>
-	page.locator('#liveblog-body .ad-slot--liveblog-inline').count();
-
-test.describe.serial('Ad slot limits', () => {
+test.describe('A minimum amount of ad slots load', () => {
```
domlander, line 54: There are two main checks in this test:
1. when there is less than the maximum amount of ads and new content blocks are inserted, check that we don't insert ads past the limit
2. when there is currently the maximum amount of ads and new content blocks are inserted, check that we don't insert ads past the limit

which in theory should be two separate tests. However, I found that combining these into one test made the pass rate much more reliable. I didn't receive any "mockLiveUpdate not available errors" when there was one test rather than two

playwright/fixtures/pages/blogs.ts
```diff
@@ -15,11 +15,11 @@ const blogs: GuPage[] = [
 	{
 		path: getTestUrl({
 			stage,
-			path: '/football/live/2023/aug/08/carabao-cup-first-round-wrexham-v-wigan-gillingham-v-southampton-live',
+			path: '/business/live/2023/nov/22/sam-altman-openai-return-jeremy-hunt-autumn-statement-tax-cuts-business-live',
```
domlander, line 18: The previous page had too many ads now we're inserting slots server-side

src/spacefinder/liveblog-adverts.ts
```diff
@@ -31,34 +31,28 @@ const AD_SPACE_MULTIPLIER = 2;
```
domlander, line 31: Bear in mind the changes in this file are to allow inserting ads client side and server side both to work, depending on whether DCR inserts slots server-side or not. Once this PR is merged, the DCR PR to insert slots server-side will follow, then another PR to remove the client-side slot insertion functionality from this file will be removed.

src/spacefinder/liveblog-adverts.ts
```diff
@@ -175,23 +209,39 @@ export const init = (): Promise<void> => {
...
-		'variant';
-	if (isServerSideAdsMode) {
+	const isMobile = getCurrentBreakpoint() === 'mobile';
+
+	// Temporary measure to identify whether we're in server-side inline ad insertion mode.
+	// Client-side ad insertion functionality will be removed very soon.
+	const numMobileAdsInServerSideMode = document.querySelectorAll(
+		'.ad-slot-container.ad-slot-mobile',
+	).length;
+	IS_SERVER_SIDE_MODE = numMobileAdsInServerSideMode > 0;
```
chrislomaxjones, line 219: Can we pull this piece of decisioning out into its own function?

```typescript
// Temporary measure to identify whether we're in server-side inline ad insertion mode.
// Client-side ad insertion functionality will be removed very soon.
const isServerSideMode = () => {
	const numMobileAdsInServerSideMode = document.querySelectorAll(
		'.ad-slot-container.ad-slot-mobile',
	).length;
	return numMobileAdsInServerSideMode > 0;
};
```

src/spacefinder/liveblog-adverts.ts
```diff
@@ -31,34 +31,28 @@ const AD_SPACE_MULTIPLIER = 2;
 
 let AD_COUNTER = 0;
 let WINDOWHEIGHT: number;
+let IS_SERVER_SIDE_MODE = false;
```
chrislomaxjones, line 34: I'd usually avoid adding additional top-level mutable state, but I assume this will go away once we remove client-side initial insertion? In which case I think this is probably fine.

src/spacefinder/liveblog-adverts.ts
```diff
@@ -77,22 +71,26 @@ const getSpaceFillerRules = (
 };
 
 const getSlotName = (isMobile: boolean, slotCounter: number): string => {
-	if (isMobile && slotCounter === 0) {
-		return 'top-above-nav';
-	} else if (isMobile) {
-		return `inline${slotCounter}`;
+	if (isMobile) {
```
chrislomaxjones, line 74: Nice refactor!

src/spacefinder/liveblog-adverts.ts
```diff
@@ -147,24 +145,60 @@ const stopListening = () => {
 	document.removeEventListener('liveblog:blocks-updated', onUpdate);
 };
 
-const fill = (rules: SpacefinderRules) => {
-	const options: SpacefinderOptions = { pass: 'inline1' };
+const getPreviousContentBlock = (element: HTMLElement): HTMLElement | null => {
+	const prevElement = element.previousSibling;
+	if (prevElement === null) return null;
```
chrislomaxjones, line 150: This check would ensure `prevElement` isn't `null` and is a `HTMLElement`, making the assertions that follow unnecessary.

```suggestion
	if (!(prevElement instanceof HTMLElement)) return null;
```

src/spacefinder/liveblog-adverts.ts
```diff
@@ -147,24 +145,60 @@ const stopListening = () => {
...
+				`.js-liveblog-body > ${adContainerClass}`,
+			);
+			if (topAdvert === null) {
+				firstSlot = undefined;
+			} else {
+				const nearestContentBlock = getPreviousContentBlock(topAdvert);
+				firstSlot =
+					nearestContentBlock !== null
+						? nearestContentBlock
+						: undefined;
```
chrislomaxjones, line 179: Is this the same thing?

```typescript
firstSlot = nearestContentBlock ?? undefined;
```

src/spacefinder/liveblog-adverts.ts
```diff
@@ -147,24 +145,60 @@ const stopListening = () => {
...
+	return getPreviousContentBlock(prevElement as HTMLElement);
+};
+
+const insertMoreAds = () => {
+	const isMobile = getCurrentBreakpoint() === 'mobile';
+
+	let adContainerClass = '.ad-slot-container';
+	if (IS_SERVER_SIDE_MODE) {
+		adContainerClass += `.ad-slot-${isMobile ? 'mobile' : 'desktop'}`;
+	}
```
chrislomaxjones, line 165: Just checking, Are we saying in server-side mode the container class should be something like `.ad-slot-container.ad-slot-mobile`? This will successfully select the slot we require?

chrislomaxjones (commented):
Nice work, and thanks for looking at the tests! Had a few questions and suggestions 👇

src/spacefinder/liveblog-adverts.ts
domlander: Yeah agreed. This is indeed a temporary measure and will go away once we remove client-side initial insertion

src/spacefinder/liveblog-adverts.ts
domlander: Yep, that's correct. The containers exist for both sets of slots, whereas the slots will only exist for one set (mobile or not).

src/spacefinder/liveblog-adverts.ts
domlander: Ah yes nice spot

src/spacefinder/liveblog-adverts.ts
domlander: This check is only temporary, so I'd prefer to leave it as it is for now.

src/spacefinder/liveblog-adverts.ts
chrislomaxjones: Got it, thanks for confirming 🙂

src/spacefinder/liveblog-adverts.ts
chrislomaxjones: Sure thing!

chrislomaxjones (approved):
Nice one!

PR #1156, opened 2023-11-23
Upgrade Prebid to v8.24.0

Jakeii:
## What does this change?
Upgrade prebid to 8.24.0

## Why?
Mainly for security updates

### DIFF
--- a/.changeset/nervous-guests-dress.md
+++ b/.changeset/nervous-guests-dress.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
+Upgrade to prebid.js v8
--- a/jest.setupTestFrameworkScriptFile.js
+++ b/jest.setupTestFrameworkScriptFile.js
@@ -32,6 +32,13 @@ window.guardian = {
 	},
 };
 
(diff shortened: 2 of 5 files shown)


jest.setupTestFrameworkScriptFile.js
```diff
@@ -32,6 +32,13 @@ window.guardian = {
 	},
 };
 
+window.fetch = jest.fn().mockImplementation(() =>
```
Jakeii, line 35: fetch is undefined in jest tests, prebid has now switched to fetch, it isn't actually called in tests but it needs to be there because prebid binds to it [here](https://github.com/prebid/Prebid.js/blob/master/src/ajax.js#L5)

.changeset/nervous-guests-dress.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
+Upgrade to prebid.js v8
```
emma-imber, line 5: There are two changesets for this change - could we remove one of them?

Jakeii: oops 🙃

PR #1157, opened 2023-11-23
Created AB test and additional sizes for MPU when no epic

deedeeh:
## What does this change?

This introduces a test to show an MPU ad when reader revenue don't show an epic which appears at the end of Articles.  

- Created the AB test with id `MpuWhenNoEpic`
- Added the functionality which allows passing `additionalSizes` as an optional parameter to customEvents. 
-  Passed that as an argument to `fillDynamicAdSlot` which takes sizes as optional parameter. 

We are passing that additionalSizes to DCR as part of the customEvent after implementing few checks. 
For more details check https://github.com/guardian/frontend/pull/26711 and https://github.com/guardian/dotcom-rendering/pull/9636

image

## Why?

Instead of not showing anything at the end of an Article we want to test that showing an MPU would increase the revenue towards our Advertising.


src/core/ad-sizes.ts
```diff
@@ -388,7 +388,7 @@ const slotSizeMappings = {
 		phablet: [adSizes.outOfPage, adSizes.empty, adSizes.leaderboard],
 	},
 	'article-end': {
-		mobile: [adSizes.fluid],
+		mobile: [], // Mappings are dynamically added for this slot using additionalSizes
```
dskamiotis, line 391: very useful comment 👍

Jakeii:
> Just wondering, the `epic` encourages user to subscribe. So assuming it doesn't pop up as stated, does that mean the user is already subscribed? 

The epic has rules to determine whether it shows, for unsubbed users after it's shown a few times it stops for a bit.

>And therefore shouldn't see ads? They may be a case where this happens Im assuming? Like subscribed and `noAds` (GUAF1) false

We check for `renderAds` in the code which will be false if the user is ad free: https://github.com/guardian/dotcom-rendering/pull/9636/files#diff-2a3e6c846cfb3ce97cafc0e87288e340ce3ec187fe7933525c58c3948935b670R163
Which you can follow back to https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/layouts/StandardLayout.tsx#L342
and
https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/lib/canRenderAds.ts#L10

src/dfp/fill-static-advert-slots.ts
```diff
@@ -33,6 +36,18 @@ const decideAdditionalSizes = (adSlot: HTMLElement): SizeMapping => {
 			phablet: [adSizes.outstreamDesktop, adSizes.outstreamGoogleDesktop],
 			desktop: [adSizes.outstreamDesktop, adSizes.outstreamGoogleDesktop],
 		};
+	} else if (name === 'article-end' && isInUsa()) {
+		return {
+			mobile: [adSizes.fluid],
+		};
+	} else if (
```
domlander, line 43: We don't need the `else`'s since we are returning when the condition is true

deedeeh: Jake and I aren't following what you mean, could you please explain more?

src/dfp/fill-static-advert-slots.ts
```diff
@@ -33,6 +36,18 @@ const decideAdditionalSizes = (adSlot: HTMLElement): SizeMapping => {
...
+		name === 'article-end' &&
+		isInVariantSynchronous(mpuWhenNoEpic, 'variant') &&
+		isInUk()
+	) {
+		return {
+			desktop: [adSizes.outstreamDesktop, adSizes.outstreamGoogleDesktop],
+		};
```
domlander, line 50:
```suggestion
	} 
	 
	 if (name === 'article-end' && isInUsa()) {
		return {
			mobile: [adSizes.fluid],
		};
	} 
	
	if (
		name === 'article-end' &&
		isInVariantSynchronous(mpuWhenNoEpic, 'variant') &&
		isInUk()
	) {
		return {
			desktop: [adSizes.outstreamDesktop, adSizes.outstreamGoogleDesktop],
		};
```

src/dfp/fill-static-advert-slots.ts
domlander: I've added a suggestion below to show what I mean

src/dfp/fill-static-advert-slots.ts
deedeeh: Would that make a big difference in our code? I mean what will add to it?

domlander: The functionality is the same either way. On latest `main` (https://github.com/guardian/commercial/blob/main/src/dfp/fill-static-advert-slots.ts) the `else`'s aren't used, so my suggestion would be to keep things consistent

deedeeh: Got it! Sorry for the slow understanding 😅

src/core/ad-sizes.ts
deedeeh: Yes, it is otherwise it would be really confusing why we have an empty array!

src/header-bidding/slot-config.ts
```diff
@@ -198,6 +199,11 @@ const getSlots = (): HeaderBiddingSizeMapping => {
 			mobile: [adSizes.mpu],
 			desktop: [adSizes.billboard],
 		},
+		'article-end': {
+			mobile: isInUk() ? [adSizes.mpu] : [],
+			tablet: isInUk() ? [adSizes.mpu] : [],
+			desktop: isInUk() ? [adSizes.mpu] : [],
```
domlander, line 205: Are these two lines necessary?

domlander (approved):

src/dfp/fill-static-advert-slots.ts
domlander: It was very confusing. We were looking at two different versions of the same file!

deedeeh: Yes but I checked what we already have in main and resolved the conflicts based on that.

src/header-bidding/slot-config.ts
deedeeh: We had just mobile at the beginning but I believe in Prebid we need to have each breakpoint and when we added them it worked when we tested it.

PR #1162, opened 2023-11-29
Only apply the mostpop test at desired breakpoints

chrislomaxjones:
## What does this change?

Only apply the `mostpop` PlayWright test from tablet and above.

## Why?

We recently made a change that hide the `mostpop` slot until the tablet breakpoint (PR). This caused these tests in Commercial to (rightly) fail.


emma-imber (approved):
Good spot, thanks for fixing this!

PR #1164, opened 2023-11-29
Remove ads-in-merch test

domlander:
## What does this change?

- Remove ads-in-merch AB test
- Moves all the sizes for both merch slots to `ad-sizes.ts`

## Why?

- It has been decided that we will keep showing ads in merch slots, allowing ads to compete with merchandising for the slot.


SiAdcock (approved):
Awesome, thanks @domlander 👌

PR #1171, opened 2023-12-05
Launch Kargo to 99% of users

Jakeii:
## What does this change?
Extend the test created in #1005 to add Kargo as a prebid bidder in the US.

Invert the check for running kargo to be when *not* in the test.

## Why?

Wewe now want to launch the change for the majority of users and hold back 1% of our audience to compare revenue.


chrislomaxjones (approved):
Makes sense!

PR #1172, opened 2023-12-05
remove feature swtch logic for user-features module

dskamiotis:
## What does this change?
This PR aims to remove logic which will conditionally load the user-features 

## Why?
We moved `user-features` to `dcr`: https://github.com/guardian/dotcom-rendering/pull/9362.
As a transitional process, we kept user-features in commercial and added a feature-switch in order to switch back if any issues.
We are now confident it is working as expected in `dcr` and loading appropriately with all cookies.

### DIFF
--- a/src/commercial.consentless.ts
+++ b/src/commercial.consentless.ts
@@ -23,18 +23,6 @@ const bootConsentless = async (
 
 	//this is added so that we can load the subscriber cookie for DCR pages and correctly hide ads
 
-	if (
-		isDotcomRendering &&
-		window.guardian.config.switches.userFeaturesDcr !== true
-	) {
-		const userFeatures = await import(
-			/* webpackChunkName: "dcr" */
-			'lib/user-features'
-		);
-
-		consentlessModuleList.push(userFeatures.refresh());
-	}
-
 	await Promise.all(consentlessModuleList);
 
 	// Since we're in single-request mode
--- a/src/commercial.consentless.ts
+++ b/src/commercial.consentless.ts
@@ -7,10 +7,7 @@ import { initConsentless } from './consentless/prepare-ootag';
 import { init as setAdTestCookie } from './lib/set-adtest-cookie';
 import { init as setAdTestInLabelsCookie } from './lib/set-adtest-in-labels-cookie';
 
-const bootConsentless = async (
-	consentState: ConsentState,
-	isDotcomRendering: boolean,
-): Promise<void> => {
+const bootConsentless = async (consentState: ConsentState): Promise<void> => {
 	const consentlessModuleList = [
 		setAdTestCookie(),
 		setAdTestInLabelsCookie(),
@@ -20,8 +17,6 @@ const bootConsentless = async (
 		initArticleInline(),
 		initLiveblogInline(),
 	];
-
-	//this is added so that we can load the subscriber cookie for DCR pages and correctly hide ads
 
 	await Promise.all(consentlessModuleList);
 
--- a/src/standalone.commercial.ts
+++ b/src/standalone.commercial.ts
@@ -50,8 +50,7 @@ import { catchErrorsWithContext } from 'utils/robust';
 
 type Modules = Array<[`${string}-${string}`, () => Promise<unknown>]>;
 
-const { isDotcomRendering, frontendAssetsFullURL, switches, page } =
-	window.guardian.config;
+const { frontendAssetsFullURL, switches, page } = window.guardian.config;
 
 const decideAssetsPath = () => {
 	if (process.env.OVERRIDE_BUNDLE_PATH) {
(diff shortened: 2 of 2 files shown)


src/standalone.commercial.ts
```diff
@@ -113,21 +113,6 @@ if (!commercialFeatures.adFree) {
  * Not sure if this is needed. Currently no separate chunk is created
  * Introduced by @tomrf1
  */
```
SiAdcock, line 115: We can probably remove the above comment 👆

SiAdcock:
Is it possible to removing `user-features` entirely? I know, it appears to [break `ad-block-ask`](https://github.com/guardian/commercial/pull/1169/files#diff-7dddf70d729feb4fdbb6d87d6fc502f3d17bc809826283cf9f3448044e1e3a7f), but what is this module adding anyway?

From what I can tell, [DCR has its own ad-block-ask implementation](https://github.com/guardian/dotcom-rendering/blob/5e4a3d1e63d3ccb83364aa0effde0c7ebed83a3c/dotcom-rendering/src/components/TopRightAdSlot.tsx#L10) (AKA Shady Pie, which is currently disabled), so this module isn't used at all in DCR. Please correct me if I'm wrong 😊

So that leaves frontend-rendered pages. `ad-block-ask` looks for elements with a class of [`js-aside-slot-container`](https://github.com/guardian/commercial/blob/f1c651437004951184d2fb0928930056fc0c2f45/src/lib/adblock-ask.ts#L42C66-L42C66). This class gets added to the [`articleAsideSlot` template](https://github.com/guardian/frontend/blob/1ed566df6dc2e06be6eff764b97e20d8e37e4b36/article/app/views/fragments/articleAsideSlot.scala.html#L9) in frontend, which gets added to [a handful of pages](https://github.com/search?q=repo%3Aguardian%2Ffrontend%20articleAsideSlot&type=code). Of these pages, only [`photoGallery`](https://github.com/guardian/frontend/blob/1ed566df6dc2e06be6eff764b97e20d8e37e4b36/article/app/views/fragments/photoEssayBody.scala.html#L47) is still rendered by frontend, so by deleting `ad-block-ask`, we're only deleting the `ad-block-ask` from this particular content type. There doesn't appear to be anything deliberate about this. It doesn't seem worth the effort maintaining this code, exclusively to support ShadyPie in photo galleries while it is disabled everywhere else on the website.

So my feeling is to kill `user-features` and kill `ad-block-ask`. Question to everyone: is there anything I'm missing?

dskamiotis:
> Is it possible to removing `user-features` entirely? I know, it appears to [break `ad-block-ask`](https://github.com/guardian/commercial/pull/1169/files#diff-7dddf70d729feb4fdbb6d87d6fc502f3d17bc809826283cf9f3448044e1e3a7f), but what is this module adding anyway?
> 
> From what I can tell, [DCR has its own ad-block-ask implementation](https://github.com/guardian/dotcom-rendering/blob/5e4a3d1e63d3ccb83364aa0effde0c7ebed83a3c/dotcom-rendering/src/components/TopRightAdSlot.tsx#L10) (AKA Shady Pie, which is currently disabled), so this module isn't used at all in DCR. Please correct me if I'm wrong 😊
> 
> So that leaves frontend-rendered pages. `ad-block-ask` looks for elements with a class of [`js-aside-slot-container`](https://github.com/guardian/commercial/blob/f1c651437004951184d2fb0928930056fc0c2f45/src/lib/adblock-ask.ts#L42C66-L42C66). This class gets added to the [`articleAsideSlot` template](https://github.com/guardian/frontend/blob/1ed566df6dc2e06be6eff764b97e20d8e37e4b36/article/app/views/fragments/articleAsideSlot.scala.html#L9) in frontend, which gets added to [a handful of pages](https://github.com/search?q=repo%3Aguardian%2Ffrontend%20articleAsideSlot&type=code). Of these pages, only [`photoGallery`](https://github.com/guardian/frontend/blob/1ed566df6dc2e06be6eff764b97e20d8e37e4b36/article/app/views/fragments/photoEssayBody.scala.html#L47) is still rendered by frontend, so by deleting `ad-block-ask`, we're only deleting the `ad-block-ask` from this particular content type. There doesn't appear to be anything deliberate about the fact that only photo galleries support `ad-block-ask`. It doesn't seem worth the effort maintaining this code, exclusively to support `ad-block-ask` in photo galleries while it is disabled everywhere else on the website.
> 
> Removing `user-features` also appears to [break `commercial-features`](https://github.com/guardian/commercial/pull/1169/files#diff-74dd5c0e094c4fc27ec0f6eb48403d452b53542a2ff1e8471475efb3a4c3ac30). I wonder if we can remedy this by inlining `isAdFreeUser` into `commercial-features`?
> 
> So my feeling is to kill `user-features`, inline `isAdFreeUser` into `commercial-features` and kill `ad-block-ask`. Question to everyone: is there anything I'm missing?

This makes a lot of sense, I made a separate ticket for removing user-features from  code base, and we added (@dominickendrick ) the `adblock` module as an item to dev time agenda for reasons you stated above :)

src/commercial.consentless.ts
```diff
@@ -23,18 +23,6 @@ const bootConsentless = async (
 
 	//this is added so that we can load the subscriber cookie for DCR pages and correctly hide ads
```
domlander, line 24: This comment can be removed too

domlander:
There looks to be a linting error

PR #1175, opened 2023-12-07
Let playwright figure out sharding/parallelism

Jakeii:
## What does this change?
Use playwright [sharding](https://playwright.dev/docs/test-sharding) to automatically assign tests to the CI workers

## Why?
Micro improvement in run time and no need for the `parallel-{x}` directories


SiAdcock (approved):
Very neat 👍

PR #1179, opened 2023-12-08
Log when a slot comes back unfilled in playwright

Jakeii:
## What does this change?
Log when a slot comes back unfilled from GAM

## Why?
Useful to help find out if it was GAM, the test code or our code that resulted in the empty slot!


arelra (approved):
Will be handy! Thanks

PR #1181, opened 2023-12-08
More commercial metrics

Jakeii:
## What does this change?
Add a few more marks to commercial metrics
- viewable
- prebidStart

also track metrics for the first 2 fronts-banners

## Why?
`viewable` - don't know why we don't have this to be honest! Records the timestamp when the slot became viewable
`prebidStart` - because we may [separate running prebid for slots from fetching the ad](https://github.com/guardian/commercial/pull/1057), this could be useful for determining if prebid is running sufficiently in advance.

### DIFF
--- a/.changeset/dull-jeans-reply.md
+++ b/.changeset/dull-jeans-reply.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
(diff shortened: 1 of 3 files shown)


chrislomaxjones (approved):
Looks good 👍

PR #1182, opened 2023-12-08
Playwright stability tweaks

arelra:
## What does this change?

Some minor tweaks to improve Playwright test stability:

- add logging to help diagnose when ad slots are not fulfilled. This is an indication that we are not requesting an adtest and therefore requesting real ads from GAM which may not fulfill (h/t @Jakeii )

- remove the opt-out specific article and adtest from the articles list

- from the opt-out UI the adtest has been changed from `comdev` to `fixed-puppies-ci`. This makes it easier to ensure we are using the default adtest across gam and opt-out tests

- don' wait for the top-above-nav slot in the page-skin tablet and mobile tests as the pageskin adtest will not appply (> desktop) and so it goes out to regular GAM ads and may not fulfill

- perf tweak to only install Chromium browsers on CI as that is all we test against

### DIFF
--- a/.github/workflows/e2e-playwright.yml
+++ b/.github/workflows/e2e-playwright.yml
@@ -56,7 +56,7 @@ jobs:
             ghcr.io/guardian/dotcom-rendering:main
 
       - name: Install Playwright Browsers
-        run: yarn playwright install --with-deps
+        run: yarn playwright install --with-deps chromium
         working-directory: ./commercial
 
       - name: Start Commercial server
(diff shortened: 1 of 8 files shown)


Jakeii (approved):
Awesome! Should we also subscribe to the commercial logger for the tests?

arelra:
> Awesome! Should we also subscribe to the commercial logger for the tests?

Ah yes I forgot to add that. I'll do as a follow up.

PR #1190, opened 2023-12-14
Enable Prebid for mobile sticky web in ROW

deedeeh:
## What does this change?

This PR enables Prebid for mobile-sticky in ROW and add the correct placement IDs for each Prebid vendor. We also refactored the unit tests and added few more to test mobile-sticky in Ozone, Ozone US, Improve, OpenX. 

In the future we should add more tests for Ozone and AppNexus.

## Why?

A follow up work #1187

### DIFF
--- a/src/header-bidding/prebid/appnexus.ts
+++ b/src/header-bidding/prebid/appnexus.ts
@@ -42,8 +42,7 @@ const getAppNexusDirectPlacementId = (sizes: HeaderBiddingSize[]): string => {
 	}
 
 	if (isInRow() && containsMobileSticky(sizes)) {
-			return '31512573';
-		}
+		return '31512573';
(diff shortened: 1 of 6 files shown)


src/header-bidding/prebid/appnexus.ts
```diff
@@ -40,6 +41,12 @@ const getAppNexusDirectPlacementId = (sizes: HeaderBiddingSize[]): string => {
 		return '11016434';
 	}
 
+	if (isInRow()) {
+		if (containsMobileSticky(sizes)) {
```
domlander, line 45:
```suggestion
	if (isInRow() && containsMobileSticky(sizes)) {
```

domlander (approved):

PR #1192, opened 2023-12-14
Remove script that creates inline ad slots on liveblogs

domlander:
## What does this change?

- Removes script that creates inline ad slots on liveblogs
- Removes disabled ad slots from the DOM before filling static slots

## Why?

- Inserting inline ad slots on liveblog pages is now performed server-side by DCR
- Don't fill disabled ad slots. Similar behaviour exists for GAM ads

## Screenshots

The extra ad container (top one of the three in _Before_) is removed.

| Before      | After      |
| ----------- | ---------- |
| image | image |

### Before

We're currently sending requests for two-sets of ad slots: one is hidden on the page, one is displayed. There is some commonality between the ad ID's in these two sets of slots, so I think there is a race condition where for two slots of the same ID, only one will be filled with an ad. This could be the visible slot or the hidden slot. The following screenshot are all taken from the same page and illustrate the current behaviour.

image
image
image


src/consentless/dynamic/liveblog-inline.ts
```diff
@@ -1,165 +0,0 @@
...
-		if (AD_COUNTER < MAX_ADS) {
-			const el = document.querySelector(
-				`${rules.bodySelector} > .ad-slot`,
-			);
-			if (el && el.previousSibling instanceof HTMLElement) {
-				firstSlot = el.previousSibling;
-			} else {
-				firstSlot = undefined;
-			}
-			startListening();
```
SiAdcock, line 141: Minor (probably lack of knowledge) question: can we retain the functionality here that adds slots when new blocks are added, just so we don't lose it? Then migrate it to DCR as a health ticket at some point?

domlander: This is a task that still need to be done. It involves some refactoring to separate the code that inserts new blocks from the code that makes calls to GAM for new ads. I'll create a ticket for this follow-up work.

SiAdcock (approved):
Tidy work @domlander 💎

PR #1194, opened 2023-12-15
collapse merchandise ad slots on viewport resize

dskamiotis:
## What does this change?
Currently when a user changes the viewport size from desktop to tablet or phone size, the merch slots still contain the `billboard` ad size which gets cut off.

This PR should collapse the `merchandise` and `merchandise-high` slot when the viewport is resized and therefore if billboard ad no longer fits in viewport.

This PR adds an callback which adds a class to `dfp-ad--merchandising` element for thje `merchandising` and `merchandising-high` ad slots.
This class is conditionally added on the premise that the adSize is `billboard` (970x250)

As a result of testing in CODE we discovered that some ads were larger than `billboard` ad like 1300x250.
As a result the aproach was to include non `fluid` ads as well.
## Why?
This is essential so that code in `dotcom-rendering` can detect this class and in turn apply appropriate `adSlotCollapseStyles`. See PR: https://github.com/guardian/dotcom-rendering/pull/9930

Tested in CODE with beta release:

Merchandising slot
Screenshot 2023-12-14 at 17 32 29

merchandising-high slot
Screenshot 2023-12-14 at 17 41 42

| Before     | After      |
| ----------- | ---------- |
| Screenshot 2023-12-18 at 11 17 06 | Screenshot 2023-12-18 at 11 23 35|

### DIFF
--- a/src/dfp/render-advert.ts
+++ b/src/dfp/render-advert.ts
@@ -200,7 +200,7 @@ const renderAdvert = (
 						advert.node.dataset.name === 'merchandising-high'
 					) {
 						void advert.updateExtraSlotClasses(
-							'ad-slot--merchandising-hide-below-desktop',
+							'ad-slot--collapse-below-desktop',
 						);
 					}
 					const sizeCallback = sizeCallbacks[advert.size.toString()];
--- a/src/dfp/render-advert.ts
+++ b/src/dfp/render-advert.ts
@@ -95,7 +95,9 @@ sizeCallbacks[adSizes.billboard.toString()] = (advert: Advert) => {
 		advert.node.dataset.name === 'merchandising' ||
 		advert.node.dataset.name === 'merchandising-high'
 	) {
-		advert.node.classList.add('ad-slot--merchandising-billboard');
+		return advert.updateExtraSlotClasses(
+			'ad-slot--merchandising-billboard',
+		);
 	}
 	return Promise.resolve();
 };
--- a/src/dfp/render-advert.ts
+++ b/src/dfp/render-advert.ts
@@ -90,18 +90,6 @@ sizeCallbacks[adSizes.outstreamMobile.toString()] = (advert: Advert) =>
 sizeCallbacks[adSizes.googleCard.toString()] = (advert: Advert) =>
 	advert.updateExtraSlotClasses('ad-slot--gc');
 
-sizeCallbacks[adSizes.billboard.toString()] = (advert: Advert) => {
-	if (
-		advert.node.dataset.name === 'merchandising' ||
-		advert.node.dataset.name === 'merchandising-high'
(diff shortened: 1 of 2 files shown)


src/dfp/render-advert.ts
```diff
@@ -90,6 +90,16 @@ sizeCallbacks[adSizes.outstreamMobile.toString()] = (advert: Advert) =>
...
 	advert.updateExtraSlotClasses('ad-slot--gc');
 
+sizeCallbacks[adSizes.billboard.toString()] = (advert: Advert) => {
+	if (
+		advert.node.dataset.name === 'merchandising' ||
+		advert.node.dataset.name === 'merchandising-high'
+	) {
+		advert.node.classList.add('ad-slot--merchandising-billboard');
+	}
+	return Promise.resolve();
```
SiAdcock, line 100: Is it appropriate to instead use advert.updateExtraNodeClasses here, as used in the cases above? I can’t find any examples of extraNodeClasses, so I’m not sure if this part of the advert API is still useful.

dskamiotis: Had to add `await` in refactored code when using the above function, as its asynchronous. I believe its because of the if is awaiting to decide on the value before conditionally adding the class.

SiAdcock: You should be able to just return the promise from `updateExtraSlotClasses`

```js
	if (
		advert.node.dataset.name === 'merchandising' ||
		advert.node.dataset.name === 'merchandising-high'
	) {
		return advert.updateExtraSlotClasses('ad-slot--merchandising-billboard');
	}
	return Promise.resolve();
```

dskamiotis: That was my initial solution, but wound up in all sorts of type issues?
```
Unsafe member access .updateExtraSlotClasses on an `any` value.eslint[@typescript-eslint/no-unsafe-member-access](https://typescript-eslint.io/rules/no-unsafe-member-access)
```

src/dfp/render-advert.ts
```diff
@@ -87,8 +87,17 @@ sizeCallbacks[adSizes.outstreamGoogleDesktop.toString()] = (advert: Advert) =>
 sizeCallbacks[adSizes.outstreamMobile.toString()] = (advert: Advert) =>
 	advert.updateExtraSlotClasses('ad-slot--outstream');
 
-sizeCallbacks[adSizes.googleCard.toString()] = (advert: Advert) =>
-	advert.updateExtraSlotClasses('ad-slot--gc');
+sizeCallbacks[adSizes.googleCard.toString()] = (advert: Advert) => {
```
domlander, line 90: Is editing the `googleCard` callback intentional?

domlander (approved):

PR #1205, opened 2024-01-03
add class name to merch slot

dskamiotis:
## What does this change?
This is a simple PR to re-instate the revert : https://github.com/guardian/commercial/pull/1203/files

## Why?
As a result of some unwanted behaviour, merchandise as slot disappearing and therefore not presenting an MPU ad.


Jakeii:
We don't need the class name anymore right? Can close this PR?

PR #1208, opened 2024-01-08
Improve consentless first visit detection

Jakeii:
## What does this change?
Use the performance API to determine the type of navigation so that we can better deduce if this is the users first visit.

[The performance API's navigation events let us distinguish the navigation type
](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming/type#value)

Only if the referrer is empty and the navigation type is `navigate` can we assume this is the users first visit, otherwise they have refreshed or clicked the back/forwards button.

## Why?

So we can target ads without consent only on their "first visit" to the site.

### DIFF
--- a/src/core/targeting/build-page-targeting.spec.ts
+++ b/src/core/targeting/build-page-targeting.spec.ts
@@ -1063,7 +1063,7 @@ describe('Build Page Targeting', () => {
 		).toBe('f');
 	});
 
-	it('should set firstvisit to false referrer is the guardian and navigation type is not navigate', () => {
+	it('should set firstvisit to false if referrer is the guardian and navigation type is not navigate', () => {
 		jest.spyOn(window, 'performance', 'get').mockReturnValue({
 			getEntriesByType: () => [
 				{
--- a/src/core/targeting/build-page-targeting.spec.ts
+++ b/src/core/targeting/build-page-targeting.spec.ts
@@ -1090,6 +1090,33 @@ describe('Build Page Targeting', () => {
 		).toBe('f');
 	});
 
+	it('should set firstvisit to true if referrer is not the guardian and navigation type is navigate', () => {
+		jest.spyOn(window, 'performance', 'get').mockReturnValue({
+			getEntriesByType: () => [
+				{
+					type: 'navigate',
+				},
+			],
+			mark: () => {
+				//
+			},
+		} as unknown as Performance);
+		jest.spyOn(document, 'referrer', 'get').mockReturnValue(
+			'https://google.com/',
+		);
+		jest.spyOn(window, 'location', 'get').mockReturnValue({
+			hostname: 'theguardian.com',
+		} as unknown as Location);
+		expect(
+			buildPageTargeting({
+				adFree: false,
+				clientSideParticipations: {},
+				consentState: emptyConsent,
+				isSignedIn: true,
+			}).firstvisit,
+		).toBe('t');
+	});
+
 	it('should not set firstvisit if consent is allowed', () => {
 		jest.spyOn(document, 'referrer', 'get').mockReturnValue('');
 		jest.spyOn(window, 'location', 'get').mockReturnValue({
--- a/src/core/targeting/build-page-targeting.spec.ts
+++ b/src/core/targeting/build-page-targeting.spec.ts
@@ -968,7 +968,10 @@ describe('Build Page Targeting', () => {
 		});
 	});
 
-	it('should set firstvisit to true if this is the users first visit to the page', () => {
+	it('should set firstvisit to true if referrer is empty and navigation api is missing', () => {
+		jest.spyOn(window, 'performance', 'get').mockReturnValue(
+			undefined as unknown as Performance,
+		);
 		jest.spyOn(document, 'referrer', 'get').mockReturnValue('');
 		jest.spyOn(window, 'location', 'get').mockReturnValue({
 			hostname: 'theguardian.com',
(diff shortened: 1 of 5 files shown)


arelra (commented):
> So we can target ads without consent only on their "first visit" to the site.

I might have misunderstood the intention of this PR but I thought we only call  `build-page-targeting.ts` once we have consent in `prepare-googletag`?

If so could we always set this to `'f'` for `build-page-targeting.ts` but then use this logic to add/override to the targeting object in `build-page-targeting-consentless.ts` ?

src/core/targeting/build-page-targeting.ts
```diff
@@ -70,13 +71,33 @@ const filterValues = (pageTargets: Record<string, unknown>) => {
...
+		return false;
+	}
+	const referrerUrl = new URL(referrer);
+	return referrerUrl.hostname !== window.location.hostname;
+};
+
 // A consentless friendly way of determining if this is the users first visit to the page
 const isFirstVisit = (referrer: string): boolean => {
-	if (!referrer) {
+	console.log('supportsPerformanceAPI', supportsPerformanceAPI());
```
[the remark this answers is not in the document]
Jakeii: Whoops, they were just for debugging

Jakeii:
> > So we can target ads without consent only on their "first visit" to the site.
> 
> I might have misunderstood the intention of this PR but I thought we only call `build-page-targeting.ts` once we have consent in `prepare-googletag`?
> 
> If so could we always set this to `'f'` for `build-page-targeting.ts` but then use this logic to add/override to the targeting object in `build-page-targeting-consentless.ts` ?

No build-page-targeting is used for both, but we have an [allowlist](https://github.com/guardian/commercial/blob/main/src/core/targeting/build-page-targeting-consentless.ts#L5) of keys that are sent to opt-out.

When @domlander and I [initially added this](https://github.com/guardian/commercial/pull/1174) I *think* it was that we didn't want to send a key to GAM that isn't used, but also keep all the targeting generation in `build-page-targeting.ts`

arelra (approved):

src/core/targeting/build-page-targeting.ts
```diff
@@ -70,13 +71,32 @@ const filterValues = (pageTargets: Record<string, unknown>) => {
...
+	const lastNavigationEvent = navigationEvents[navigationEvents.length - 1];
+	return lastNavigationEvent?.type;
+};
+
+const referrerMatchesHost = (referrer: string): boolean => {
+	if (!referrer) {
+		return false;
+	}
+	const referrerUrl = new URL(referrer);
+	return referrerUrl.hostname !== window.location.hostname;
```
domlander, line 88:
```suggestion
	return referrerUrl.hostname === window.location.hostname;
```

src/core/targeting/build-page-targeting.spec.ts
```diff
@@ -983,7 +986,94 @@ describe('Build Page Targeting', () => {
...
+			buildPageTargeting({
+				adFree: false,
+				clientSideParticipations: {},
+				consentState: emptyConsent,
+				isSignedIn: true,
+			}).firstvisit,
+		).toBe('f');
+	});
+
+	it('should set firstvisit to false referrer is the guardian and navigation type is not navigate', () => {
```
domlander, line 1066:
```suggestion
	it('should set firstvisit to false if referrer is the guardian and navigation type is not navigate', () => {
```

src/core/targeting/build-page-targeting.ts
```diff
@@ -70,13 +71,32 @@ const filterValues = (pageTargets: Record<string, unknown>) => {
...
+};
+
 // A consentless friendly way of determining if this is the users first visit to the page
 const isFirstVisit = (referrer: string): boolean => {
-	if (!referrer) {
+	if (!supportsPerformanceAPI()) {
+		return referrer === '' && !referrerMatchesHost(referrer);
+	}
+	if (getLastNavigationType() === 'navigate' && referrer === '') {
```
domlander, line 96:
```suggestion
	if (getLastNavigationType() === 'navigate' && !referrerMatchesHost(referrer)) {
```

src/core/targeting/build-page-targeting.ts
```diff
@@ -70,13 +71,32 @@ const filterValues = (pageTargets: Record<string, unknown>) => {
...
+	}
+	const referrerUrl = new URL(referrer);
+	return referrerUrl.hostname !== window.location.hostname;
+};
+
 // A consentless friendly way of determining if this is the users first visit to the page
 const isFirstVisit = (referrer: string): boolean => {
-	if (!referrer) {
+	if (!supportsPerformanceAPI()) {
+		return referrer === '' && !referrerMatchesHost(referrer);
```
domlander, line 94:
```suggestion
		return !referrerMatchesHost(referrer);
```

Jakeii:
Cheers @domlander, I was wondering why the logic seemed so hard to get my head around!

src/core/targeting/build-page-targeting.ts
```diff
@@ -70,13 +71,35 @@ const filterValues = (pageTargets: Record<string, unknown>) => {
...
+const referrerMatchesHost = (referrer: string): boolean => {
+	if (!referrer) {
+		return false;
+	}
+	const referrerUrl = new URL(referrer);
+	return referrerUrl.hostname === window.location.hostname;
+};
+
 // A consentless friendly way of determining if this is the users first visit to the page
 const isFirstVisit = (referrer: string): boolean => {
```
domlander, line 92: I don't know if you think the following is better or worse, but I think it's functionally the same. Will leave it up to you.

const isFirstVisit = (referrer: string): boolean => {
if (supportsPerformanceAPI() && getLastNavigationType() !== 'navigate' ) {
		return false;
	}

return !referrerMatchesHost(referrer)
}

domlander (approved):

PR #1211, opened 2024-01-10
Use relative import path between core modules

Jakeii:
## What does this change?
Use relative import, not doing this has led to a strange failure in frontend https://github.com/guardian/frontend/actions/runs/7478681099/job/20354162665?pr=26808

### DIFF
--- a/.changeset/lovely-vans-help.md
+++ b/.changeset/lovely-vans-help.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': patch
+---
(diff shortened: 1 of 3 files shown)


domlander (approved):

SiAdcock:
Good catch. Any way we can avoid it cropping up in the future?

Jakeii:
It may actually because I removed "rootDir" from the tsconfig, I've also restored that now.

PR #1214, opened 2024-01-16
Move fill slot listener to standalone

Jakeii:
## What does this change?
Move the fill slot listener to be initialised in `standalone.commercial`

## Why?
We need the event listener to be added ASAP.

At the moment it's possible for the event to be sent before the listener has been setup.

The particular case where this is happening is the `article-end` slot. If the page is refreshed while it's in view, the slot is inserted after the fixed ad slots are initialised but before the fill-slot-listener has been setup, resulting in an empty slot.

Adding `window.googletag.cmd.push` will ensure the slot will only be filled if googletag loads.

### DIFF
--- a/src/init/consented/fill-slot-listener.ts
+++ b/src/init/consented/fill-slot-listener.ts
@@ -1,6 +1,6 @@
 import type { SizeMapping } from 'core';
-import { dfpEnv } from './dfp-env';
-import { fillDynamicAdSlot } from './fill-dynamic-advert-slot';
+import { dfpEnv } from '../../dfp/dfp-env';
+import { fillDynamicAdSlot } from '../../dfp/fill-dynamic-advert-slot';
 
 type ExternalSlotCustomEvent = CustomEvent<{
 	slotId: string;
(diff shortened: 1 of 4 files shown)


deedeeh (commented):
I reviewed the code in this PR and I have 2 questions:

- Why we want to initiate it in `standalone.commercial`?
- Does that change needs to be merged in another PR or something to see if it works?

I tried to test it to check if the `article-end` will appear or not by running Commercial and DCR locally. Am I missing Frontend or there is something else to see it working?

Jakeii:
> I reviewed the code in this PR and I have 2 questions:
> 
> * Why we want to initiate it in `standalone.commercial`?

We want to register the listener as soon as possible, and this is the earliest stage of the commercial code

> * Does that change needs to be merged in another PR or something to see if it works?
> 
> I tried to test it to check if the `article-end` will appear or not by running Commercial and DCR locally. Am I missing Frontend or there is something else to see it working?

You need to also opt into the ab test and use the teads ad test `?adtest=teads#ab-MpuWhenNoEpic=variant`

Jakeii:
Nice spot @deedeeh that CSS is in DCR so will required another PR there to address

deedeeh:
> Nice spot @deedeeh that CSS is in DCR so will required another PR there to address

Does the DCR one should be merged before commercial or that won't make a difference?

src/init/consented/fill-slot-listener.ts
```diff
@@ -27,20 +27,28 @@ const isCustomEvent = (event: Event): event is CustomEvent => {
  * These events will not be received from a restricted iframe such, such as a
  * cross-origin or safeframe iframe.
  */
-export const createSlotFillListener = () => {
+const createSlotFillListener = () => {
 	document.addEventListener('gu.commercial.slot.fill', (event: Event) => {
-		if (isCustomEvent(event)) {
-			const { slotId, additionalSizes } = (<ExternalSlotCustomEvent>event)
-				.detail;
+		window.googletag.cmd.push(() => {
```
deedeeh, line 32: What if something went wrong with googletag what will happen then? I understand that the slot won't be filled but what we will have instead?

deedeeh (approved):
This PR looks good and with the CLS change in DCR that should work as expected.

PR #1216, opened 2024-01-16
Add Ozone and AppNexus prebid unit tests for placementID

deedeeh:
## What does this change?

This PR adds more Ozone prebid unit tests and creates AppNexus prebid tests which was noted in a previous PR #1190.

All the added tests checks if `placementID` is correct based on region, breakpoint and ad size. 

## Why?

Covering more unit test cases.

### DIFF
--- a/src/header-bidding/prebid/appnexus.spec.ts
+++ b/src/header-bidding/prebid/appnexus.spec.ts
@@ -6,6 +6,7 @@ import {
 	isInUsa as isInUsa_,
 } from 'utils/geo-utils';
 import {
+	containsBillboard as containsBillboard_,
 	containsBillboardNotLeaderboard as containsBillboardNotLeaderboard_,
 	containsDmpu as containsDmpu_,
 	containsLeaderboard as containsLeaderboard_,
@@ -18,6 +19,7 @@ import {
 import { getAppNexusDirectPlacementId } from './appnexus';
 
 jest.mock('../utils');
+const containsBillboard = containsBillboard_ as jest.Mock;
 const containsBillboardNotLeaderboard =
 	containsBillboardNotLeaderboard_ as jest.Mock;
 const containsDmpu = containsDmpu_ as jest.Mock;
@@ -77,7 +79,7 @@ describe('getAppNexusDirectPlacementId', () => {
 		resetConfig();
 	});
 
-	test('should return correct placementID for any existing biding sizes in Australia or New Zealand', () => {
+	test('should return correct placementID for any existing bidding sizes in Australia or New Zealand', () => {
 		isInAuOrNz.mockReturnValue(true);
 		getBreakpointKey.mockReturnValue('D');
 		containsMpu.mockReturnValue(true);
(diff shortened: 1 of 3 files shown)


src/header-bidding/prebid/appnexus.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+	window.guardian.config.page.isDev = false;
+};
+
+describe('getAppNexusDirectPlacementId', () => {
+	afterEach(() => {
+		jest.resetAllMocks();
+		resetConfig();
+	});
+
+	test('should return correct placementID for any existing biding sizes in Australia or New Zealand', () => {
```
emma-imber, line 80: biding -> bidding

src/header-bidding/prebid/appnexus.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+		isInRow.mockReturnValue(true);
+		getBreakpointKey.mockReturnValue('D');
+		containsDmpu.mockReturnValue(true);
+		expect(getAppNexusDirectPlacementId([createAdSize(300, 600)])).toBe(
+			'9251752',
+		);
+	});
+
+	test('should return correct placementID for desktop billboard not leaderboad in UK, ROW and US', () => {
+		isInUk.mockReturnValue(true);
```
emma-imber, line 108: I think here we should just mock `isInUk` and **not** `isInRow` or `isInUsa`. At the moment in this test we're telling the check that we're simultaneously in the UK, ROW and USA, which isn't a valid test case. This comment applies to all the following checks too, where we mock 3 location values at once, we should only mock 1.

src/header-bidding/prebid/appnexus.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+	});
+
+	test('should return correct placementID for desktop leaderboard in UK, ROW and US', () => {
+		isInUk.mockReturnValue(true);
+		isInRow.mockReturnValue(true);
+		isInUsa.mockReturnValue(true);
+		getBreakpointKey.mockReturnValue('D');
+		containsLeaderboard.mockReturnValue(true);
+		expect(getAppNexusDirectPlacementId([createAdSize(728, 90)])).toBe(
+			'9251752',
```
emma-imber, line 153: I would expect this placementID to be 9926678. `containsLeaderboardOrBillboard` would be true if the ad is a leaderboard size, so it should be the same as the test above.

deedeeh: I get your point here because of the breakpoint desktop but when I run the tests it gives me the placementID for tablet for `containsLeaderboard` function. Is that because I use `containsLeaderboard`? There is something not quite right about it.

deedeeh: Based on our in person chat I added `containsBillboard` to be be false but it gives me the same error and it expects it to be`9251752` which doesn't make sense because this is a placementID for tablet. Not sure what to try next 😅

emma-imber: I would do something like this 
```
containsLeaderboard.mockReturnValue(true);
containsBillboard.mockReturnValue(false);
containsLeaderboardOrBillboard.mockReturnValue(
    containsBillboard() || containsLeaderboard(),
);
```
That should give you the correct placement ID 🙂

src/header-bidding/prebid/appnexus.spec.ts
deedeeh: Done in 12648f9372f9fb09dbe7a793a47d6c6bdbe748aa

src/header-bidding/prebid/appnexus.spec.ts
deedeeh: Ok after you explanation it totally makes sense! Done in 12648f9372f9fb09dbe7a793a47d6c6bdbe748aa

src/header-bidding/prebid/appnexus.spec.ts
deedeeh: That actually worked. I wouldn't think about using `containsLeaderboardOrBillboard` function for this test but I understand now why it has to be used. Done in 12648f9372f9fb09dbe7a793a47d6c6bdbe748aa

emma-imber (approved):

PR #1218, opened 2024-01-17
Updates to the Spacefinder debugger

emma-imber:
## What does this change?
This PR makes a couple of small changes to the Spacefinder debugger:

- Renames `inline2` to `subsequent-inlines` to reflect that the button means all inlines after inline1, not just inline2
- Selects the first button in the Spacefinder control panel by default, so that we have the annotation appearing on the page and it doesn't appear blank
- Updates the documentation to reflect the fact that your URL only needs to contain `?sfdebug` - there's no difference between using `?sfdebug=1` or `?sfdebug=2` any more

## Why?
At the moment, when you load a page with `?sfdebug` in your URL, only the ad slots are highlighted by default, not the paragraphs, so the functionality of the debugger isn't immediately clear without clicking on a button on the control panel.

The name of inline2 was also misleading - it might make you think it only referred to the Spacefinder process for inline2 - not all inline adverts afterwards. Hopefully these small changes improve the user-friendliness of the the Spacefinder debugger.

### DIFF
--- a/docs/spacefinder/readme.md
+++ b/docs/spacefinder/readme.md
@@ -86,6 +86,6 @@ On desktop, the first inline ad sits inside the body of the article copy. All su
 
 ## Debugging
 
-Adding `sfdebug=1` to the URL shows information about `inline1` ads. Adding `sfdebug=2` shows information about the other inline slots.
+Adding `?sfdebug` to the URL opens the Spacefinder debugger control panel. The panel contains buttons to show the data for the different times that Spacefinder runs to find space for an ad. You can click these buttons to see the data for that run.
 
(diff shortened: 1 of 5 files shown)


arelra (approved):
Nice! I wonder if we could add an e2e test to check that the debugger is working? Could be a follow up.

emma-imber:
> Nice! I wonder if we could add an e2e test to check that the debugger is working? Could be a follow up.

Nice idea! I've added a super basic e2e test to check if the control panel loads when the URL parameter is added. Could always expand the test file if we think of more test cases in future 🙂

domlander (approved):

PR #1222, opened 2024-01-22
Update Prebid to resolve peer dependency mismatches with DCR

emma-imber:
## What does this change?
Updates Prebid to bring in the latest changes. This will resolve the peer dependency mismatches with dotcom-rendering.

This update has been tested on CODE to make sure that Prebid still runs as expected.


arelra (approved):
✨

PR #1225, opened 2024-01-24
Refresh page on consent change

emma-imber:
## What does this change?
Adds a module to refresh the page when consent state changes, and an E2E test to wait for a page reload when consent state changes.

Also deletes the module to remove consented ads on consent withdrawal, as we no longer need this if we're refreshing the page immediately.

## Why?
This means our data will correctly reflect the page consent status, and that users will immediately experience the appropriate ad experience on consent change.


Jakeii (approved):
Nice! Finally over the line!

src/init/shared/reload-page-on-consent-change.ts
```diff
@@ -0,0 +1,26 @@
...
+		if (initialConsentState === undefined) {
+			initialConsentState = consent;
+		}
+		if (initialConsentState.canTarget !== consent.canTarget) {
+			window.location.reload();
+		}
+	}, true);
+};
+
+const reloadPageOnConsentChange = once(() =>
```
arelra, line 22: Not sure why we need to once() as I thought we would only call this in the boot script?

emma-imber: Good spot - I think this is from the original PR back in frontend. I've removed it and seems to work fine without!

arelra (approved):
🥳 

ps @zekehuntergreen 😉

PR #1226, opened 2024-01-24
Replace `ophan-tracker-js` with `@guardian/ophan-tracker-js`

arelra:
## What does this change?

Replace `ophan-tracker-js` with `@guardian/ophan-tracker-js`

## Why?

`ophan-tracker-js` is deprecated and we should replace with `@guardian/ophan-tracker-js`.

v2.0.4 will also bring in this fix to how Ophan constructs its requests: https://github.com/guardian/ophan/pull/5624

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,10 +105,10 @@
 	},
 	"dependencies": {
 		"@changesets/cli": "^2.26.2",
+		"@guardian/ophan-tracker-js": "2.0.4",
 		"@octokit/core": "^4.0.5",
 		"fastdom": "^1.0.11",
 		"lodash-es": "^4.17.21",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -105,10 +105,10 @@
 	},
 	"dependencies": {
 		"@changesets/cli": "^2.26.2",
+		"@guardian/ophan-tracker-js": "2.0.4",
```
domlander, line 108: The carat has been removed. Was this change intentional?

arelra: Yes I think its good to fix the versions so we have predictable builds. Also makes it easier to keep deps aligned with our consumers. DCR has switched to fixed versions.

domlander (approved):

PR #1233, opened 2024-01-29
Update @guardian/consent-management-platform to 13.8.0

akinsola-guardian:
## What does this change?
Updating CMP to the latest version.
## Why?
To remain compliant.

To test this I created a beta and tested in dcr and frontend.


yarn.lock
```diff
@@ -6930,7 +6930,7 @@ prebid.js@guardian/prebid.js#91cabf5cdfc3c01745f4b10e55044f95a9d7d1b4:
     "@babel/plugin-transform-runtime" "^7.18.9"
     "@babel/preset-env" "^7.16.8"
     "@babel/runtime" "^7.18.9"
-    "@guardian/libs" "^16.0.0"
+    "@guardian/libs" "^15.6.4"
```
emma-imber, line 6933: This upgrade shouldn't be affecting the libs dependency in Prebid

akinsola-guardian: Yes, this is strange. 

I just updated the cmp version and ran `yarn start` which shouldn't affect the libs dependency in Prebid. How was this section generated? or was it manually added?

Jakeii (approved):
Thanks @akinsola-guardian!

PR #1234, opened 2024-01-29
Organise modules

Jakeii:
## What does this change?
Reorganises the "main" modules into directories according to their phase in the ad rendering process, outlined in this diagram:
commercial

Most of this is just moving modules around, there are a couple other changes that's part of this:
- messenger initialization moved to it's own module from `prepare-googletag.ts`
- grouped dynamic ad slot modules and they are now called via a new `dynamic-ad-slots.ts` module
- split standalone into the new entrypoint `commercial.ts` and moved consented module loading to `init/consented.ts` similar to `init/consentless.ts`

Still not 100% sure on the naming of the phases/directories open to ideas!

## Why?
The commercial code has evolved over time and hasn't had many "high-level" changes organisations to make the code more discoverable and easier to understand. This goes some way to helping with that.


arelra (approved):
Great to see this. Big step forwards on our re-architecture of the codebase! 🚀 

The new directories for reference:

Screenshot 2024-01-31 at 10 30 00

I wonder if we could lose some of the 'ad' and 'slot' post-fixes just to shorten the naming. For example:

`create`
`dynamic`
`render`

We should decide on what `ad` and `slot` mean too. I had thought previously that `slot` is the container that we control and `ad` is the ad iframe placed by googletag into the slot. But admittedly we use those terms interchangeably in conversation and in the code.

Also I think we've talked in the past about possibly aligning with the google tag phases such as `define` and `display` as outlined here:

https://github.com/guardian/commercial-playground/blob/main/simple-ad-example/render-ad.html

Worth talking over with the team to see what we converge on.

src/commercial.ts
```diff
@@ -0,0 +1,39 @@
...
+		consentState.tcfv2 &&
+		!getConsentFor('googletag', consentState) &&
+		!commercialFeatures.adFree
+	) {
+		void import(
+			/* webpackChunkName: "consentless" */
+			'./init/consentless'
+		).then(({ bootConsentless }) => bootConsentless(consentState));
+	} else {
+		bootCommercialWhenReady();
```
SiAdcock, line 27: Would it be worth code splitting the consented init, so we don't have to load it if there's no consent?

Jakeii: Looks like we decided to split it off for the AB test and never revisited https://github.com/guardian/frontend/pull/25682

SiAdcock: Ah cool, good detectiving! Maybe we can investigate whether this would be worth it as a separate ticket? Wouldn't want to delay merging this change

Jakeii: I think it's worth thinking a bit about this seperatly, I'll create a card to follow up

PR #1239, opened 2024-02-01
Show ads on mobile discussion

deedeeh:
## What does this change?

This PR is an OKR [ticket](https://trello.com/c/A5HUxCJN/2762-show-ads-on-mobile-discussion-dcr) to insert MPU ads in mobile discussion as the discussion readers are our most engaged readers. Other PRs linked to this one https://github.com/guardian/dotcom-rendering/pull/10448 and https://github.com/guardian/frontend/pull/26899

The first ad is shown after the 4th comment and then an ad is inserted after every 5 comments. We don't insert every 5 comments from the beginning because the number of comments is usually a multiple of 5, and showing an ad after the last comment doesn't look right from a UX point of view.

The code listens for the custom event `comments-loaded` from DCR and we insert `comments-expanded-${id}` ads.

The code also listens for the `comments-state-change` event which is fired when the user makes a change in filter or to the page number. When we receive the `comments-state-change` event we remove the previously inserted comments-expanded adverts. We make sure we are on mobile breakpoint so we don't remove the `comments-expanded` ads on the right column on desktop view.

This feature is behind a feature switch for now.

## Why?

Increase revenue and ad ratio without impacting the user experience. 

## Demo

https://github.com/guardian/commercial/assets/108270776/0d5ef5af-a32c-4bf2-8add-62037eca3424


arelra (approved):
📝

domlander:
It is possible that a `comments-expanded` ad can appear at the bottom of the comments section when it is expanded. This is fine in the below image as it sits above the Most Viewed container. Will the Most Viewed container always directly below the comments? Are there any situation where a different type of advert could be below the comments container?

image

Jakeii:
You may have already explained this, but what made you decide to add the slots via JS, rather than in the react code in DCR?

emma-imber:
> Not an issue, just something to think about:
> 
> It is possible that a `comments-expanded` ad can appear at the bottom of the comments section when it is expanded. This is fine in the below image as it sits above the Most Viewed container. Will the Most Viewed container always directly below the comments? Are there any situation where a different type of advert could be below the comments container?
> 
> image

The only case in DCR where the Discussion Component doesn't appear above the Most Viewed component is when the content is paid for. I may be wrong but I've never seen comments on paid for content?

src/insert/comments-expanded-advert.ts
```diff
@@ -43,6 +69,16 @@ const getRightColumn = (): HTMLElement => {
 	return rightColumn;
 };
 
+const getCommentsColumn = async (): Promise<HTMLElement> => {
+	return fastdom.measure(() => {
+		const commentsColumn: HTMLElement | null =
+			document.querySelector('.comments-column');
```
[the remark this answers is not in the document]
emma-imber: Yeah good point! We added the class name ourselves in DCR, we can add a comment to the code to explain to other developers why it was added to help discourage deletion. An e2e is test is a good idea - we'll have a look at implementing that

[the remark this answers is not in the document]
domlander: Would using a data attribute be more appropriate? We use data attributes for spacefinder: https://github.com/guardian/dotcom-rendering/blob/75379021a041660acee798cd3ed8902775bf14cf/dotcom-rendering/docs/contracts/001-commercial-selectors.md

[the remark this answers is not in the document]
emma-imber: I wonder if we should save writing an e2e test until we have the AB test results? We don't know for sure that this feature will be implemented yet. Would definitely be a valuable addition if we do implement the feature though!

src/insert/comments-expanded-advert.ts
```diff
@@ -67,6 +106,30 @@ const createResizeObserver = (rightColumnNode: HTMLElement) => {
 	resizeObserver.observe(rightColumnNode);
 };
 
+const removeMobileCommentsExpandedAds = (): Promise<void> => {
+	const currentBreakpoint = getBreakpoint(getViewport().width);
+	if (currentBreakpoint === 'mobile') {
```
domlander, line 111: Could we include tablet as well? We don't show a right column until the desktop breakpoint. Happy for this to be a follow-up ticket.

src/insert/comments-expanded-advert.ts
```diff
@@ -53,6 +89,9 @@ const isEnoughSpaceForAd = (rightColumnNode: HTMLElement): boolean => {
 	return rightColumnNode.offsetHeight >= minHeightToPlaceAd;
 };
 
+const isEnoughCommentsForAd = (commentsColumn: HTMLElement): boolean =>
+	commentsColumn.childElementCount > 5;
```
domlander, line 93: If we insert an advert after every fourth ad and it's acceptable for an advert to be at the bottom of the comments, should this be `commentsColumn.childElementCount >= 4;`?

src/insert/comments-expanded-advert.ts
emma-imber: The ticket just specified mobile for this work which is why we haven't implemented tablet. I think it makes more sense as a follow-up ticket

src/insert/comments-expanded-advert.ts
emma-imber: We've checked this with Alex and have decided that an advert should never be at the bottom of the comments. I've introduced a check to make sure that this doesn't happen.

PR #1263, opened 2024-02-12
Rename TagFront endpoint to TagPage

arelra:
## What does this change?

Rename TagFront endpoint to TagPage in e2e tests following https://github.com/guardian/dotcom-rendering/pull/10542

### DIFF
--- a/playwright/fixtures/pages/fronts.ts
+++ b/playwright/fixtures/pages/fronts.ts
@@ -5,7 +5,7 @@ type Front = GuPage & {
 	section: string;
 };
 
-type TagFront = GuPage;
+type TagPage = GuPage;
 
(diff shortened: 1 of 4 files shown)


domlander:
There's one more reference to tag fronts in this file on line 18: `playwright/tests/front-inline-slots.spec.ts`.

Does this also need a changeset?

arelra:
> There's one more reference to tagged fronts in this file on line 18: `playwright/tests/front-inline-slots.spec.ts`.
> 
> Does this also need a changeset?

Thanks! I don't think this needs a changeset as it isn't a change to commercial from a consumer point of view.

Jakeii (approved):

PR #1281, opened 2024-02-26
Changes to Prebid for comments-expanded ads

emma-imber:
## What does this change?
- Enables Prebid for desktop `comments-expanded` ad.
- Don't run Prebid for mobile `comments-expanded` ads - we only want MPU sizes in these slots and Prebid can send through ads that aren't the size we expect.
- Uses a minHeight on the mobile discussion ads ad slot instead of a fixed height on the container.

Note that the mobile `comments-expanded` ads are numbered in the same way as `inline` or `fronts-banner` ads, so by adding `comments-expanded` to slot-config, we're only targeting the desktop slots with the name `comments-expanded`, not the mobile slots that are called `comments-expanded-2` for example.

## Why?
Desktop `comments-expanded` ads have never had Prebid enabled on them. Hopefully by enabling it we can get some higher value ads in that slot. I've tested locally and it seems to be working as expected:

Screenshot 2024-02-27 at 15 20 29

Some third-party ads were coming through into the mobile `comments-expanded` slot that weren't the MPU size we expected. This caused display issues as we were setting a fixed height on the as slot container. See image for example of this incorrect behaviour. We've decided to switch off Prebid for these slots for now, as well as using a min-height instead of a fixed height.

### DIFF
--- a/.changeset/hungry-hounds-rule.md
+++ b/.changeset/hungry-hounds-rule.md
@@ -2,4 +2,4 @@
 '@guardian/commercial': minor
 ---
 
-Use minHeight instead of height for mobile discussion ads and don't run prebid for these slots
+Enable Prebid for desktop comments-expanded ads, and disable it for mobile comments-expanded ads.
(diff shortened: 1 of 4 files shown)


deedeeh (approved):
That looks good 👍🏼

Jakeii (approved):

PR #1289, opened 2024-03-01
Remove top-level container from DOM when removing fronts-banner ad

domlander:
## What does this change?

- Removes the top-level ad container from the DOM when removing a `fronts-banner` advert
- Adds a test

## Why?

- The top-level ad container has a minimum height on it to prevent layout shift, so we need to remove this too when removing a `fronts-banner` advert

### Before

https://github.com/guardian/commercial/assets/9574885/22e2dd0c-35dc-459d-9a16-c8a647172062

### After

https://github.com/guardian/commercial/assets/9574885/5df91746-db79-477c-bffb-15ae58ac6256

### DIFF
--- a/.changeset/spicy-rats-shout.md
+++ b/.changeset/spicy-rats-shout.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': patch
+---
+
+Remove top-level container from DOM when removing fronts-banner ad
--- a/src/events/empty-advert.spec.ts
+++ b/src/events/empty-advert.spec.ts
@@ -1,0 +1,43 @@
+import { _ } from './empty-advert';
+
+const { findElementToRemove } = _;
+
+const adSelector = '.js-ad-slot';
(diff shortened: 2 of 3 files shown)


Jakeii (approved):
Nice!

I was wondering recently why removing slots is necessary, I would have thought `display: none;` might suffice, as removing the slot from googletag and the dom makes debugging more difficult when we're not sure why an ad is not showing.

domlander:
> Nice!
> 
> I was wondering recently why removing slots is necessary, I would have thought `display: none;` might suffice, as removing the slot from googletag and the dom makes debugging more difficult when we're not sure why an ad is not showing.

I think that makes a lot of sense. We destroy the slot here and remove the advert from our internal `dfpEnv` object, so there should be no difference between hiding and removing the slot. Hiding would give us more debugging power as you say, so should be the better option. Maybe create a ticket to investigate this further?

PR #1296, opened 2024-03-06
Create advert density by section AB test

domlander:
## What does this change?

Fix a bug with spacefinder not use the expected `minAboveSlot` when avoiding other winners on mobile, it adds 250px (the height of an MPU). 

This was because spacefinder was using the same approach for floated right ads and inline ads. [The additions to the docs](https://github.com/guardian/commercial/blob/doml/section-ad-density-test/docs/spacefinder/readme.md#avoiding-other-winning-candidates) explains why different approaches are needed for each type of ad.

So as not to change the ad density everywhere, I've fixed the bug, and changed the margin to 750px for mobile and retained 500px for desktop. 

This PR also adds an AB test that will increase ad-density on articles in specific sections of the Guardian, e.g. business, money. For these test "high value sections" we will reduce it to `500px` on mobile.

And finally fixes a bug with spacefinder avoiding elements inserted by the spacefinder debugger

## Why?

- This AB test will be used to investigate the impact of increasing ad-ratio in specific sections


playwright/lib/util.ts
```diff
@@ -201,7 +201,6 @@ const countLiveblogInlineSlots = async (page: Page, isMobile: boolean) => {
 	return await page.locator(locator).count();
 };
 
-// @ts-expect-error -- used when logging uncommented
```
domlander, line 204: We get a warning on every PR with this unused function. Can we leave the `console.warn` in? It's only used for Playwright.

src/insert/spacefinder/article-body-adverts.ts
```diff
@@ -147,9 +151,10 @@ const addDesktopInline1 = (): Promise<boolean> => {
 		minAbove: isImmersive ? 700 : 300,
 		minBelow: 300,
 		opponentSelectorRules: {
+			// don't place ads right after a heading
```
domlander, line 154: This seemed a little contradictory before: We allow an ad >5px below a heading but we must keep >190px above a heading? We shouldn't put ads closely below heading as this is likely to break up semantically related content. I think this was perhaps a mistake given the confusing naming of `minAboveSlot `.

src/insert/spacefinder/article-body-adverts.ts
```diff
@@ -288,14 +292,16 @@ const addDesktopInline2PlusAds = (): Promise<boolean> => {
 };
 
 const addMobileInlineAds = (): Promise<boolean> => {
+	const minDistanceFromArticleTop = isInAdDensityVariant ? 100 : 200;
```
domlander, line 295: From testing on my real device (N=1), this ad is still below the fold by a fair way, so I think this should be OK.

src/insert/spacefinder/spacefinder.ts
```diff
@@ -28,26 +34,49 @@ type SpacefinderItem = {
 type SpacefinderRules = {
 	bodySelector: string;
 	body?: HTMLElement | Document;
-	// selector(s) for the elements that we want to allow inserting ads above
+	/**
+	 * Selector(s) for the elements that we want to allow inserting ads above
+	 */
 	candidateSelector: string | string[];
```
domlander, line 40: This is so that the VS Code feature of hovering over a property to see its description works.

Jakeii (approved):

PR #1297, opened 2024-03-06
Remove the `hr` element as a candidate from mobile spacefinder

Jakeii:
## What does this change?
Remove the `hr` element as a candidate from mobile spacefinder

## Why?
These elements have a large top margin (~50px), an ad would obviously be above this margin but spacefinder is unable to account for it, leading to ads being inserted a tad closer than we'd like to images if a `hr` follows an image. 

With the introduction of the `bypassMinBelow` rule it's unnecessary to have it be a candidate anyway.

Before:
Screenshot 2024-03-06 at 10 58 39

After:
Screenshot 2024-03-06 at 10 58 21

### DIFF
--- a/.changeset/tough-rice-enjoy.md
+++ b/.changeset/tough-rice-enjoy.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': patch
+---
+
+Remove HR as candidate for mobile spacefinder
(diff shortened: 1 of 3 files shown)


domlander:
Why is it unnecessary to have `hr` be a candidate with the introduction of the bypassMinBelow rule?

domlander (approved):

Jakeii:
> Why is it unnecessary to have `hr` be a candidate with the introduction of the bypassMinBelow rule?

Having hr's as candidates was part of throwing things at the wall to get ads working on list/ranked articles but it just turns out it's unnecessary!

PR #1300, opened 2024-03-12
Cmp libs

vlkerag:
## What does this change?
The cmp has moved to libs. Hence pointing to libs instead of cmp.


Jakeii (approved):
LGTM! Maybe add a link to a PR with some more information? https://github.com/guardian/consent-management-platform/pull/960

PR #1302, opened 2024-03-13
Add Playwright test for comments-expanded ads

emma-imber:
## What does this change?
Adds a Playwright test to check that the desktop and mobile versions of the `comments-expanded` ads are added when the comments are opened.

## Why?
In the comments on [this PR](https://github.com/guardian/commercial/pull/1239), I promised to add an E2E test for the `comments-expanded` ads when the new mobile ads were stable.

### DIFF
--- a/playwright/tests/comments-expanded.spec.ts
+++ b/playwright/tests/comments-expanded.spec.ts
@@ -1,0 +1,65 @@
+import { breakpoints } from '@guardian/source-foundations';
+import { expect, test } from '@playwright/test';
+import { articles } from '../fixtures/pages';
+import type { GuPage } from '../fixtures/pages/Page';
+import { cmpAcceptAll } from '../lib/cmp';
+import { loadPage } from '../lib/load-page';
+import { waitForIsland, waitForSlot } from '../lib/util';
+
+const { path } = articles[6] as unknown as GuPage;
+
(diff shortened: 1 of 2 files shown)


playwright/tests/comments-expanded.spec.ts
```diff
@@ -0,0 +1,65 @@
+import { breakpoints } from '@guardian/source-foundations';
+import { expect, test } from '@playwright/test';
+import { articles } from '../fixtures/pages';
+import type { GuPage } from '../fixtures/pages/Page';
+import { cmpAcceptAll } from '../lib/cmp';
+import { loadPage } from '../lib/load-page';
+import { waitForIsland, waitForSlot } from '../lib/util';
+
+const { path } = articles[6] as unknown as GuPage;
```
arelra, line 9: Is there only one page we use for this test? In which case we could create the path here rather than adding to `articles.ts`, which I think also increases the time taken for other tests as they would now run against that article too.

arelra (approved):
Nice thanks for coming back to this! ✨

playwright/tests/comments-expanded.spec.ts
emma-imber: Good shout, we only need it for this test. I'll define it just for this file

PR #1304, opened 2024-03-14
add 2x2 mobile sticky size

dskamiotis:
## What does this change?
We are adding the (2x2) size to the mobile-sticky slot.
This behaviour cannot be replicated in UK geo location, so using VPN to test with AUS.
## Why?
This will remove the ad-slot for the mobile-sticky when a Line Item with a 2x2 creative targets the slot.

## Testing

| mobile-sticky slot| googletag.console() |
|----------|----------|
| Screenshot 2024-03-14 at 15 50 39 | Screenshot 2024-03-14 at 15 42 25 |

### DIFF
--- a/src/core/ad-sizes.ts
+++ b/src/core/ad-sizes.ts
@@ -387,7 +387,7 @@ const slotSizeMappings = {
 		mobile: [adSizes.fluid],
 	},
 	'mobile-sticky': {
-		mobile: [adSizes.mobilesticky],
+		mobile: [adSizes.mobilesticky, adSizes.empty],
 	},
 	'crossword-banner-mobile': {
 		mobile: [adSizes.mobilesticky],


src/core/ad-sizes.ts
```diff
@@ -387,7 +387,7 @@ const slotSizeMappings = {
 		mobile: [adSizes.fluid],
 	},
 	'mobile-sticky': {
-		mobile: [adSizes.mobilesticky],
+		mobile: [adSizes.mobilesticky, adSizes.empty],
 	},
 	'crossword-banner-mobile': {
 		mobile: [adSizes.mobilesticky],
```
Jakeii, line 393: While you're at it do we need it here too? @Amouzle

domlander (approved):

src/core/ad-sizes.ts
domlander: We'll add this in a separate PR

PR #1310, opened 2024-03-20
Migrate from yarn to pnpm

i-hardy:
## What does this change?

This changes our package manager from yarn to pnpm

## Why?

pnpm offers several advantages over yarn including speed and better monorepo support, and brings us in line with DCR and CSNX, and offers the ability to better manage issues in our dependency tree (although we are intially somewhat brute forcing this due to an issue witha sub-dependency in Source).


.npmrc
```diff
@@ -0,0 +1,3 @@
+save-prefix=''
+# Avoids a sub-dependency of Source failing to resolve and breaking the build
+shamefully-hoist=true
```
arelra, line 3: [shamefully-hoist](https://pnpm.io/npmrc#shamefully-hoist)

aiui the docs suggest the hoisting issue is coming from our code or tooling rather than a dependency in node_modules?

arelra: Could this be a straightforward-ish fix in source? @sndrs might be interested in this issue

Jakeii:
Awesome work finding and fixing all the broken stuff! 

I had an inkling the jest issue would be resolved by using the `transformIgnorePattern` option that DCR uses.

Changing it to:
```
const esModules = ['@guardian/', 'lodash-es'].join('|');
...
	transformIgnorePatterns: [`/node_modules/.pnpm/(?!${esModules})`],
```

and reinstalling `@guardian/libs` for some reason.

Let's me delete the `.npmrc` overrides without breaking stuff. (Locally anyway...)

i-hardy:
Magic! I'd tried just inserting `.pnpm/` into the string and it didn't work, looks like I'd missed an extra tweak to the regex. Having that array makes it a bit more flexible for adding anything that causes problems in the future too.

.npmrc
i-hardy: Not actually a Source issue in the end; I discovered in our Webpack config we had symlink resolution disabled, so this was probably just the first of many potential problems that the build hit trying to find various modules.

.github/workflows/e2e-playwright.yml
```diff
@@ -26,14 +26,20 @@ jobs:
         with:
           path: ./commercial
 
+      - run: corepack enable
+        working-directory: ./commercial
+        shell: bash
+
       - name: Setup node
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
```
Jakeii, line 34: Can this use your nifty setup node env action too?

Jakeii (approved):
Great to have this done!

arelra (approved):
Great work! Thanks for persevering with this 🎉 

In terms of a release I don't think we need to do a changeset and can just merge. This is a technical chore so there shouldn't be a change in behaviour and all our unit and e2e tests pass.

.npmrc
sndrs: sorry was on leave last week. yeah we did the same in dcr. we found whenever we had to do something to get pnpm to work, it was almost always because we were already doing the wrong (or a wierd) thing somewhere.

sndrs:
> In terms of a release I don't think we need to do a changeset and can just merge. This is a technical chore so there shouldn't be a change in behaviour and all our unit and e2e tests pass.

we also tend to make these kinds of things patches in CSTI, just to so you can isolate changes. otherwise it means the next release _also_ contains this change. this one should be low risk, but it could help finding issues if it turns out something that needed to change to accommodate this had unintended consequences.

arelra:
> > In terms of a release I don't think we need to do a changeset and can just merge. This is a technical chore so there shouldn't be a change in behaviour and all our unit and e2e tests pass.
> 
> we also tend to make these kinds of things patches in CSTI, just to so you can isolate changes. otherwise it means the next release _also_ contains this change. this one should be low risk, but it could help finding issues if it turns out something that needed to change to accommodate this had unintended consequences.

Thanks @sndrs we decided to go with a patch in the end too

PR #1313, opened 2024-03-21
Enable first party cookie for comscore

Jakeii:
## Why?
To deal with the depreciation of third party cookies in Google Chrome, Comscore has requested changes to the tag which currently sits on site. The changes will allow them to drop their cookie as a first-party cookie on our domain.

This has been through and been approved as part of the Tech Review process.

Once added, if consent is given for comscore a new first party cookie will be added `_scor_uid`

### DIFF
--- a/src/init/consented/comscore.ts
+++ b/src/init/consented/comscore.ts
@@ -16,6 +16,9 @@ const getGlobals = (keywords: string): ComscoreGlobals => {
 		c1: comscoreC1,
 		c2: comscoreC2,
 		cs_ucfr: '1',
+		options: {
+			enableFirstPartyCookie: true,
+		},
 	};
 
 	if (keywords !== 'Network Front') {
--- a/.changeset/yellow-hotels-train.md
+++ b/.changeset/yellow-hotels-train.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
+Enable comscore first party cookie
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
 		"@babel/runtime": "^7.11.2",
 		"@guardian/ab-core": "5.0.0",
 		"@guardian/browserslist-config": "^5.1.0",
-		"@guardian/consent-management-platform": "13.12.0",
+		"@guardian/consent-management-platform": "13.12.1",
 		"@guardian/core-web-vitals": "5.1.0",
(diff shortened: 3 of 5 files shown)


emma-imber (approved):

src/init/consented/comscore.ts
```diff
@@ -16,6 +16,9 @@ const getGlobals = (keywords: string): ComscoreGlobals => {
 		c1: comscoreC1,
 		c2: comscoreC2,
 		cs_ucfr: '1',
+		options: {
```
SiAdcock, line 19: I think we also need to be passing a 1st party ID up to comscore as `cs_fpid`. Talking to @Amouzle we could use the [PPID](https://github.com/guardian/commercial/blob/33985dce5c80dc2bd925f5c61298c242ded2baff/src/init/consented/prepare-googletag.ts#L23)

Jakeii: The documentation mentions that it's automatically collected?

mchv: @Jakeii Do you know how how they would automatically collect the PPID?

Amouzle: Ah I think there should be a distinction here:
`cs_fpid`: The first party identifier added by Comscore, this is the mandatory change requested by ComScore by end of March. 
`cs_fpit` (optional): This would be our own identity ID (like we use for PPID) which we would need to populate, allowing ComScore to deduplicate unique users when that data is available. 

There was an ask from the US for us to include `cs_fpit` in this work. However, for expediency, I think we should roll out without this for now and we can add this as a follow up (as I'm not sure all stakeholders are on the same page about the necessity of this anyhow).

Amouzle: (deleted my last comment, I confused my IDs, apologies). Will chat with you @Jakeii today

Jakeii: @mchv Upon further inspection `cs_fpid` is just the value of the cookie `_scor_uid`
Screenshot 2024-03-22 at 10 39 35

Will chat with @SiAdcock and @Amouzle to confirm what we need to pass in

Jakeii: @mchv, and @SiAdcock so it doesn't collect the PPID automatically.

SiAdcock (approved):
👌 Thanks @Jakeii

PR #1326, opened 2024-04-04
Fix collapsing of slots on opt out when no ad is served

Jakeii:
## What does this change?
Collapses the container of slots as we do for consented, now uses the same function used by consented code path to located the parent element that should be removed.

## Why?
The space for ads is sometimes reserved via the slot's container, so it also needs to be removed.

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td> 
 <td> 
</table>


deedeeh (commented):
The code looks good but it seems like there is something wrong with playwright on Github env cause the test passed locally.

Jakeii:
@deedeeh Ah looks like it wasn't waiting any length of time before checking the slot was empty, fixed!

playwright/fixtures/pages/fronts.ts
```diff
@@ -60,4 +60,15 @@ const frontWithPageSkin: Front = {
 	section: 'uk',
 };
 
-export { frontWithPageSkin, fronts, tagPages };
+// This is a tag front, not supported by DCR yet, we can use prod and
```
emma-imber, line 63: Minor - this comment reads like it's incomplete or 'and' should be something else

emma-imber (approved):

PR #1327, opened 2024-04-05
Align devDependencies and peerDependencies major versions (and fix error in github actions)

emma-imber:
## What does this change?
Updates some devDependencies to align with the more up to date peerDependencies major versions. Note that I haven't included a changeset as this change only affects devDependencies.

**Update to add**: tried updating the `nvmrc` version of node to see if that fixed the `URL.canParse` error in the _Set up Node_ action step as it was making all my tests fail. Seems to have magically worked! So will include that change in this PR too. I've rerun the Playwright tests a few times and everything looks stable now.

## Why?
Some of the dependencies were a few major versions out of sync.


Jakeii (approved):
Amazing, the GH action issue has been a huge pain!

PR #1334, opened 2024-04-08
Don't add an advertisement label to native templates served via opt out

emma-imber:
## What does this change?
Checks the opt out ad request for the `'NATIVE'` tag which has been added to the native template in the ad server. If present, we add an `ad-slot--native` class to the ad slot. We then check that this class isn't present before adding a label.

Also removes some redundant logic that was a workaround from when we only used to show house ads in the merch slots. This is no longer needed so we can tidy it up 🧹

## Why?
We'll be using opt out to show some native templates but they currently appear with an advertisement label, which we don't want. This change aligns the code with the way these ads show when they come from GAM.

### Native ad before - label showing
Screenshot 2024-04-08 at 16 23 00

### Native ad after - label not showing
Screenshot 2024-04-08 at 16 22 43

### Normal ad in merch slot after - label still showing as expected
Screenshot 2024-04-08 at 16 20 29

### DIFF
--- a/.changeset/clever-brooms-judge.md
+++ b/.changeset/clever-brooms-judge.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
+Don't add an advertisement label for native templates served through opt out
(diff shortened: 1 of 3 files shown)


deedeeh (approved):
I really like the tidy up along the way in `render-advert-label.ts` 👍🏼

PR #1335, opened 2024-04-09
Use same spacefinder rules for both consented and consentless

Jakeii:
## What does this change?
Make spacefinder accept a function to fill each slot it creates, and use the same rules for filling consentless ads as consented, similar to how an `insertAds` function is passed to spacefinder already.

Also fixed a bug with using the `inline-right` slot on mobile instead of `inline`.

## Why?
The 2 configurations have drifted over time, consentless mostly missing out on updates we've made, which would result in poor ad-ratio by comparison. It was also missing the first `inline1` pass.


src/insert/spacefinder/article.ts
```diff
@@ -492,34 +495,26 @@ const attemptToAddInlineMerchAd = (): Promise<boolean> => {
...
 	return im;
 };
 
-/**
- * Initialise article body ad slots
- */
-export const init = (): Promise<boolean> => {
-	// Also init when the main article is redisplayed
-	// For instance by the signin gate.
-	mediator.on('page:article:redisplayed', doInit);
```
Jakeii, line 519: Mediator hasn't worked for a significant amount of time, I think we should remove this.

emma-imber (approved):
Nice refactor! I think this is easier to read now 🚀

arelra (approved):
Great refactor!

V.minor - the name of the modules that fill inline slots is different for consented and consentless i.e.:

`article-body-adverts`
`article-inline`

Would it help discoverability to align these as (I think) they do the same thing?

Jakeii:
> Great refactor!
> 
> V.minor - the name of the modules that fill inline slots is different for consented and consentless i.e.:
> 
> `article-body-adverts` `article-inline`
> 
> Would it help discoverability to align these as (I think) they do the same thing?

Yes, makes sense! will sort that out

PR #1337, opened 2024-04-09
Add support for nested blocks in articles

Jakeii:
## What does this change?
Add support for nested blocks in articles, relying on a `[data-spacefinder-role="nested"]` data attribute on elements within blocks that support ads in them.

At the moment the candidate selector and many opponent selectors use the [direct child combinator (`>`)](https://developer.mozilla.org/en-US/docs/Web/CSS/Child_combinator), to avoid inserting ads inside of blocks.

We can't simply remove it as we don't want ads inside of most blocks. So we can use the data attribute `[data-spacefinder-role="nested"]` on an element within a block that may also have ads inserted in it (as a direct child only, same as the article root).

I played with the idea of adding the nested indicator as a new spacefinder option to apply to all selectors, but I think it could lead to unexpected results/confusion and better to explicitly add the `[data-spacefinder-role="nested"]` to candidate and opponent rules where needed.

I made an additional change to the way spacefinder selectors work, introducing the use of the `:scope` selector which is [well supported](https://caniuse.com/mdn-css_selectors_scope), so that we can reduce the amount of rules needed. I don't think this was totally necessary but I prefer it, the rule selectors are now all full selectors.

## Why?
New block types are being introduced soon, most pressingly the "Key Takeaway" block.


src/insert/spacefinder/spacefinder.ts
```diff
@@ -306,7 +306,10 @@ const newTestCandidate = (
 	candidate: SpacefinderItem,
 	opponent: SpacefinderItem,
 ): boolean => {
-	if (candidate.element === opponent.element) {
+	if (
+		candidate.element === opponent.element ||
+		opponent.element.contains(candidate.element)
```
arelra, line 311: Would you mind explaining this please?

src/insert/spacefinder/article-body-adverts.ts
```diff
@@ -232,12 +232,12 @@ const addDesktopRightRailAds = (): Promise<boolean> => {
 	const largestSizeForSlot = adSizes.halfPage.height;
 	const rules: SpacefinderRules = {
 		bodySelector: articleBodySelector,
-		candidateSelector: ' > p',
+		candidateSelector: ':scope > p, [data-spacefinder-role="nested"] > p',
```
arelra, line 235: Do we not need scope here?

```suggestion
		candidateSelector: ':scope > p, :scope [data-spacefinder-role="nested"] > p',
```

arelra (approved):
Nice work!

In hindsight I think it would have been preferable to split out the scope updates to a separate PR. The scope chanages are the majority of changes here which makes the nested blocks updates less obvious but lets go with what we have for now.

Jakeii:
> Nice work!
> 
> In hindsight I think it would have been preferable to split out the scope updates to a separate PR. The scope chanages are the majority of changes here which makes the nested blocks updates less obvious but lets go with what we have for now.

I think it would be pretty easy to split it out the `:scope`, So I'll go ahead and remove it from this PR

src/insert/spacefinder/spacefinder.ts
Jakeii: This does looks weird, took a moment to remember, will add a comment.

So for example with key takeaways, the key takeaways block itself will be an opponent but will have descendants that are candidates. The descendant will be trying to avoid it's ancestor.

There's never a situation where want a candidate to avoid an ancestor element,   so we bypass the check.

src/insert/spacefinder/article-body-adverts.ts
Jakeii: Not necessarily, `[data-spacefinder-role="nested"] > p` seems specific enough? happy to prefix all the nested selectors with scope also though

src/insert/spacefinder/spacefinder.ts
arelra: Thanks for updating! Much clearer

PR #1338, opened 2024-04-10
Remove old amiused call

emma-imber:
## What does this change?
Removes an `amiused` call which has been there for a pretty long time.

## Why?
It is used:
Screenshot 2024-04-10 at 10 58 57

Removing this means we can reduce a bit of noise in our logging.


arelra (approved):
🧹

PR #1339, opened 2024-04-10
Interscroller video progress tracking

emma-imber:
## What does this change?
Sends the percentage of a video interscroller ad that was shown to a user as part of commercial metrics. Also finds the `creativeId` of the interscroller so that we can match video progress data to specific creatives. We do this by searching the slots on the page to find which one is using the interscroller template, as only one native interscroller will appear on any one pageview. Tested with two different creatives on CODE and the data is added to commercial metrics payload as expected:

Screenshot 2024-04-22 at 17 49 42
Screenshot 2024-04-22 at 17 52 59

## Why?
DAP would like to be able to offer video progress tracking for our native ads, as this is often a metric that advertisers are particularly interested in which we don't currently support.

### DIFF
--- a/src/core/messenger/background.ts
+++ b/src/core/messenger/background.ts
@@ -230,11 +230,6 @@ const setupBackground = async (
 				video.style.transform = 'translate(-50%, -50%)';
 				background.appendChild(video);
 
-				// video.ontimeupdate = function () {
-				// 	const percent = 100 * (video.currentTime / video.duration);
-				// 	console.log(`${percent}% complete`);
-				// };
-
 				if (!window.guardian.config.switches.sentinelLogger) return;
 
 				const sendVideoProgress = () => {
--- a/src/core/messenger/background.ts
+++ b/src/core/messenger/background.ts
@@ -237,6 +237,7 @@ const setupBackground = async (
 						? '//logs.code.dev-guardianapis.com/log'
 						: '//logs.guardianapis.com/log';
 
+					// TODO: might need to add a new variable to the template to allow us to link video data to a specific creative
 					const videoAdId = 'id_goes_here';
 
 					const event = {
@@ -275,7 +276,7 @@ const setupBackground = async (
 					}
 				};
 
-				// Report all available metrics when the page is unloaded or in background.
+				// Report video ad progress when the page is unloaded or in background.
 				window.addEventListener('visibilitychange', listener, {
 					once: true,
 				});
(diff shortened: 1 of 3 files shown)


src/core/messenger/background.ts
```diff
@@ -229,6 +229,60 @@ const setupBackground = async (
...
+							return slot.getResponseInformation()?.creativeId;
+						}
+					}
+					return null;
+				};
+
+				const sendVideoProgress = () => {
+					const endpoint = window.guardian.config.page.isDev
+						? '//logs.code.dev-guardianapis.com/log'
+						: '//logs.guardianapis.com/log';
```
arelra, line 251: Another option for the data collection is to create an `EventTimer.mark()` so it is sent along with the commercial metrics to the data-lake.

https://github.com/guardian/commercial/blob/f935c3dd1e56f142912dc3399d0e2ee7e8208bba/src/core/event-timer.ts#L215

The `track-scroll-depth` module uses this approach:

https://github.com/guardian/commercial/blob/main/src/core/track-scroll-depth.ts#L15-L16

Was there any particular reason why we have opted to use the fastly endpoint here?

arelra (approved):
Nice work!

This is just a nitpick but the naming of this listener as 'background' is a little confusing to me. Is this message only ever used for interscrollers? Just wondering if we could name it something more descriptive. Could be a follow up.

emma-imber:
> Nice work!
> 
> This is just a nitpick but the naming of this listener as 'background' is a little confusing to me. Is this message only ever used for interscrollers? Just wondering if we could name it something more descriptive. Could be a follow up.

The background file handles interscrollers but also parallax ads - this can include parallax MPUs and fabrics I believe, although I don't know how common these are? Interscrollers are definitely the most common type of ad that this file handles. I think it's called background because all of these ads work by setting a background image on a parent div, so I guess it was just a common theme that linked these ad types? 🤷‍♀️

src/core/messenger/background.ts
emma-imber: No strong reason! It's the same endpoint we use for the `amiused` function so the CODE env was empty and easy to use for testing. Happy to switch to using the same endpoint as commercial metrics though if that's preferred. Does using `EventTimer.mark()` allow us to send two values together (the video progress and interscroller creativeId) or would we need to add them to metrics as separate properties?

emma-imber: I've made the change to add the data to the commercial metrics payload instead

Jakeii (approved):

PR #1342, opened 2024-04-16
New commercial metric for ad block detection

chrislomaxjones:
## What does this change?

Add a new commercial metric _property_ that we can set when we detect the user is running an ad blocker.

Note there is an [existing metric](https://github.com/guardian/commercial-core/blob/03b572557fa492e05b7f30a896acd23d527df1bf/src/core/send-commercial-metrics.ts#L54), but we'd like to keep this distinct as we roll out the new [AdBlockAsk](https://github.com/guardian/dotcom-rendering/pull/11124).

## Why?

So that we can record the number of pageviews for which we perform adblock detection.

### DIFF
--- a/.changeset/calm-dryers-poke.md
+++ b/.changeset/calm-dryers-poke.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
+New commercial metric for recording ad-block detection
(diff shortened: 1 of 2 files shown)


Jakeii (approved):
Do we need to add this the `clean_fastly_commercial` data transformation?

chrislomaxjones:
> Do we need to add this the `clean_fastly_commercial` data transformation?

Good shout I'd overlooked this, it should be safe to do that as a followup.

PR #1344, opened 2024-04-19
Less strict DOM structure for banner containers

chrislomaxjones:
## What does this change?

This PR loosens the assumptions around how the DOM is structured for banner slots (that's front banners and the top-above-nav "banner" container element).

Previously, we assumed the `top-fronts-banner-ad-container` and `top-banner-ad-container` must be direct parents of the ad slot container. Instead, these can now be any ancestor of the ad slot container. This allows for more flexibility in how the DOM is structured, with additional elements placed between the `ad-slot-container` elements and the top-level banner containers.

## Why?

Fix a bug whereby the top-above-nav slot is no longer properly collapsed, due to a change in the DOM structure introduced in https://github.com/guardian/dotcom-rendering/pull/11124.

## How has this been verified?

The existing unit tests pass, and I've added three test cases that assert the desired behaviour in this change.

I've also tested this locally, and observed the top-above-nav slot collapsing when I reject all in the CMP.

### DIFF
--- a/.changeset/friendly-zoos-float.md
+++ b/.changeset/friendly-zoos-float.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
+Assume less about DOM structure when finding ad container elements to remove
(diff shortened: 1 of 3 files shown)


emma-imber (approved):
Makes sense!

PR #1346, opened 2024-04-22
Bump `@guardian/support-dotcom-components`

mxdvl:
## What does this change?

Bumps `@guardian/support-dotcom-components` to `v2.0.1`.
This is a major change [as per our recommendations](https://github.com/guardian/recommendations/blob/main/npm-packages.md#changes-to-peerdependencies-ranges-are-breaking), and consumers should
ensure that they provide a compatible version of the package.

## Why?

The latest is v2.0.1 and this is what is used in `dotcom-rendering`.

### DIFF
--- a/.changeset/tasty-kids-return.md
+++ b/.changeset/tasty-kids-return.md
@@ -1,0 +1,7 @@
+---
+'@guardian/commercial': major
+---
+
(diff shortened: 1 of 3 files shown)


Jakeii:
Can we drop this as a dependency? It's not used here and not a peer of a dep as far as I can see?

PR #1347, opened 2024-04-22
Drop @guardian/support-dotcom-components as a dependency

Jakeii:
## Why?
Whatever code used it has been subsequently been removed, so we can drop it


mxdvl (approved):
Closes #1346

arelra (approved):

PR #1349, opened 2024-04-24
Tweaks for interactive ad article slots

Jakeii:
## What does this change?
Add a size mapping for `interactive` ad slots to our size mapping

Rather than fallback to data attributes, merge them with any other size mappings for that slot.

## Why?
Allows us to remove the `outofpage` and `empty` sizes from interactive slot data attributes.

### DIFF
--- a/src/core/__mocks__/ad-sizes.ts
+++ b/src/core/__mocks__/ad-sizes.ts
@@ -1,0 +1,21 @@
+import { AdSize as RealAdSize } from 'core/ad-sizes';
+
+class AdSize extends RealAdSize {
+	constructor(arraySize: number);
+	constructor([width, height]: [number, number]);
+	constructor(args: [number, number] | number) {
+		// The class needs to be able to be instantiable with a single number like a normal `Array`, so that it can be used in jest, where it uses this internally somehow.
+		if (typeof args === 'number') {
+			super([0, 0]);
+		} else {
+			super(args);
+		}
+	}
+}
+
+const createAdSize = (width: number, height: number): AdSize => {
+	return new AdSize([width, height]);
+};
+
+export * from './ad-sizes';
+export { createAdSize };
--- a/src/core/create-ad-slot.ts
+++ b/src/core/create-ad-slot.ts
@@ -1,5 +1,4 @@
 import { log } from '@guardian/libs';
-import { findAppliedSizesForBreakpoint } from './ad-sizes';
 import type { SizeMapping } from './ad-sizes';
 import { isBreakpoint } from './lib/breakpoint';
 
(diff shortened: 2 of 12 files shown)


emma-imber (dismissed):

src/core/create-ad-slot.ts
```diff
@@ -117,20 +116,24 @@ const concatSizeMappings = (
...
-			// Only perform concatenation if breakpoint is of the correct type
-			if (isBreakpoint(breakpoint)) {
-				const sizes = findAppliedSizesForBreakpoint(
-					sizeMappings,
-					breakpoint,
-				);
-
-				// Concatenate the option sizes onto any existing sizes present for a given breakpoint
-				sizeMappings[breakpoint] = sizes.concat(optionSizes);
+	Object.entries(optionSizeMappings).reduce(
```
Jakeii, line 119: The previous implementation was not correctly combining the sizeMappings, it would have duplicate sizes and add sizes for smaller breakpoints to larger.

I don't think it actually caused any problems? But the results were unexpected.

src/core/lib/breakpoint.ts
```diff
@@ -1,7 +1,7 @@
 /**
  * Breakpoints ordered from smallest to largest
  */
-const breakpoints = ['mobile', 'phablet', 'tablet', 'desktop'] as const;
+const breakpoints = ['mobile', 'phablet', 'tablet', 'desktop', 'wide'] as const;
```
Jakeii, line 4: the wide breakpoint is used by [pageskins](https://github.com/guardian/frontend/blob/main/common/app/views/fragments/commercial/pageSkin.scala.html#L6)

src/core/__mocks__/ad-sizes.ts
```diff
@@ -0,0 +1,21 @@
+import { AdSize as RealAdSize } from 'core/ad-sizes';
+
+class AdSize extends RealAdSize {
```
Jakeii, line 3: There's a jest error that pop's up now and again when trying to assert on instances of `AdSize`, this resolves it. I think jest tries to copy this by using `new AdSize(length)` which won't work.

src/lib/header-bidding/slot-config.spec.ts
```diff
@@ -65,10 +65,7 @@ describe('getPrebidAdSlots', () => {
 		expect(getHeaderBiddingAdSlots(buildAdvert('top-above-nav'))).toEqual([
 			{
 				key: 'top-above-nav',
-				sizes: [
-					[970, 250],
-					[728, 90],
-				],
+				sizes: [createAdSize(970, 250), createAdSize(728, 90)],
```
Jakeii, line 68: The AdSize mock means these all need changing

src/lib/header-bidding/slot-config.spec.ts
```diff
@@ -207,9 +207,9 @@ describe('getPrebidAdSlots', () => {
 		);
 		expect(hbSlots).toHaveLength(1);
 		expect(hbSlots[0]?.sizes).toEqual([
-			[160, 600],
-			[300, 600],
-			[300, 250],
+			adSizes.skyscraper,
```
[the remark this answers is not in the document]
arelra: ... unless the function is tested elsewhere and it is secondary to this test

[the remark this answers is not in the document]
Jakeii: good point!

arelra (approved):
✨

PR #1363, opened 2024-05-07
Add PR check for average time to load top-above-nav

emma-imber:
## What does this change?
Adds a new workflow to PR checks to check the average time for top-above-nav to load over 20 articles. We're using a high number of articles to hopefully average out the effects of any anomalous results. You can view the result by clicking on the details for the **Test ad load time** check and checking the average ad render time value under the **Run Playwright** step.

Screenshot 2024-05-08 at 15 45 45

It would be nice to add the result of this check as a comment on the PR in future, but maybe that would be a good improvement for the future as it isn't necessary now.

## Why?
It'll be useful to have a quick indication of whether a change might speed up or slow down the first ad appearing on the page. Hopefully this added check can give an 'at-a-glance' view of the potential impact of a change.


.github/workflows/test-ad-load-time.yml
```diff
@@ -0,0 +1,75 @@
...
+  test:
+    name: Test time to load top-above-nav
+    timeout-minutes: 30
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      # Commercial
+      - name: Checkout
+        uses: actions/checkout@v3
```
mchv, line 19: checkout@v4

i-hardy (approved):
👍

PR #1365, opened 2024-05-07
Deeply read right ab test

DanielCliftonGuardian:
## What does this change?
Adds 650px to top of desktopRightRailAds when in deeplyReadMostViewed test variant
## Why?
Corresponding change to DCR (pr https://github.com/guardian/dotcom-rendering/pull/11299)
Frontend (pr https://github.com/guardian/frontend/pull/27084)

### DIFF
--- a/src/insert/spacefinder/article.ts
+++ b/src/insert/spacefinder/article.ts
@@ -27,6 +27,14 @@ import { isInHighValueSection } from './utils';
 
 type SlotName = Parameters<typeof createAdSlot>[0];
 
+/**
+ * As estimation of the height of the most viewed island.
+ * This appears from desktop breakpoints on the right-hand side.
+ * Knowing the height of the element is useful when
+ * calculating where to place ads in the right column.
+ */
+const MOST_VIEWED_HEIGHT = 600;
+
 const articleBodySelector = '.article-body-commercial-selector';
 
 const isPaidContent = window.guardian.config.page.isPaidContent;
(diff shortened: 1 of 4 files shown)


src/insert/spacefinder/article.ts
```diff
@@ -210,7 +217,7 @@ const addDesktopInline1 = (fillSlot: FillAdSlot): Promise<boolean> => {
  * Inserts all inline ads on desktop except for inline1.
  */
 const addDesktopRightRailAds = (fillSlot: FillAdSlot): Promise<boolean> => {
-	let minAbove = 1000;
+	let minAbove = isInDeeplyReadMostViewedVariant ? 1650 : 1000;
```
domlander, line 221:
```suggestion
	let minAbove = 1000;
	
	if (isInDeeplyReadMostViewedVariant) {
	    minAbove += 650
	}
```

This follows the existing pattern below and this new code will be slightly easier to remove once the test is complete.

domlander (approved):

src/insert/spacefinder/article.ts
```diff
@@ -212,6 +219,10 @@ const addDesktopInline1 = (fillSlot: FillAdSlot): Promise<boolean> => {
 const addDesktopRightRailAds = (fillSlot: FillAdSlot): Promise<boolean> => {
 	let minAbove = 1000;
 
+	if (isInDeeplyReadMostViewedVariant) {
```
Jakeii, line 222: I think this will lead to an extra large `minAbove` when `isPaidContent` is `true`

domlander: Thanks Jake, good spot. I've updated the code to not add the extra space twice when the Most Viewed island is missing

Jakeii (approved):
Looks good! Just need to add a changeset

PR #1370, opened 2024-05-08
Run changeset publish as changeset-bot

Jakeii:
## What does this change?
Run changeset publish as changeset-bot

## Why
This should allow workflows/checks to run on version bump PRs so we don't have to force-merge


arelra (approved):
Nice!

PR #1271, opened 2024-02-14
Improve spacefinder's handling of ranked and list articles on mobile

Jakeii:
## What does this change?

Sample articles
https://www.theguardian.com/football/2024/feb/09/premier-league-10-things-to-look-out-for-this-weekend
https://www.theguardian.com/culture/2024/feb/15/miski-omar-the-10-funniest-things-i-have-ever-seen-on-the-internet
https://www.theguardian.com/culture/2024/feb/08/say-it-with-a-kiss-the-20-greatest-smooches-on-film-ranked

### Improve ad insertion around headings and `hr` elements

On these articles spacefinder is avoiding placing ads near anything that isn't a paragraph by quite a wide margin of 200px.

We have decided that we wouldn't mind ads appearing above headings, so we can add h2 elements to our candidate selectors, [some articles](https://www.theguardian.com/football/2024/feb/09/premier-league-10-things-to-look-out-for-this-weekend) also place hr's above these headings [and](https://www.theguardian.com/culture/2024/feb/15/miski-omar-the-10-funniest-things-i-have-ever-seen-on-the-internet) [others](https://www.theguardian.com/culture/2024/feb/08/say-it-with-a-kiss-the-20-greatest-smooches-on-film-ranked) have media (images/videos/embeds) directly after that have rules to not have ads within 200px above.

This rule makes sense so that ads are not inserted between paragraphs related to the media and the media itself. However a heading element above would indicate that the content above the heading is not directly related to the media, so an ad above the heading would be fine.

After some iteration, the best way to handle it I thought was to allow headings to "bypass" the minBelow rule for an opponent. So that heading candidates are allowed to have an ad inserted above even if they are above a media element, other elements still need to be further away.

### Handle candidates/opponents with negative margins

There was another issue with this because the headings have a negative top margin and the `hr`'s are only 1 pixel tall, so they were neither above or below it.

Examples of new placement:

https://github.com/guardian/commercial/assets/1731150/2342604b-1a2c-4127-995d-012f122008ae

https://github.com/guardian/commercial/assets/1731150/2c4a9a17-ece9-4386-a092-760dcfcc2d49

## Why?
Otherwise there would be no ads on these articles on mobile and we can't have that.

PR #1377, opened 2024-05-13
Spacefinder handle opponents/candidates overlapping

Jakeii:
## What does this change?
Handle where spacefinder candidates/opponents overlap each other by improving the check for opponents being below - to also a check that `opponent.top` is below `candidate.bottom` and handling when opponents/candidates are adjacent/overlapping.

I suspect this may be happening now due to #1271 which fixed some bugs in the `testCandidate` code, there's a rule [here](https://github.com/guardian/commercial/pull/1377/files#diff-40044f4d0100808638e7ed46790f93dbaff168485b9043e4a6bacb4df4578ae6L174) that set a 500px margin for interactives that I think was a "hack" that worked with the old `testCandidate` but not the new one.

With this change spacefinder can now handle a bit more gracefully this edge case and can remove the "hack" setting the margin to a more reasonable 100px.

Also updated spacefinder debugger to display when it's overlapping like this.

## Why?
The code currently considers any opponent that satisfies `opponent.bottom > candidate.bottom` to be below the candidate and anything that doesn't to be above.

This doesn't account for opponents that are not above or below but adjacent/overlapping:
Untitled-2023-02-24-1349

### DIFF
--- a/.changeset/itchy-cows-relate.md
+++ b/.changeset/itchy-cows-relate.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
+Spacefinder - Handle opponents/candidates overlapping
--- a/playwright/tests/front-inline-slots.spec.ts
+++ b/playwright/tests/front-inline-slots.spec.ts
@@ -15,7 +15,8 @@ test.describe('Slots and iframes load on fronts pages', () => {
 	});
 });
 
-test.describe('Slots and iframes load on tag pages', () => {
+// currently no banners on tag fronts
+test.skip('Slots and iframes load on tag pages', () => {
 	tagPages.forEach(({ path }) => {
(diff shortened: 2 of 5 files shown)


Jakeii:
Tag front has changed breaking the playwright test 🤦‍♂️

domlander:
Will the figure element always finish its rendering before Spacefinder selects ad locations?

https://github.com/guardian/commercial/assets/9574885/799c1ee8-8d7b-4783-99b2-1d0251b973a7

Jakeii:
> Will the figure element always finish its rendering before Spacefinder selects ad locations?
> 
>  Screen.Recording.2024-05-13.at.17.19.36.mov

It was my first thought too, but when investigating the wait for interactives code does seem to work as intended.

SiAdcock (approved):
Makes sense, thanks @Jakeii  👍

PR #1379, opened 2024-05-14
Add right column and mobile sticky ads to football fixtures pages

i-hardy:
## What does this change?

This adds an ad slot in the right hand column on football fixtures pages, and adds these pages to those which will show the mobile sticky ad slot outside the UK

Examples of pages affected:

- https://www.theguardian.com/football/fixtures
- https://www.theguardian.com/football/results
- https://www.theguardian.com/football/tables
- https://www.theguardian.com/football/live
- https://www.theguardian.com/football/euro-2024/fixtures
- https://www.theguardian.com/football/championship/results

### Desktop
fixtures-right

### Mobile
fixtures-mobile

## Why?

This is in response to a request from the US team.


src/insert/fixures.ts
```diff
@@ -0,0 +1,85 @@
...
+			anchor.style.height = '100%';
+			anchor.appendChild(container);
+		})
+		.then(() => fillDynamicAdSlot(slot, false));
+};
+
+const isPageWithAdSpace = () => {
+	const { pageId } = window.guardian.config.page;
+
+	const isFootballPage = pageId.startsWith('football/');
```
Jakeii, line 36: This might fit better in https://github.com/guardian/commercial/blob/main/src/lib/commercial-features.ts#L61

PR #1380, opened 2024-05-15
Insert im after the first inline ad has been inserted

emma-imber:
## What does this change?

- Splits `addInlineAds` into two functions - `addFirstInlineAd` and `addSubsequentInlineAds`
- Splits `addMobileInlineAds` into two functions - `addMobileTopAboveNav` and `addMobileSubsequentInlineAds`
- Changes the order of spacefinder passes so that the `inline1` ad on desktop/tablet and `top-above-nav` ad on mobile are inserted before we find space for an `inline-merch` ad. We then add the right hand rail or subsequent mobile inline ads after the im is inserted. 
 
## Why?
Follow on from this PR https://github.com/guardian/commercial/pull/1360

We want to insert the first inline ad before we insert an im ad, because sometimes the im ad won't serve, and this pushes down the first inline ad, impacting revenue.

After the previous PR, `inline-merch` ads currently don't serve, because spacefinder is inserting all inline ads before the im pass, so there's no room for the im ad on the page. This PR moves the im pass before subsequent inline ads are inserted. This is early enough so that the ad can find a slot, but it still happens after the first inline insertion. I've had to split up the `addInlineAds` function to achieve this.

## Screen recordings of how the inline-merch ad appears now
The im can now sometimes appear at the bottom of the article. I've checked with Alex and we're ok with this happening.

https://github.com/guardian/commercial/assets/108270776/c79b48b4-1ca6-4667-8038-96ac4343e82b

https://github.com/guardian/commercial/assets/108270776/7f9077b5-aad4-4558-9de0-794a7a5f92eb

https://github.com/guardian/commercial/assets/108270776/abfd364e-9275-486b-ac4a-19b2c0b8d53d


src/insert/spacefinder/article.ts
```diff
@@ -462,17 +498,21 @@ const init = async (fillAdSlot: FillAdSlot): Promise<boolean> => {
...
 
-	await addInlineAds(fillAdSlot);
+	// We add the first inline ad before finding a place for an im so that the inline1 doesn't get pushed down
+	await addFirstInlineAd(fillAdSlot);
 
 	const im = window.guardian.config.page.hasInlineMerchandise
 		? attemptToAddInlineMerchAd(fillAdSlot)
 		: Promise.resolve(false);
 	const inlineMerchAdded = await im;
 	if (inlineMerchAdded) await waitForAdvert('dfp-ad--im');
```
Jakeii, line 508: do we still need `waitForAdvert` here? we don't wait for first inline, so should be fine to remove?

src/insert/spacefinder/article.ts
```diff
@@ -350,8 +381,8 @@ const addMobileInlineAds = (fillSlot: FillAdSlot): Promise<boolean> => {
 
 	const insertAds: SpacefinderWriter = async (paras) => {
 		const slots = paras.map(async (para, i) => {
-			const name = i === 0 ? 'top-above-nav' : `inline${i}`;
-			const type = i === 0 ? 'top-above-nav' : 'inline';
+			const name = `inline${i}`;
```
Jakeii, line 384:
```suggestion
			const name = `inline${i + 1}`;
```

emma-imber: Yep you're totally right 🤦‍♀️ Great spot!

src/insert/spacefinder/article.ts
emma-imber: Do we know what this function is used for? I've traced it back and it looks like a pretty old bit of code that was split out from some dfp functionality: https://github.com/guardian/frontend/blob/e0013c355da31e7902a7e985a3d46c861729469a/static/src/javascripts/projects/commercial/modules/dfp/wait-for-advert.js

This is the only time we call function so if we think we don't need it we could delete the file too?

Jakeii: I don't think we need to return the im promise anymore either, could it simplify to:
```ts
if (window.guardian.config.page.hasInlineMerchandise) {
		await attemptToAddInlineMerchAd(fillAdSlot);
}

Jakeii (approved):
Awesome✅

PR #1382, opened 2024-05-16
use fetch to send commercial metrics

Jakeii:
## What does this change?
replace `sendBeacon` with fetch for sending commercial metrics

## Why?
fetch should be more reliable than `sendBeacon`

Upon testing `fetch` will go through on the occasions that `sendBeacon` is blocked.

@emma-imber also [did some testing on this](https://github.com/guardian/commercial/pull/888) a while back and determined `fetch` to be more reliable.

### DIFF
--- a/.changeset/lovely-zoos-wink.md
+++ b/.changeset/lovely-zoos-wink.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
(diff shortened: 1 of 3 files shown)


arelra (approved):
✨

emma-imber (approved):
Very nice! 🚀

PR #1384, opened 2024-05-17
Increase spacefinder mobile minAbove distance by 100px

arelra:
## Why?
The current rule specifies the first ad can appear 100px from the top of the article. This means that sometimes the first ad appears after the opening sentence/paragraph, and this has raised concerns within the editorial team. 

Increasing the rule to 200px should mean that the at least a couple of sentences are included before the first ad on mobile.

Testing suggests the issue is not as common on desktop (where the rule is 300px).

### DIFF
--- a/.changeset/sour-peas-train.md
+++ b/.changeset/sour-peas-train.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': patch
+---
+
+Increase spacefinder minAbove distance by 100px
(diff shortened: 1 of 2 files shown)


Jakeii (approved):
Definitely the right variables to tweak, just need to confirm right numbers!

ashishpuliyel (approved):
Nothing like a very immediate response to an editorial concern.

PR #1386, opened 2024-05-17
Increase spacefinder mobile minAbove distance to 250px

arelra:
## What does this change?

Increase spacefinder mobile minAbove distance to 250px

Tweak to #1384 

## Screenshots

| Before      | After      |
| ----------- | ---------- |
| ![before][] | ![after][] |

[before]: https://github.com/guardian/commercial/assets/7014230/bda2336c-59a6-4956-8518-761fd395bb55
[after]: https://github.com/guardian/commercial/assets/7014230/a81344e3-e630-4ca2-95eb-224d637e818c

### DIFF
--- a/.changeset/green-clouds-hammer.md
+++ b/.changeset/green-clouds-hammer.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': patch
+---
+
+Increase spacefinder mobile minAbove distance to 250px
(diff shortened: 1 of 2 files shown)


ashishpuliyel (approved):

domlander:
I wonder if a change as potentially impactful to revenue such as this (100px -> 250px) should be AB tested. i.e. we keep 5% back at 100px and test the difference in revenue.

arelra:
> I wonder if a change as potentially impactful to revenue such as this (100px -> 250px) should be AB tested. i.e. we keep 5% back at 100px and test the difference in revenue.

You're right @domlander we were asked to do this as a quick change and will reassess next week.

btw 100px was a recent change, it was 200px historically.

Issue #1388, opened 2024-05-19
initTrackScrollDepth exercises Firefox for Android bug causing crash


emdash-ie:
There is [a bug in Firefox for Android](https://bugzilla.mozilla.org/show_bug.cgi?id=1853078) in which `window.innerHeight` is 0 in `window.onload` when a page is opened in a new tab. This was discovered in [a bug report about guardian pages causing Firefox for Android to crash](https://bugzilla.mozilla.org/show_bug.cgi?id=1788530), in which someone tracked the problem to the use of `window.innerHeight` [in initTrackScrollDepth](https://github.com/guardian/commercial/blob/901e31f9699c45cf093227c3632d29fedc03c2c5/src/core/track-scroll-depth.ts#L12).

Firefox for Android presumably will fix this bug, but in the meantime could we make a change here to avoid it (possibly just for firefox for android)? The current crashing behaviour is very frustrating to users (of which I am one), and we could possibly make a change more quickly than Firefox for Android will. Also, workarounds suggested for users in those bug threads are to click reject all on the consent banner or to use a uBlock Origin script to block the commercial bundle, so avoiding triggering the bug may avoid a reduction in the number of users who are shown consented ads.

emdash-ie:
This is also the cause of https://github.com/guardian/dotcom-rendering/issues/6103

mxdvl:
Specifically, a [division by `0` return `Infinity` in JS for the `pageHeightVH`](https://github.com/guardian/commercial/blob/901e31f9699c45cf093227c3632d29fedc03c2c5/src/core/track-scroll-depth.ts#L14)

PR #1390, opened 2024-05-20
Check that innerHeight is greater than 0

emma-imber:
## What does this change?
Checks that `window.innerHeight` is more than 0 before we then divide by that value and loop through infinity. Dividing by 0 is not a thing we want to do!

## Why?
There's a bug on Firefox for Android where the `window.innerHeight` for a new tab can be 0, which means we end up looping through infinity which can crash the page. See issue here: https://github.com/guardian/commercial/issues/1388

### DIFF
--- a/src/core/track-scroll-depth.ts
+++ b/src/core/track-scroll-depth.ts
@@ -10,35 +10,43 @@ import { EventTimer } from './event-timer';
 const initTrackScrollDepth = () => {
 	const pageHeight = document.body.offsetHeight;
 	const intViewportHeight = window.innerHeight;
-	// how many viewports tall is the page?
-	const pageHeightVH = Math.floor(pageHeight / intViewportHeight);
-	const eventTimer = EventTimer.get();
-	eventTimer.setProperty('pageHeightVH', pageHeightVH);
 
-	const observer = new IntersectionObserver(
-		/* istanbul ignore next */
-		(entries) => {
-			entries.forEach((entry) => {
-				if (entry.isIntersecting) {
-					const currentDepthVH = String(
-						entry.target.getAttribute('data-depth'),
-					);
-					log('commercial', `current scroll depth ${currentDepthVH}`);
-					eventTimer.mark(`scroll-depth-vh-${currentDepthVH}`);
-					observer.unobserve(entry.target);
-				}
-			});
-		},
-	);
+	// this if statement is here to handle a bug in Firefox in Android where the innerHeight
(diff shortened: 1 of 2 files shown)


arelra (approved):

src/core/track-scroll-depth.ts
```diff
@@ -10,6 +10,13 @@ import { EventTimer } from './event-timer';
 const initTrackScrollDepth = () => {
 	const pageHeight = document.body.offsetHeight;
 	const intViewportHeight = window.innerHeight;
+
+	// this if statement is here to handle a bug in Firefox in Android where the innerHeight
+	// of a new tab can be 0, so we end up dividing by 0 and looping through infinity
+	if (intViewportHeight === 0) {
+		return;
```
domlander, line 17: Maybe also logging to the commercial logger would be useful here? In case a developer is using firefox and debugging scroll depth tracking

domlander (approved):

src/core/track-scroll-depth.ts
emma-imber: Good suggestion!

src/core/track-scroll-depth.ts
```diff
@@ -10,6 +10,17 @@ import { EventTimer } from './event-timer';
 const initTrackScrollDepth = () => {
 	const pageHeight = document.body.offsetHeight;
 	const intViewportHeight = window.innerHeight;
+
+	// this if statement is here to handle a bug in Firefox in Android where the innerHeight
+	// of a new tab can be 0, so we end up dividing by 0 and looping through infinity
+	if (intViewportHeight === 0) {
```
mxdvl, line 16: Just chiming to ensure we capture other “invalid values”. I think you’d probably want to discard any value that’s below 1, such as 0.75 or 0.000001, for example, which would yield to `pageHeightVH` to be greater than `pageHeight`. Not sure if this could actually happen, so feel free to disregard!

emma-imber: Doesn't `window.innerHeight` only return integer values?

mxdvl: It should do… unless there's a bug in the implementation… like returning 0 😜

emma-imber: 0 is of type integer though, so although it causes issues here it isn't a breach in type. Given this, I feel like getting a value between 0 and 1 is pretty unlikely.

PR #1392, opened 2024-05-20
Prepare the code to remove GA

deedeeh:


deedeeh:
I need to revert the commit and that will close the PR so will close that one and open a new PR.

PR #1396, opened 2024-05-29
Only play the video interscroller when it's in the viewport

emma-imber:
## What does this change?
Uses an intersection observer to play the video interscroller when it's in the viewport, and pause it when it isn't. This is the default behaviour on Chrome on Android, but not on other browsers.

## Why?
This replicates the behaviour of the fabric video ad, and should improve the accuracy of the video progress data we've been seeing.

### Before - on Safari
Video does not pause when we scroll away from the ad.

https://github.com/guardian/commercial/assets/108270776/dad6c0cb-059b-4a0e-8c69-ea28f00d10ff

### After - on Safari
Video does pause when we scroll away from the ad.

https://github.com/guardian/commercial/assets/108270776/c5e85c9f-94f6-4301-8b9c-888f6288513e

### DIFF
--- a/src/core/messenger/background.ts
+++ b/src/core/messenger/background.ts
@@ -251,9 +251,9 @@ const setupBackground = async (
 					(entries) => {
 						entries.forEach((entry) => {
 							if (entry.isIntersecting && !played) {
-								void video.play();
+								video.paused && void video.play();
 							} else {
-								video.pause();
+								!video.paused && video.pause();
 							}
 						});
 					},
--- a/src/core/messenger/background.ts
+++ b/src/core/messenger/background.ts
@@ -244,6 +244,25 @@ const setupBackground = async (
 					return undefined;
 				};
 
+				let played = false;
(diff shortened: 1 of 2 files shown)


src/core/messenger/background.ts
```diff
@@ -244,6 +244,24 @@ const setupBackground = async (
...
 				};
 
+				let played = false;
+				video.onended = () => (played = true);
+
+				const observer = new IntersectionObserver(
+					(entries) => {
+						entries.forEach((entry) => {
+							if (entry.isIntersecting && !played) {
+								void video.play();
```
Jakeii, line 254: is it okay to call play` and `pause` every time the callback is called? maybe throw a `video.paused && void video.play()` and `!video.paused && void video.pause()` in there?

emma-imber: Yeah fair point - added!

Jakeii (approved):
Nice!

PR #1401, opened 2024-06-03
Tweaks to benchmarking

Jakeii:
## What does this change?
Builds on the work done by @emma-imber in https://github.com/guardian/commercial/pull/1363

Reconfigures phases into seperate [projects](https://playwright.dev/docs/test-projects) which can depend on one another.

Use a project to set up the consent state and [save the storage state](https://playwright.dev/docs/api-testing#reusing-authentication-state) for very rapid testing (we can apply this to our other tests!)

Builds a prod build then runs the tests against prod while overriding commercial files to the locally built versions to get as realistic results as possible.

Run the benchmarks in parallel

Write the results to files so they can be averaged across the workers.

Adds a little PR message with the results (see below).

## Why?
Often we don't confidently know how changes will affect ad rendering time, e.g. [this PR](https://github.com/guardian/commercial/pull/1397) where we're not sure, and running an AB test just to find out isn't optimal. 

~~It's also revealed that with poorer network connections the difference between ad load times between consented vs consentless is not very pronounced 🤔~~
That was a mistake the consentless benchmark was still accepting all!


webpack.config.prod.js
```diff
@@ -16,11 +16,13 @@ const gitCommitSHA = () => {
...
 };
 
+const prefix = process.env.BUNDLE_PREFIX ?? '[chunkhash]/';
+
 module.exports = webpackMerge.smart(config, {
 	mode: 'production',
 	output: {
-		filename: `[chunkhash]/graun.standalone.commercial.js`,
-		chunkFilename: `[chunkhash]/graun.[name].commercial.js`,
+		filename: `${prefix}graun.standalone.commercial.js`,
```
Jakeii, line 24: so that we can optionally build prod without a hash, so playwright can know file names to override the requests to.

playwright.benchmark.config.ts
```diff
@@ -0,0 +1,73 @@
+import { defineConfig, devices } from '@playwright/test';
```
Jakeii, line 1: New seperate playwright config using playwright [projects](https://playwright.dev/docs/test-projects) which can depend on each other.

.github/workflows/test-ad-load-time.yml
```diff
@@ -21,38 +21,56 @@ jobs:
...
-      - name: Start Commercial server
-        run: pnpm serve & npx wait-on -v -i 1000 http://localhost:3031/graun.standalone.commercial.js
+      - name: Build Prod (used for rewriting by playwright)
+        run: pnpm build:prod
+        env:
+          BUNDLE_PREFIX: ''
 
       - name: Run Playwright
-        run: pnpm playwright test test-ad-load-time.spec.ts
+        run: pnpm benchmark
```
emma-imber, line 33: Should we update the run command in the `e2e-playwright.yml` file to reflect the new file structure?

emma-imber (approved):
Very nice!

.github/workflows/test-ad-load-time.yml
```diff
@@ -21,38 +21,56 @@ jobs:
...
+            const consentless = readFileSync('./benchmark-results/consentless/average.txt', 'utf8');
+
+            const body = `### Ad load time test results
+
+            For \`consented\`, \`top-above-nav\` took on average ${consented}ms to load.
+            For \`consentless\`, \`top-above-nav\` took on average ${consentless}ms to load.
+
+            Test conditions:
+            - 5mpbs download speed
+            - 1.5mpbs upload speed
```
emma-imber, line 50: Units:
```suggestion
            - 5mbps download speed
            - 1.5mbps upload speed
```

.github/workflows/test-ad-load-time.yml
Jakeii: yeah good spot

PR #1403, opened 2024-06-04
Add 300x50 ad size to mobile-sticky

deedeeh:
## What does this change?

This enables a new ad size 300x50 to `mobile-sticky`.

I tested in CODE and using a VPN to be in a region that has a `mobile-sticky` slot. Here is a screenshot from `googletag.openConsole()` with the new size.

image

## Why?

This request was made by AdOps which will give us the opportunity to serve creatives in smaller size.

### DIFF
--- a/src/core/ad-sizes.ts
+++ b/src/core/ad-sizes.ts
@@ -389,7 +389,7 @@ const slotSizeMappings = {
 		mobile: [adSizes.fluid],
 	},
 	'mobile-sticky': {
-		mobile: [adSizes.mobilesticky, adSizes.empty],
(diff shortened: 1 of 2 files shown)


src/core/ad-sizes.ts
```diff
@@ -389,7 +389,7 @@ const slotSizeMappings = {
 		mobile: [adSizes.fluid],
 	},
 	'mobile-sticky': {
-		mobile: [adSizes.mobilesticky, adSizes.empty],
+		mobile: [adSizes.mobilesticky, adSizes.empty, createAdSize(300, 50)],
```
dskamiotis, line 392: Looking good 👏

dskamiotis (approved):

PR #1405, opened 2024-06-05
Delete redplanet

emma-imber:
## What does this change?
Deletes all files related to the running of redplanet - this includes launchpad, which is used by the redplanet code.

There will also be PRs needed to delete the switch in frontend, and references to the switch in dotcom-rendering. We can also look to remove redplanet from the vendor list when these code changes have been made.

## Why?
We have permission to remove redplanet from the site.

## Ad load speed impact

Results on this PR are about 300ms quicker for consented.


Jakeii (approved):
🧹Big if ad load results are to be trusted, seems to be all over the place! in the changeset PR its 4400 🤔https://github.com/guardian/commercial/pull/1398#issuecomment-2147432799

arelra (approved):
Ah the smell of freshly deleted code in the morning ✨

PR #1415, opened 2024-06-11
Only import sf debug tools if sfdebug url param is present

emma-imber:
## What does this change?
Moves the check for the `?sfdebug` url param to be before the debug tools are imported. This allows us to only import the debug tools if someone wants to run debug mode.

## Why?
At the moment, we import the 6kb of code needed for spacefinder debug mode for all users. By selectively importing when the `?sfedebug` param is present, we can ship less javascript to users.

### DIFF
--- a/.changeset/loud-cameras-learn.md
+++ b/.changeset/loud-cameras-learn.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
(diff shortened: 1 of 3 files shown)


deedeeh (approved):
I just left a tiny comment but it's non blocking. It would be nice to have a screenshot for the reduced size when we don't have `sfdebug` and when we have it.

PR #1416, opened 2024-06-11
Add test to compare sendBeacon and fetch numbers for users with no ad blocker

emma-imber:
## What does this change?
Adds a test `sendBeacon` and `fetch` call to 10% of messenger background calls.

## Why?
Jake and Ashish realised that `sendBeacon` is blocked by some ad blockers. In earlier testing, we discovered that `sendBeacon` is less reliable than `fetch`. It would be good to test if **all** of the flakiness of `sendBeacon` is explained by ad blockers. By adding a test to a part of the code that only executes if there’s no ad blocker, we can see if we then observe parity between `fetch` and `sendBeacon`.

This messenger code is only called by certain commercial templates, which will only be sending messages on the page if there is no ad blocker. This allows us to control for ad blockers, and hopefully to determine if ad blockers are the sole cause of `sendBeacon`'s flakiness.

### DIFF
--- a/.changeset/big-pens-agree.md
+++ b/.changeset/big-pens-agree.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': patch
+---
+
+Add sendBeacon test to messenger code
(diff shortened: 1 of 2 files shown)


Jakeii (approved):

Jakeii:
🕵️

PR #1417, opened 2024-06-13
Reduce CLS for inline1 when Teads passback - step 1

deedeeh:
## What does this change?

This PR removes the second layout shift from `inline1` if Teads passes back the slot and centres in the middle of the slot in the case of an `mpu` ad size. Check the [ticket](https://trello.com/c/YxHvJgnv/2931-blip-reduce-layout-shift-for-inline1-teads-step-1) for more info.

For context we reserve `250px` for `inline1` in articles. Teads resizes it to `350px` (first shift).
If Teads passback then it resizes back to `250px` (second shift).

I made some checks to implement the changes just for desktop as we don't want to have a `350px` height for mobile.

These changes are tested locally and in CODE.

You can test an mpu locally by adding `?adtest=teadspassback2022` to the end of the article URL.

You can test a direct sold outstream by adding `?adtest=outstreampassback2024` to this article URL https://www.theguardian.com/technology/2017/mar/02/the-legend-of-zelda-breath-of-the-wild-review-link-nintendo-switch

**Before** 

https://github.com/guardian/commercial/assets/23424805/2556eb41-ad3a-4830-a909-2262b09e18c4

**After**

https://github.com/guardian/commercial/assets/23424805/bc195659-bed6-479d-867d-ff7a64117196

## Why?
Reduces CLS which means better user experience.

### DIFF
--- a/src/core/messenger/passback.ts
+++ b/src/core/messenger/passback.ts
@@ -127,14 +127,14 @@ const init = (register: RegisterListener): void => {
 								passbackElement,
 							);
 						})
-						.then(() => passbackElement.id);
+						.then(() => passbackElement);
 				},
 			);
 
 			/**
 			 * Create and display the new passback slot
 			 */
-			void createNewSlotElementPromise.then((passbackElementId) => {
+			void createNewSlotElementPromise.then((passbackElement) => {
 				/**
 				 * Find the initial slot object from googletag
 				 */
@@ -196,7 +196,7 @@ const init = (register: RegisterListener): void => {
 							event: googletag.events.SlotRenderEndedEvent,
 						) {
 							const slotId = event.slot.getSlotElementId();
-							if (slotId === passbackElementId) {
+							if (slotId === passbackElement.id) {
 								const size = event.size;
 								if (Array.isArray(size) && size[1]) {
 									const adHeight = size[1];
@@ -205,34 +205,28 @@ const init = (register: RegisterListener): void => {
 										`Passback: ad height is ${adHeight}`,
 									);
 									void fastdom.mutate(() => {
-										//The 100 is to redue CLS if Teads passes back so we don't resize back to 274px for desktop if we get an mpu
-										const slotHeight =
-											getCurrentBreakpoint() === 'mobile'
-												? `${
-														adHeight + adLabelHeight
-												  }px`
-												: `${
-														adHeight +
-														adLabelHeight +
-														100
-												  }px`;
+										const slotHeight = `${
+											(getCurrentBreakpoint() === 'mobile'
+												? adHeight
+												: adSizes.outstreamDesktop
+														.height) + adLabelHeight
+										}px`;
 										log(
 											'commercial',
 											`Passback: setting height of passback slot to ${slotHeight}`,
 										);
 										slotElement.style.height = slotHeight;
-										const passbackElement =
-											document.getElementById(
-												`${passbackElementId}`,
-											) as HTMLDivElement | null;
-										if (
-											passbackElement &&
-											getCurrentBreakpoint() ===
-												'desktop' &&
-											passbackElement.offsetHeight === 250
-										) {
-											passbackElement.style.top = '60px';
-										}
+
+										/*The centre styling is added in here instead of where the element is created
+										because googletag removes the display style on the passbackElement */
+										passbackElement.style.display = 'flex';
+										passbackElement.style.flexDirection =
+											'column';
+										passbackElement.style.justifyContent =
+											'center';
+										passbackElement.style.height =
+											'calc(100% - 24px)';
+
 										// Also resize the initial outstream iframe so
 										// it doesn't block text selection directly under
 										// the new ad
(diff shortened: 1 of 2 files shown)


src/core/messenger/passback.ts
```diff
@@ -205,15 +205,34 @@ const init = (register: RegisterListener): void => {
 										`Passback: ad height is ${adHeight}`,
 									);
 									void fastdom.mutate(() => {
-										const slotHeight = `${
-											adHeight + adLabelHeight
-										}px`;
+										//The 100 is to redue CLS if Teads passes back so we don't resize back to 274px for desktop if we get an mpu
```
deedeeh, line 208: I know the ticket probably mentions desktop for removing the second shift and don't resize back to `250px` but does that means we keep mobile as it is at the moment? I need some clarification on that.

src/core/messenger/passback.ts
```diff
@@ -205,15 +205,34 @@ const init = (register: RegisterListener): void => {
...
+										const slotHeight =
+											getCurrentBreakpoint() === 'mobile'
+												? `${
+														adHeight + adLabelHeight
+												  }px`
+												: `${
+														adHeight +
```
Jakeii, line 215: I don't think we need to use the ad height here as the number we want is not related to it (350px + label), also I think in some circumstances the passback ad can also be 350px so that would reserve 450px.

src/core/messenger/passback.ts
```diff
@@ -205,15 +205,34 @@ const init = (register: RegisterListener): void => {
...
+											document.getElementById(
+												`${passbackElementId}`,
+											) as HTMLDivElement | null;
+										if (
+											passbackElement &&
+											getCurrentBreakpoint() ===
+												'desktop' &&
+											passbackElement.offsetHeight === 250
+										) {
+											passbackElement.style.top = '60px';
```
Jakeii, line 234: I think this would be better to do with css, also as mentioned above I think the passback ad can be a different size.

src/core/messenger/passback.ts
deedeeh: Do you think it would be better to check `size[1]` if it is not `350` then we add that 100? or better not touch that code at all?

src/core/messenger/passback.ts
deedeeh: When you say the CSS you mean in DCR https://github.com/guardian/dotcom-rendering/blob/669991c0ce39b03cebb4e005a8f408ea1dfb6c9a/dotcom-rendering/src/components/ArticleContainer.tsx#L160?

I know the above link will take me to the reserved part and it will be done in step 2 https://trello.com/c/l3yqAcyP/2936-blip-reduce-layout-shift-for-inline1-teads-step-2

src/core/messenger/passback.ts
```diff
@@ -205,15 +205,34 @@ const init = (register: RegisterListener): void => {
...
+										const slotHeight =
+											getCurrentBreakpoint() === 'mobile'
+												? `${
+														adHeight + adLabelHeight
+												  }px`
+												: `${
+														adHeight +
+														adLabelHeight +
+														100
+												  }px`;
```
Jakeii, line 218:
```suggestion
												: `${
													adSizes.outstreamDesktop.height +
														adLabelHeight
												  }px`;
```

src/core/messenger/passback.ts
Jakeii: Have added a suggestion below

src/core/messenger/passback.ts
Jakeii: Sorry, to clarify, I think checking the height of the element and adding 60px feels a little unreliable, I would suggest using flexbox to automatically centre the ad no matter what height it may be.

src/core/messenger/passback.ts
deedeeh: Thank you for the suggestion Jake.

src/core/messenger/passback.ts
deedeeh: I agree with you. Flexbox seems a better solution. Thank you for clarifying more on a call.

src/core/messenger/passback.ts
deedeeh: Done in 8ab6cacd77fd5b4a342079eab203c4b96b35a692

src/core/messenger/passback.ts
```diff
@@ -206,14 +206,29 @@ const init = (register: RegisterListener): void => {
...
+										because googletag removes the display style on the passbackElement */
+										passbackElement.style.display = 'flex';
+										passbackElement.style.flexDirection =
+											'column';
+										passbackElement.style.justifyContent =
+											'center';
+										passbackElement.style.alignItems =
+											'center';
+										passbackElement.style.height =
+											'calc(100% - 24px)';
```
arelra, line 230: Is the 24px for the ad label height?

```suggestion
											`calc(100% - ${adLabelHeight}px)`;
```

deedeeh: Yes Ravi. I am not sure why I missed adding the variable instead of the label height. Thank you for the suggestion.

arelra (approved):
Nice work! 🎉

PR #1419, opened 2024-06-17
fix native ads cut off with max height

dskamiotis:
## What does this change?
This allows the viewport to be made <980px in width or to change orientation on an iPad without the native ad getting cut off at the the bottom.

## Why?
Native ads `resize` on render and therefore change ratio (width and height), therefore when we reduce the width of the viewport or change the orientation on the iPad the bottom gets cut off.
The CSS states a max-height of 624px which persists at all breakpoints.

The `resize` is a Post Message, so when messenger tells us to resize (which only happens for in-house creatives) it should also check the container and remove max-height if necessary.

### Screenshot

| Before | After |
|---|---|
| Screenshot 2024-06-17 at 16 42 03 | Screenshot 2024-06-17 at 16 01 52  |

### DIFF
--- a/src/core/messenger/resize.ts
+++ b/src/core/messenger/resize.ts
@@ -67,6 +67,7 @@ const resize = (
 
 		if (iframeContainer) {
 			Object.assign(iframeContainer.style, styles);
+			adSlot.style.maxHeight = 'none';
 		}
 	});
 };
--- a/src/core/messenger/resize.ts
+++ b/src/core/messenger/resize.ts
@@ -67,8 +67,9 @@ const resize = (
 
 		if (iframeContainer) {
 			Object.assign(iframeContainer.style, styles);
-			adSlot.style.maxHeight = 'none';
 		}
+
+		adSlot.style.maxHeight = 'none';
 	});
 };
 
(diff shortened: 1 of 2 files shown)


src/core/messenger/resize.ts
```diff
@@ -62,6 +62,8 @@ const resize = (
 		styles.height = normalise(specs.height);
 	}
 
+	adSlot.style.maxHeight = 'none';
```
Jakeii, line 65: This needs to be inside the `fastdom.mutate`

src/core/messenger/resize.ts
```diff
@@ -67,6 +67,7 @@ const resize = (
 
 		if (iframeContainer) {
 			Object.assign(iframeContainer.style, styles);
+			adSlot.style.maxHeight = 'none';
```
Jakeii, line 70: Should this be in the `iframeContainer` conditional?

dskamiotis: I thought this would be appropriate, as we need an iFrameContainer to be present in order to apply the styling?
Do you mean to add the styling regardless of conditions?
```js
return fastdom.mutate(() => {
    Object.assign(iframe.style, styles);

    if (iframeContainer) {
        Object.assign(iframeContainer.style, styles);
    }
    
    adSlot.style.maxHeight = 'none'; // Applied regardless of iframeContainer
});
```

Jakeii (approved):

PR #1430, opened 2024-06-20
Tell dependabot not to bump major versions

Jakeii:
## Why?
Major versions may contain breaking changes, so we'll want to do those bumps manually


dskamiotis (approved):
Makes sense, assuming this is part of the deploy process?

Jakeii:
@dskamiotis this is for dependabot config https://github.com/guardian/commercial/pull/1434

PR #1438, opened 2024-06-25
Add commercialMetricsInitialised to window

emma-imber:
## What does this change?
Adds a new property - `commercialMetricsInitialised` - to the Guardian window config.

## Why?
At the moment, `bypassCommercialMetricsSampling` doesn't work in the commercial repository, because DCR initialises the metrics, so the commercial bundle doesn't have access to that version of the `initialised` variable. By storing the state of initialisation on the window instead, we can share the value between DCR and commercial, which means `bypassCommercialMetricsSampling` should now work for commercial code.

Tested on CODE:
Screenshot 2024-06-25 at 16 56 29

### DIFF
--- a/.changeset/silly-pandas-give.md
+++ b/.changeset/silly-pandas-give.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
+Add commercialMetricsInitialised to the window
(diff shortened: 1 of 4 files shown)


Jakeii (approved):
Nice! Is it no longer logging the metrics not initialised message?

emma-imber:
> Nice! Is it no longer logging the metrics not initialised message?

Yeah I had a look on CODE and it seems to be behaving itself now - not logging the 'metrics not initialised message' when the metrics are indeed initialised

PR #1439, opened 2024-06-25
Configure dependabot to keep up to date with dependency bumps in Commercial

deedeeh:
## What does this change?

This PR groups dependencies in `dependabot.yml` and adds more configuration rules to make it granular and easier to keep up to date with dependency bumps in Commercial.

## Why?

The main reason is to avoid being out of sync with other dependencies we rely on to run Commercial or the other way around when other dependencies relies on Commercial . This will save us time and development effort.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,10 +22,9 @@ updates:
       guardian:
         patterns:
           - '@guardian/*'
-      typescript:
+      types:
         patterns:
           - '@types/*'
-          - 'typescript'
           - 'type-fest'
           - 'tslib'
       eslint:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -9,14 +9,98 @@ updates:
...
+          - 'tslib'
+          - '*eslint*'
+          - '*prettier*'
+          - '@playwright/*'
+          - '*jest*'
+          - '*webpack*'
+          - '*-loader'
+      # We can have a PR for all major updates for production or development dependencies.
+      # if we want to be more granular we can create a group with update-type major for each dependency.
+      major-dependencies:
```
deedeeh, line 94: I grouped major for dependencies and a separate group for devDependencies. I know Jake suggested to have a single major PR for each dependency but as far as I understand there is no way to do that unless we specify a group with for single dependency with major.

SiAdcock: If we omit group config for major dependencies, will it just continue to raise single PRs for each major dependency update?

From [Configure dependabot.yml > groups](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups): 

> If a dependency doesn't belong to any group, Dependabot will continue to raise single pull requests to update the dependency to its latest version as normal. GitHub reports in the logs if a group is empty. For more information, see "[Dependabot fails to group a set of dependencies into a single pull request](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/troubleshooting-dependabot-errors#dependabot-fails-to-group-a-set-of-dependencies-into-a-single-pull-request)."

deedeeh: Yes you are right! Not sure how I missed that. Shall I remove the `major-dependencies` and `major-devDependencies` groups so we expect to have a single dependency PR for major?

SiAdcock: Yes, I think that would be a good starting place. We can always change it if it's too noisy 😊

deedeeh: Ok sounds good. Addressed in 7f55a9881b0974a9504a54eeb923124199509ece

SiAdcock (approved):

PR #1458, opened 2024-07-05
Spacefinder rule optimisation

Jakeii:
## What does this change?
Move spacefinder rules to their own file
Remove the use of the "catch-all" rules, aka the ones that look like:
```js
	let ignoreList = `:scope > :not(p):not(h2):not(ul):not(.${adSlotContainerClass}):not(#sign-in-gate):not(.sfdebug)`;
	if (hasLeftCol) {
		ignoreList +=
			':not([data-spacefinder-role="richLink"]):not([data-spacefinder-role="thumbnail"])';
	}
```

## Why?

What these selectors are doing is selecting elements that are not candidates or already have an opponents selector rule, effectively a "catch-all" for any other elements.

The behaviour (and the selector) is a little fuzzy and confusing, it would be better if we could reliably select all the elements we want to avoid.

Since all articles are rendered by DCR now, we know all the blocks/elements that can appear (https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/lib/renderElement.tsx), so we can more reliably select them.

We already have `data-spacefinder-role` on all blocks wrapped in a figure, which tells us which columns left or right the element expands into, [by adding it to the rest of elements](https://github.com/guardian/dotcom-rendering/pull/11919) (except candidate elements i.e. paragraphs and headings), we can use it to reliable select the elements/blocks to avoid.

The changes in this PR will replicate the current behaviour of avoiding most elements by 400px, which may not be optimal, but we can look into that in later PRs.

### DIFF
--- a/src/insert/spacefinder/article.ts
+++ b/src/insert/spacefinder/article.ts
@@ -7,62 +7,19 @@ import {
 	wrapSlotInContainer,
 } from 'core/create-ad-slot';
 import { spaceFiller } from 'insert/spacefinder/space-filler';
-import type {
-	RuleSpacing,
-	SpacefinderRules,
-	SpacefinderWriter,
-} from 'insert/spacefinder/spacefinder';
+import type { SpacefinderWriter } from 'insert/spacefinder/spacefinder';
 import { commercialFeatures } from 'lib/commercial-features';
-import {
-	getCurrentBreakpoint,
-	getCurrentTweakpoint,
-} from 'lib/detect/detect-breakpoint';
+import { getCurrentBreakpoint } from 'lib/detect/detect-breakpoint';
 import fastdom from '../../utils/fastdom-promise';
 import { computeStickyHeights, insertHeightStyles } from '../sticky-inlines';
 import { initCarrot } from './carrot-traffic-driver';
-import { isInHighValueSection } from './utils';
+import { rules } from './rules';
 
 type SlotName = Parameters<typeof createAdSlot>[0];
-
-/**
- * As estimation of the height of the most viewed island.
- * This appears from desktop breakpoints on the right-hand side.
- * Knowing the height of the element is useful when
- * calculating where to place ads in the right column.
- */
-const MOST_VIEWED_HEIGHT = 600;
 
 const articleBodySelector = '.article-body-commercial-selector';
 
 const isPaidContent = window.guardian.config.page.isPaidContent;
-
-const hasImages = !!window.guardian.config.page.lightboxImages?.images.length;
-
-const hasShowcaseMainElement =
-	window.guardian.config.page.hasShowcaseMainElement;
-
-const minDistanceBetweenRightRailAds = 500;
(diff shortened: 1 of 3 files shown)


src/insert/spacefinder/article.ts
```diff
@@ -143,47 +100,6 @@ const decideAdditionalSizes = async (
...
-			},
-			[`.${adSlotContainerClass}`]: {
-				minAboveSlot: 500,
-				minBelowSlot: 500,
-			},
-			[ignoreList]: {
-				minAboveSlot: 35,
-				minBelowSlot: 400,
-			},
-			'[data-spacefinder-role="immersive"]': {
```
deedeeh, line 176: Why is there a different `minAbove` and minBelow` for immersive in the refactor? I believe it called `rightColumnOpponentSelector` in `rules.ts` which has `minAboveSlot: 35` and `minBelowSlot: 400`.

src/insert/spacefinder/article.ts
```diff
@@ -143,47 +100,6 @@ const decideAdditionalSizes = async (
...
-				minAboveSlot: 0,
-				minBelowSlot: 600,
-			},
-			'figure.element--supporting': {
-				minAboveSlot: 100,
```
deedeeh, line 181: The same for this one which I can see it is one of the `inlineOpponentSelector` values but it has different `minAbove` and `minBelow` than what we have in the old code.

src/insert/spacefinder/rules.ts
```diff
@@ -0,0 +1,221 @@
...
+const mobileOpponentSelectorRules = {
+	// don't place ads right after a heading
+	[mobileHeadingSelector]: {
+		minAboveSlot: 100,
+		minBelowSlot: 0,
+	},
+	[adSlotContainerSelector]: {
+		minAboveSlot: minDistanceBetweenInlineAds,
+		minBelowSlot: minDistanceBetweenInlineAds,
+	},
```
deedeeh, line 130: I haven't seen an `adSlotContainerSelector` used for the old code for `mobileSubsequentInlineAds` also the value for these will be `500` or `750`. I haven't noticed any high numbers like that for mobile breakpoint. Could you please explain why we have it?

deedeeh (commented):
I really enjoyed reviewing the PR! I left couple of comments.

src/insert/spacefinder/article.ts
Jakeii: I think this rule is actually a mistake, there's no reason for inline1 to be 600px from immersive elements! 400 also seems excessive but safer I think.

src/insert/spacefinder/article.ts
Jakeii: I found it odd that supporting elements have their own rule (something to be investigated I think), but you're right I'll add this back in.

deedeeh: I believe if it needs investigation then like you said we can have it back and probably there will be another iteration to drop the rules that we think we might not need it.

PR #1459, opened 2024-07-08
Ogury passback for mobile sticky slot

arelra:
## What does this change?

Adds a new Passback handler for the vendor Ogury which runs in the mobile-sticky slot.

- added a decideSizes function based on the source of the passback being Teads or Ogury
- passback.ts will now disable refresh of the parent slot as mobile-sticky refreshes by default

### DIFF
--- a/.changeset/empty-icons-mate.md
+++ b/.changeset/empty-icons-mate.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
+Ogury passback for mobile sticky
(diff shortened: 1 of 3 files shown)


emma-imber (approved):

dskamiotis (approved):
Assumin this `Disable refreshing of the parent slot from dfpenv` is the latest change/commit.

Looks sensible and effective to me along with the code to set up the conditional passback

PR #1463, opened 2024-07-09
Refactor metrics bypassing

emma-imber:
## What does this change?
Refactors the metrics bypass functionality to make it easier to read and work with.

## Why?
At the moment, metrics are sent when either a `visibilitychange` or `pagehide` event listener fires. The event listeners are added when someone is in the dev environment (eg testing locally or on CODE) and when someone is in the sampling group. The `bypassCommercialMetricsSampling` function works by adding the event listeners when called.

This isn't what you would intuitively expect the `bypassCommercialMetricsSampling` function to do. It also means that listeners could be coming from DCR or from commercial, which might be harder to trace and debug.

Instead, this PR refactors the code to add listeners as part of the `initCommercialMetrics` function. This means listeners only get added in one place, at one time. Now, if someone is in the dev environment, in the sampling group, or calls the `bypassCommercialMetricsSampling` function, a property is set on the window, called `shouldSendCommercialMetrics`. This means we can set the value from DCR or the commercial code, and share the state.

Now when the event listener fires, we check if the `shouldSendCommercialMetrics` value on the window is true. If so, we send the metrics.

I think this implementation is easier to read, and should be easier to work with and debug in future. Hopefully it should also mean that the bypassing works from the commercial code as expected now.

### DIFF
--- a/src/core/send-commercial-metrics.ts
+++ b/src/core/send-commercial-metrics.ts
@@ -185,20 +185,17 @@ function gatherMetricsOnPageUnload(): void {
 }
 
 const listener = (e: Event): void => {
-	switch (e.type) {
-		case 'visibilitychange':
-			if (
-				document.visibilityState === 'hidden' &&
-				window.guardian.config.shouldSendCommercialMetrics === true
-			) {
+	if (window.guardian.config.shouldSendCommercialMetrics === true) {
+		switch (e.type) {
+			case 'visibilitychange':
+				if (document.visibilityState === 'hidden') {
+					gatherMetricsOnPageUnload();
+				}
+				return;
+			case 'pagehide':
 				gatherMetricsOnPageUnload();
-			}
-			return;
-		case 'pagehide':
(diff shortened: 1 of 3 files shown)


src/core/send-commercial-metrics.ts
```diff
@@ -187,12 +187,17 @@ function gatherMetricsOnPageUnload(): void {
...
+			if (
+				document.visibilityState === 'hidden' &&
+				window.guardian.config.shouldSendCommercialMetrics === true
+			) {
 				gatherMetricsOnPageUnload();
 			}
 			return;
 		case 'pagehide':
-			gatherMetricsOnPageUnload();
+			if (window.guardian.config.shouldSendCommercialMetrics === true) {
```
Jakeii, line 198: Could the if go around the switch statement?

Jakeii (approved):
Nice write up! much better way for bypassing to work👌

src/core/send-commercial-metrics.ts
emma-imber: Good shout! Done

PR #1464, opened 2024-07-10
Don't add advertisement labels to Hosted By ads served by opt out

emma-imber:
## What does this change?
Adds a class to opt out ads with the `HOSTEDBY` tag. If this class is present on an ad, we don't attach an ad label.

## Why?
We'd like to start serving more native templates on opt out, so we need to remove the label from certain templates.

### Before:

Screenshot 2024-07-10 at 16 00 49

### After:

Screenshot 2024-07-10 at 15 36 55

### DIFF
--- a/.changeset/wise-schools-bathe.md
+++ b/.changeset/wise-schools-bathe.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': patch
+---
+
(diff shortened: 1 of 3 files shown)


deedeeh (approved):
That looks good Emma 🚀

PR #1467, opened 2024-07-16
Integrate Magnite aka Rubicon prebid partner

deedeeh:
## What does this change?

This PR adds Magnite, previously named Rubicon and that's the name used in docs, as a prebid bidder.

It includes:
- `accountId`, `siteId` and `zoneId` as they are required for integration. Please check [Rubicon prebid docs](https://docs.prebid.org/dev-docs/bidders/rubicon.html) for more info.
- Bump guardian/Prebid.js [v8.52.0-1](https://www.npmjs.com/package/@guardian/prebid.js/v/8.52.0-1) for enabling Magnite bid adapter in our prebid.js fork https://github.com/guardian/Prebid.js/pull/149 
- 1% AB test to receive data and check if they are in the correct key/value pairs. You can opt-in by adding `#ab-PrebidMagnite=variant` at the end of the url.

Other related PRs to this work:

- Frontend where AB test was added: https://github.com/guardian/frontend/pull/27329
- Frontend https://github.com/guardian/frontend/pull/27335 for adding a switch. This one is already in PROD.

Here is the [spreadsheet](https://docs.google.com/spreadsheets/d/1cPnPQrnQNrOpuuIVO-niW1NZvyqJV5RgZjbeMV4Wtm8/edit?pli=1&gid=0#gid=0) with all the info regarding ad-sizes, siteId and zoneId based on the different regions.

In CODE or PROD the commercial switch and the AB test switch must be on and you opt in the test  and this is how you see `rubicon` as part of the prebid bidders in the commercial logger. 

I was able to test the switch locally and in CODE but the AB test I tested it just in CODE as there is an issue with opt in client-side tests locally. 

If you want to test locally checkout to this branch and comment out the logic related to `&&` ab test so you can check the request and response for rubicon based on the commercial switch. 

Screenshot 2024-07-23 at 12 42 30

Screenshot 2024-07-23 at 12 43 27

**Note:** Rubicon is the old bidder name but it's still used for the integration. Magnite is the current name so in the code you will see the two names as I am using rubicon when it is required for integration which is absolutely necessary for sending requests and getting responses and magnite is for anything else related to just our code. There will be an iteration/refactor in another PR to name everything in our codebase `rubicon` but we just need to get this out at the moment.

## Why?

We would like to add magnite (rubicon) as a prebid bidder behind a 1% AB test.


src/experiments/tests/prebid-magnite.ts
```diff
@@ -0,0 +1,28 @@
+import type { ABTest } from '@guardian/ab-core';
+
+export const prebidMagnite: ABTest = {
+	id: 'PrebidMagnite',
+	author: '@commercial-dev',
+	start: '2024-07-18',
+	expiry: '2024-09-30',
+	audience: 1 / 100,
+	audienceOffset: 0 / 100,
```
deedeeh, line 9: Just need to double check if the audienceOffset needs to be changed? As far as I remember this is to not have a user in multiple variants at the same time or separate variant and control audience. Please if someone can add to this.

dskamiotis: As far as Im aware, thats right, in that it may be to stop certain users cross over into multiple different test groups.
Im not sure how this would be needed in this instance?

deedeeh: Does that work the same way with server-side and client-side tests?

domlander: With server-side tests, the offset is chosen for you, as our test groups have no overlap (ignoring the 50% group). With client-side tests, you select a portion of the audience using the audienceOffset and it is helpful to avoid clashing with other client-side tests that are running. Looking at our chart in tools for visualising this, it looks like there aren't any clashes, so leaving audienceOffset at `0 / 100` here is fine 👍

src/lib/header-bidding/prebid/bid-config.ts
```diff
@@ -533,6 +536,20 @@ const kargoBidder: PrebidBidder = {
 	}),
 };
 
+const magniteBidder: PrebidBidder = {
+	//Rubicon is the old name for Magnite but it is still used for the integration
+	name: 'rubicon',
+	switchName: 'prebidMagnite',
+	bidParams: (
+		_slotId: string,
```
Jakeii, line 544: You can lose the underscore, it just make typescript ignore unused function arguments.

src/lib/header-bidding/prebid/magnite.ts
```diff
@@ -0,0 +1,190 @@
...
+					return 3471464;
+				}
+				if (containsMobileSticky(sizes)) {
+					return 3471468;
+				}
+				break;
+			default:
+				return -1;
+		}
+	}
```
Jakeii, line 149: This could be written with a bit less code by having the region check inside the size check:
```suggestion
	switch (getBreakpointKey()) {
		case 'D':
			if (containsMpuOrDmpu(sizes) || containsWS(sizes)) {
				if (isInUk()) {
					return ...'
				} else if (isInRow()) {
					...
				} else if (isInUsOrCa()) {
				    ...
				} else if (isInAuOrNz()) {
				 ...
				}
				
			}
			// top-above-nav on desktop
			if (
				containsLeaderboardOrBillboard(sizes) &&
				slotId === 'dfp-ad--top-above-nav'
			) {
				if (isInUk()) {
					return ...'
				} else if (isInRow()) {
					...
				} else if (isInUsOrCa()) {
				    ...
				} else if (isInAuOrNz()) {
				 ...
				}
			}
			// Fronts-banners on desktop
			if (
				containsBillboard(sizes) &&
				slotId.includes('fronts-banner')
			) {
				if (isInUk()) {
					return ...'
				} else if (isInRow()) {
					...
				} else if (isInUsOrCa()) {
				    ...
				} else if (isInAuOrNz()) {
				 ...
				}
			}
			break;
		case 'M':
			if (containsMpu(sizes) || containsPortraitInterstitial(sizes)) {
				if (isInUk()) {
					return ...'
				} else if (isInRow()) {
					...
				} else if (isInUsOrCa()) {
				    ...
				} else if (isInAuOrNz()) {
				 ...
				}
			}
			break;
		default:
			return -1;
	}
	
```

src/lib/header-bidding/prebid/bid-config.ts
deedeeh: I didn't know about it. Learning something new everyday! Done 952d54fce3f22bd81cf61c266c92c35b34d5cd15

src/experiments/tests/prebid-magnite.ts
deedeeh: That's a very useful explanation Dom! Every time I forget what audienceOffset do I will get back to this PR to check your comment 😄 Thank you.

Jakeii (approved):
Nice work!

PR #1469, opened 2024-07-16
update @guardian/* packages and to typescript 5.5

Jakeii:
## What does this change?
Update internal packages which are now using typescript@5.5

## Why?
Keep things up to date


arelra (approved):
✨

PR #1480, opened 2024-07-25
Fix husky hooks

Jakeii:
## What does this change?
For some reason I merged #1479 as it had all green ticks, but some lines in the husky hooks have been deprecated and no longer work https://github.com/typicode/husky/releases/tag/v9.1.1
and need to be removed

## Why?
Breaking CI


deedeeh (approved):
It's weird that they haven't be removed automatically because that's what's being said in the release notes under deprecation. Thanks for fixing it  👍🏼

PR #1481, opened 2024-07-25
spacefinder should avoid h3's

Jakeii:
## What does this change?
avoid h3 the same as we do for h2's

## Why?
I've spotted `h3`'s in the wild that are not being avoided since #1458, I think a `data-spacefinder-role="heading"` might be more suitable but can follow up with that.


deedeeh (approved):
Looks good 👍🏼 A follow up with your suggestion would be a better solution.

PR #1483, opened 2024-07-25
Rename spacefinder options to be easier to understand

Jakeii:
## Why?

`minAbove`/`Below` is hard to understand conceptually, even after appending `Slot` to the names it is still rather confusing, margins relative to the opponent make much more sense.

Same for the `minAbove`/`Below` of the article itself, `minDistanceFromTop`/`Bottom` make it much easier to understand what is going on.


deedeeh (approved):
Clearer naming makes our lives easier 🎉

arelra (approved):
Thanks for doing this @Jakeii much easier to understand

PR #1484, opened 2024-07-26
conditionally render ad labels with data-show-label attribute

dskamiotis:
## What does this change?

This PR is part of a bigger project to remove the Native ad-labels at inception from commercial templates: related PR.

Currently the logic which decides if an label should be rendered lies here in the Commercial bundle. This logic will applies a `data-label-show='true'` class attribute to the `adSlotNode`.

This will enable the CSS set out in [DCR](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/components/AdSlot.web.tsx#L118), to apply the ad label styling.

## Why?
The AdSlots are added to the article server-side rendered (in DCR). The CSS for rendering the Ad label is also server-side rendered and targets .ad-slot[data-label-show='true']:not(.ad-slot--interscroller)::before.

The commercial bundle dynamically adds the data-label-show='true' attribute, handled by the shouldRenderLabel function.

The "Advertisement" label has been removed from the Native ads in commercial-templates as part of [this PR](https://github.com/guardian/commercial-templates/pull/394).

When the commercial code detects sizeCallbacks[adSizes.fluid.toString()] (Native ads), it adds the class. However, there is one Native style ad slot where we do not want to add the Ad Label. Instead of removing the ad-slot--fluid class, this is handled in the shouldRenderLabel function.

Logic has been added to not add a label for these Native Format ID's defined in GAM.

| From GAM Native Format ID |
|------------|
|Screenshot 2024-08-02 at 10 41 04 | 
|Screenshot 2024-08-02 at 11 11 57 |

- 10077207, // CAPI_MULTIPLE_HOSTED
- 10069167, // CAPI_MULTIPLE_PAIDFOR
- 12317037, // EVENTS_MULTIPLE
- 10070487, // CAPI_SINGLE_PAIDFOR
- 10063287, // MANUAL_MULTIPLE

It is only the in house fabric we add that we want labels to be added to.
## Images

| Before | After |
|----------|----------|
|Screenshot 2024-07-26 at 14 39 29|Screenshot 2024-07-26 at 14 35 17 | 

### Example of merchandising advert not rendering an ad label with 

| Merchandising slot | adSlotNode Native ID CAPI_SINGLE_PAIDFOR |
|----------|----------|
| Screenshot 2024-08-06 at 14 12 20| Screenshot 2024-08-06 at 14 13 42|

### DIFF
--- a/src/core/messenger/background.ts
+++ b/src/core/messenger/background.ts
@@ -184,6 +184,8 @@ const setupBackground = async (
 		specs.scrollType,
 	);
 
+	const nativeTemplateId = 11885667;
+
 	return fastdom.mutate(() => {
 		setBackgroundStyles(specs, background);
 
@@ -210,7 +212,7 @@ const setupBackground = async (
 				adSlot.style.position = 'relative';
 			}
 
-			void renderAdvertLabel(adSlot);
+			void renderAdvertLabel(adSlot, nativeTemplateId);
 			void renderStickyScrollForMoreLabel(backgroundParent);
 
 			isDCR && renderBottomLine(background, backgroundParent);
--- a/src/core/messenger/background.ts
+++ b/src/core/messenger/background.ts
@@ -184,7 +184,7 @@ const setupBackground = async (
 		specs.scrollType,
 	);
 
-	const nativeTemplateId = 11885667;
+	const interscrollerTemplateId = 11885667;
 
 	return fastdom.mutate(() => {
 		setBackgroundStyles(specs, background);
@@ -212,7 +212,7 @@ const setupBackground = async (
 				adSlot.style.position = 'relative';
 			}
 
-			void renderAdvertLabel(adSlot, nativeTemplateId);
+			void renderAdvertLabel(adSlot, interscrollerTemplateId);
 			void renderStickyScrollForMoreLabel(backgroundParent);
 
 			isDCR && renderBottomLine(background, backgroundParent);
--- a/src/events/render-advert-label.spec.ts
+++ b/src/events/render-advert-label.spec.ts
@@ -61,11 +61,11 @@ describe('shouldRenderLabel', () => {
 		});
 	});
 
-	it('does NOT render an ad label for fluid ads', async () => {
+	it('renders an ad label for fluid ads', async () => {
 		createAd(adverts['fluid']);
 		return renderAdvertLabel(getAd()).then(() => {
 			const ad = getAd();
-			expect(shouldRenderLabel(ad)).toBeFalsy();
+			expect(shouldRenderLabel(ad)).toBeTruthy();
 		});
 	});
 
--- a/src/events/render-advert-label.spec.ts
+++ b/src/events/render-advert-label.spec.ts
@@ -1,4 +1,8 @@
-import { renderAdvertLabel, shouldRenderLabel } from './render-advert-label';
+import {
+	renderAdvertLabel,
+	shouldRenderLabel,
+	templatesWithoutLabels,
+} from './render-advert-label';
 
 jest.mock('lib/commercial-features', () => ({
 	commercialFeatures: {},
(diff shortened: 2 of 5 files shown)


src/events/render-advert-label.ts
```diff
@@ -108,6 +100,7 @@ const renderAdvertLabel = (adSlotNode: HTMLElement): Promise<Promise<void>> => {
 
 			return fastdom.mutate(() => {
 				adSlotNode.setAttribute('data-label-show', 'true');
+				adSlotNode.setAttribute('my-test-one', 'true');
```
arelra, line 103: Don't think we want to keep this?

src/events/render-advert-label.ts
```diff
@@ -8,14 +8,6 @@ import fastdom from 'utils/fastdom-promise';
 import crossIcon from '../../static/svg/icon/cross.svg';
 
 const shouldRenderLabel = (adSlotNode: HTMLElement): boolean => {
-	if (
-		adSlotNode.classList.contains('ad-slot--fluid') &&
-		!adSlotNode.classList.contains('ad-slot--interscroller') &&
-		!adSlotNode.classList.contains('ad-slot--article-end')
```
arelra, line 14: Did we mean to remove `ad-slot--article-end` ?

Jakeii:
I think you're before/after pictures are reversed?

Also is the case for no label at all for merch ads still covered?

Screenshot 2024-07-31 at 16 51 13

src/events/render-advert-label.ts
dskamiotis: I thought in this case the logic for this case was tied into the logic for first condition `adSlotNode.classList.contains('ad-slot--fluid')`?
So I think this can be re added added to return false.

arelra: Ah so now we don't want to render the label for any fluid?

dskamiotis:
> Wrt the before and after pictures, I assume with this change the screenshots would be visually the same except now the commercial code is adding the ad labels?



> Wrt the before and after pictures, I assume with this change the screenshots would be visually the same except now the commercial code is adding the ad labels?

Not the clearest screenshots as the labels are small, Im trying to demo the Advertisement Label being centred - but as Jake said I put them back to fromt (so before labels left aligned and after centre :))

src/events/render-advert-label.ts
```diff
@@ -7,15 +7,25 @@ import { getCookie } from '@guardian/libs';
...
+): boolean => {
 	if (
-		adSlotNode.classList.contains('ad-slot--fluid') &&
-		!adSlotNode.classList.contains('ad-slot--interscroller') &&
-		!adSlotNode.classList.contains('ad-slot--article-end')
+		creativeTemplateId === CAPI_SINGLE_PAIDFOR ||
+		creativeTemplateId === CAPI_MULTIPLE_PAIDFOR ||
+		creativeTemplateId === EVENTS_API_MULTIPLE_STYLE ||
+		creativeTemplateId === CAPI_MULTIPLE_HOSTED ||
+		creativeTemplateId === MANUAL_MULTIPLE
```
Jakeii, line 25:
```suggestion
		templatesWithoutLabels.includes(creativeTemplateId)
```

src/events/render-advert-label.ts
```diff
@@ -7,15 +7,25 @@ import { getCookie } from '@guardian/libs';
...
 	) {
-		return false;
+		return true;
```
Jakeii, line 27:
```suggestion
		return false;
```

src/events/render-advert.ts
```diff
@@ -214,7 +215,9 @@ const renderAdvert = (
 					: Promise.resolve();
 
 			return callSizeCallback()
-				.then(() => renderAdvertLabel(advert.node))
+				.then(() =>
+					renderAdvertLabel(advert.node, creativeTemplateId ?? 0),
```
emma-imber, line 219: We also call `renderAdvertLabel` in the interscroller code, in `background.ts`, so the parameter will need to be added there too

src/events/render-advert-label.spec.ts
```diff
@@ -61,11 +61,11 @@ describe('shouldRenderLabel', () => {
 		});
 	});
 
-	it('does NOT render an ad label for fluid ads', async () => {
+	it('renders an ad label for fluid ads', async () => {
```
emma-imber, line 64: Instead of testing for fluid ads, could we check that a label should render for a fabric ad, and shouldn't render for a manual multiple, for example? That would make the test more relevant to the new logic

src/events/render-advert-label.ts
```diff
@@ -89,9 +99,13 @@ const createAdTestCookieRemovalLink = (): HTMLElement => {
...
 
-const renderAdvertLabel = (adSlotNode: HTMLElement): Promise<Promise<void>> => {
+const renderAdvertLabel = (
+	adSlotNode: HTMLElement,
+	creativeTemplateId: number,
+): Promise<Promise<void>> => {
 	return fastdom.measure(() => {
-		if (shouldRenderLabel(adSlotNode)) {
+		if (shouldRenderLabel(adSlotNode, creativeTemplateId)) {
+			///here
```
emma-imber, line 108:
```suggestion
```

src/events/render-advert-label.ts
```diff
@@ -7,15 +7,25 @@ import { getCookie } from '@guardian/libs';
 import fastdom from 'utils/fastdom-promise';
 import crossIcon from '../../static/svg/icon/cross.svg';
 
-const shouldRenderLabel = (adSlotNode: HTMLElement): boolean => {
+const CAPI_SINGLE_PAIDFOR = 10077207;
+const CAPI_MULTIPLE_PAIDFOR = 10069167;
+const EVENTS_API_MULTIPLE_STYLE = 12317037;
+const CAPI_MULTIPLE_HOSTED = 10070487;
+const MANUAL_MULTIPLE = 10063287;
```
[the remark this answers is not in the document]
emma-imber: Also there's a slight mix up in the template ID values here. CAPI Single PaidFor and CAPI Multiple Hosted are actually the other way around:
```
const templatesWithoutLabels = [
	10077207, // CAPI_MULTIPLE_HOSTED
	10069167, // CAPI_MULTIPLE_PAIDFOR
	12317037, // EVENTS_MULTIPLE
	10070487, // CAPI_SINGLE_PAIDFOR
	10063287 // MANUAL_MULTIPLE
];
```

src/events/render-advert-label.spec.ts
```diff
@@ -45,75 +49,86 @@ describe('shouldRenderLabel', () => {
...
 			const ad = getAd();
-			expect(shouldRenderLabel(ad)).toBeTruthy();
+			expect(shouldRenderLabel(ad, testCreativeTemplateId)).toBeTruthy();
 		});
 	});
 
 	it('renders an ad label for interscroller ads', async () => {
 		createAd(adverts['interscroller']);
-		return renderAdvertLabel(getAd()).then(() => {
+		return renderAdvertLabel(getAd(), testCreativeTemplateId).then(() => {
```
emma-imber, line 64: It would be more accurate to use the interscroller template ID for this test, since the test is specifically about the interscroller

src/core/messenger/background.ts
```diff
@@ -184,6 +184,8 @@ const setupBackground = async (
 		specs.scrollType,
 	);
 
+	const nativeTemplateId = 11885667;
```
emma-imber, line 187:
```suggestion
	const interscrollerTemplateId = 11885667;
```
Doesn't hurt to be specific here!

emma-imber (commented):
Just a few minor suggestions!

dskamiotis:
> Just a few minor suggestions!

Thanks @emma-imber nice refinements 👍

src/events/render-advert-label.ts
```diff
@@ -7,15 +7,21 @@ import { getCookie } from '@guardian/libs';
...
+	10077207, // CAPI_MULTIPLE_HOSTED
+	10069167, // CAPI_MULTIPLE_PAIDFOR
+	12317037, // EVENTS_MULTIPLE
+	10070487, // CAPI_SINGLE_PAIDFOR
+	10063287, // MANUAL_MULTIPLE
+];
+
+const shouldRenderLabel = (
+	adSlotNode: HTMLElement,
+	creativeTemplateId: number,
```
Jakeii, line 20: I think this should be optional because most adverts don't even have a `creativeTemplateId`, coercing it to `0` below doesn't feel right.

dskamiotis: I was thinking the same when I was updating the tests with new paramater.

src/events/render-advert.ts
Jakeii, line 219:
```suggestion
					renderAdvertLabel(advert.node, creativeTemplateId),
```

src/events/render-advert.ts
```diff
@@ -187,6 +187,7 @@ const renderAdvert = (
 
 	return getAdIframe(advert.node)
 		.then((isRendered) => {
+			const creativeTemplateId = slotRenderEndedEvent.creativeTemplateId;
```
Jakeii, line 190:
```suggestion
			const creativeTemplateId = slotRenderEndedEvent.creativeTemplateId ?? undefined;
```

src/events/render-advert-label.ts
```diff
@@ -7,15 +7,21 @@ import { getCookie } from '@guardian/libs';
...
+	12317037, // EVENTS_MULTIPLE
+	10070487, // CAPI_SINGLE_PAIDFOR
+	10063287, // MANUAL_MULTIPLE
+];
+
+const shouldRenderLabel = (
+	adSlotNode: HTMLElement,
+	creativeTemplateId?: number,
+): boolean => {
+	if (templatesWithoutLabels.includes(creativeTemplateId ?? 0)) {
```
Jakeii, line 22:
```suggestion
	if (creativeTemplateId && templatesWithoutLabels.includes(creativeTemplateId)) {
```

Jakeii (approved):
I have some last syntax suggestions but otherwise I think good to go.

PR #1485, opened 2024-07-29
Remove Magnite prebid AB test 

deedeeh:
## What does this change?

This PR removes the Magnite prebid AB test as we have a confirmation that we are delivering the correct key/value pairs and getting impressions so now we can safely rollout. 

## Why?

We don't need the AB test anymore as it's only reason for Magnite was for testing purposes.


arelra (approved):

src/lib/header-bidding/utils.ts
```diff
@@ -201,8 +199,7 @@ export const shouldIncludeSmart = (): boolean => isInUk() || isInRow();
 export const shouldIncludeKargo = (): boolean => isInUsa();
 
 export const shouldIncludeMagnite = (): boolean =>
-	!!window.guardian.config.switches.prebidMagnite &&
-	isUserInVariant(prebidMagnite, 'variant');
+	!!window.guardian.config.switches.prebidMagnite;
```
arelra, line 202: This removes the switch too - I assume thats intentional.

PR #1497, opened 2024-08-06
Record interscroller progress with logs and metrics

Jakeii:
## What does this change?
Log interscroller via logs as well as via commercial metrics

## Why?
We're not getting the volumes of progress data that we expect, this could be due to the way that metrics is initialized in DCR, via an island that may not run due to the adaptive site mechanism.


emma-imber (approved):
Looks good!

PR #1507, opened 2024-08-06
Refine dependabot.yml

Jakeii:
## What does this change?
Move some of the groups into the devDependency group

## Why?
There's still an unwieldy amount of PRs being raised by dependabot


deedeeh (approved):
Lets see if that will be more suitable for us. Thanks @Jakeii for doing this 🙏🏻

PR #1514, opened 2024-08-06
Remove ophan-tracker-js from commercial

arelra:
WIP

## What does this change?

Remove Ophan from commercial

## Why?

Commercial shouldn't need `ophan-tracker-js` this is probably a hangover from Frontend.

We still have Ophan related functionality:

- Ophan `pageViewId` and `browserId` from `window.guardian.ophan` _and_ `window.guardian.config.ophan` - pick one!
- Ophan is a required property to the [AB API](https://github.com/guardian/commercial/blob/88665db419b066dc68bdff52eff8474bf2456638/src/experiments/ab.ts#L43-L65), we grab Ophan from `window.guardian.ophan` to pass to it - how is this used?


Jakeii:
I looked into the ophan requirement for ab-core, looks like it's just to track these impression and success callbacks that we never use.

Screenshot 2024-08-07 at 10 03 34


https://github.com/guardian/csnx/tree/main/libs/@guardian/ab-core#ab-test-definition

Jakeii (approved):
Nice! 🧹

arelra:
Follow up actions:

- align on `window.guardian.ophan` and `window.guardian.config.ophan` 
- investigate if we can remove `impression` and `success` callbacks from [ab-core](https://github.com/guardian/csnx/tree/main/libs/%40guardian/ab-core)

PR #1516, opened 2024-08-08
Remove `inline-merchandising`/`im` slot

deedeeh:
## What does this change?

This PR removes the `inline-merchandising` slot, aka `im`,  from theguardian.com 

Based on the code removal from Commercial, DCR and Frontend you shouldn't see:
1.  An `im` slot in the DOM.
2. "Inline Merchandising Slot" in Frontend tools -> Commercial tools -> under Targeting
3. `guardian.config.page.hasInlineMerchandise`

Other PRs related to this work https://github.com/guardian/frontend/pull/27394 and https://github.com/guardian/dotcom-rendering/pull/12104

This change has been tested locally and in CODE.

## Why?

Consolidate our ad slot supply, ensuring the available slots are delivering the best revenue performance.


src/display/display-lazy-ads.ts
```diff
@@ -3,7 +3,7 @@ import { dfpEnv } from '../lib/dfp/dfp-env';
 import { enableLazyLoad } from './lazy-load';
 import { loadAdvert } from './load-advert';
 
-const instantLoadAdvertIds = ['dfp-ad--im'];
+const instantLoadAdvertIds: string[] = [];
```
deedeeh, line 6: I can see this variable is being used on line 14 that is why I added an empty array but not 100% sure if we still need it or we can pass something else to partition as a second parameter?

arelra: If there is no other slot using this I think we can remove this feature!

deedeeh: Thank you Ravi for reviewing this and checking the code in our remove `im` catchup. This is now done 27a7758bffa8311c345c222a3702e50094033757

src/insert/spacefinder/article.ts
```diff
@@ -293,10 +270,6 @@ const init = async (fillAdSlot: FillAdSlot): Promise<void> => {
 	// We add the first inline ad before finding a place for an im so that the inline1 doesn't get pushed down
 	await addFirstInlineAd(fillAdSlot);
 
-	if (window.guardian.config.page.hasInlineMerchandise) {
-		await attemptToAddInlineMerchAd(fillAdSlot);
-	}
-
 	// Add the other inline slots after the im, so that there is space left for the im ad
 	await addSubsequentInlineAds(fillAdSlot);
```
Jakeii, line 274: I think the first and subsequent passes can be combined again now

emma-imber: The consentless code also uses this `addSubsequentInlineAds` function, and we need to keep the first and subsequent passes separate to fix the bug I'm currently looking at. Could we keep them separate for now?

deedeeh: Yes like Emma said it's used by consentless so I think keeping them separate for now is a good idea until a refactor in a separate PR maybe.

src/insert/spacefinder/article.ts
```diff
@@ -258,29 +258,6 @@ const addSubsequentInlineAds = (fillSlot: FillAdSlot): Promise<boolean> => {
...
-
-	return spaceFiller.fillSpace(rules.inlineMerchandising, insertAds, {
-		waitForImages: true,
-		waitForInteractives: true,
-		pass: 'im',
-	});
-};
-
 /**
  * Init all the article body adverts, including `im` and `carrot`
```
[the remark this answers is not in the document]
deedeeh: It's easy to miss comments 😅 Thank you Dom! Done 897cff10548ee718c52522b6711cd60c58ccd55e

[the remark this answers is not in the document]
domlander: Thanks Dina. I'm so delighted to see the end of the `im` slot that I had to take a look at the PR!

src/insert/spacefinder/article.ts
emma-imber: Ok I've done some testing, I was wrong and we can merge these back together. I will pick it up as a separate follow up PR so that we don't delay releasing this!

emma-imber (approved):
Great to see this getting deleted!

src/insert/spacefinder/article.ts
deedeeh: Ok Emma sounds like a plan. Thank you!

PR #1517, opened 2024-08-08
Add Spacefinder inline1 desktop margin changes behind a 0% test

emma-imber:
## What does this change?
Adds some potential spacefinder optimisations for the desktop inline1 slot behind a 0% test. This rule set has some pretty large margins that we're convinced need to be so cautious. We want to AB test these changes to see if they boost ad ratio, so we're popping them behind a 0% test for now.

More detailed examples of each change are given below.

### Rich links
These are currently avoided in exactly the same way as an inline image, which seems overly strict. By relaxing the rules we can improve the placement of inline1 significantly and it still looks fine.

#### Before:
No inline1 appeared on this whole article because the paragraphs were considered too close to the rich link
Screenshot 2024-08-08 at 12 28 46

#### After:
Inline1 appears, and I think this looks absolutely fine with the rich link still
Screenshot 2024-08-08 at 12 30 44

### Thumbnails
Rule changes now mean that ads can appear underneath thumbnail images. They still can't overlap, but it allows inline1 to appear higher up on a lot of articles - especially recipes which use thumbnail images a lot.

#### Before:
Screenshot 2024-08-08 at 12 22 03

#### After:
Screenshot 2024-08-08 at 12 21 44

### Immersives
At the moment we set a 600px margin on immersive elements (images that span into the right column). This means that on picture essays, we don't really get inline 1 ads. By reducing the margin, we should be able to introduce inline1 ads into some picture essays.

#### Before:
Screenshot 2024-08-08 at 16 08 21

#### After:
Screenshot 2024-08-08 at 16 07 59

### Inlines
By reducing the margin from 400 to 200 we can get a few more inline1 ads, without them being overly close to other inline elements.

#### Before:
The newsletter sign up block is too close for an inline1 to be inserted on this article.
Screenshot 2024-08-08 at 18 13 41

#### After:
Screenshot 2024-08-08 at 18 13 22

### DIFF
--- a/src/insert/spacefinder/rules.ts
+++ b/src/insert/spacefinder/rules.ts
@@ -1,5 +1,7 @@
 import { adSizes } from 'core';
 import { adSlotContainerClass } from 'core/create-ad-slot';
+import { isUserInVariant } from 'experiments/ab';
+import { optimiseSpacefinderInline } from 'experiments/tests/optimise-spacefinder-inline';
 import type { OpponentSelectorRules, SpacefinderRules } from './spacefinder';
 import { isInHighValueSection } from './utils';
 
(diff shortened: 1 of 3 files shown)


src/insert/spacefinder/rules.ts
```diff
@@ -28,19 +30,21 @@ const minDistanceBetweenInlineAds = isInHighValueSection ? 500 : 750;
 
 const candidateSelector = ':scope > p, [data-spacefinder-role="nested"] > p';
 
+const leftColumnOpponentSelector = ['richLink', 'thumbnail']
+	.map((role) => `[data-spacefinder-role="${role}"]`)
```
emma-imber, line 34: Splitting out richlink and thumbnail allows us to apply more specific rules here, instead of treating them the same as inline elements

Jakeii (approved):
Should be great for ad ratio! 🤑

PR #1526, opened 2024-08-14
Add a bigger rail min above for consentless ads

emma-imber:
## What does this change?
Adds a bigger `desktopRightRailMinAbove` for consentless ads.

## Why?
We've been seeing opt out inline2 ads overlapping the most viewed column, and this should fix it. The overlapping happens whenever an ad doesn't fill inline1, and the inline1 slot collapses. This happens pretty frequently for consentless ads.

### Before
Screenshot 2024-08-14 at 10 11 24

### After
Screenshot 2024-08-14 at 10 11 07

### DIFF
--- a/.changeset/shy-crabs-travel.md
+++ b/.changeset/shy-crabs-travel.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': patch
+---
+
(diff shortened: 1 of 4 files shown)


deedeeh (approved):
Looks good Emma 🚀

PR #1528, opened 2024-08-14
Add pageviewId to video progress logs behind consent check

Jakeii:
## Why?
There are way more progress reports than impressions, hopefully recording the `pageviewId` will uncover if we're getting duplicates

This code only executes when googletag is already loaded but for safety/paranoia have added the same consent check that commercial metrics uses.


emma-imber (approved):
Will be interesting to see if we have duplicates!

PR #1532, opened 2024-08-15
Merge first and subsequent inline ad functions

emma-imber:
## What does this change?
Merge the `addFirstInlineAd` and `addSubsequentInlineAds` functions into one function.

## Why?
Now that inline-merch has been deleted, we no longer need to separate the code to add inline ads into inline1 and subsequent inlines. This results in less lines of code and things being a bit easier to read.


Jakeii (approved):
Nice 🧹

PR #1533, opened 2024-08-19
Only execute sendProgress once

emma-imber:
## What does this change?
Wraps `sendProgress` in `once()`.

## Why?
This should mean it only gets executed once. At the moment, we think it can be called twice by the two different event listeners. Hopefully this should reduce the duplicate log entries we see.


Jakeii (approved):
speedy work! 🚀

PR #1541, opened 2024-08-20
remove Raven client and refactor related functions

dskamiotis:
## What does this change?

This PR updates the error reporting mechanism in the commercial module by removing the Raven client and transitioning to Sentry's new reporting system. We now use the window.guardian.modules.sentry object to directly report errors to Sentry.

Changes Made:
- Removed Raven Client:
-- The Raven client, previously used for error reporting, has been removed from the codebase. This is part of our move to streamline error handling and leverage the latest tools.

- Integrated Sentry Reporting:
-- Replaced the Raven-based error reporting code with direct calls to the Sentry API.
Errors are now reported using the reportError function accessed via window.guardian.modules.sentry.

## Why?
Unified Error Reporting: 
- The transition to using window.guardian.modules.sentry aligns with our strategy to centralize error reporting and utilize Sentry's updated features and capabilities.

Improved Error Management:
- By moving away from Raven, we benefit from Sentry's enhanced functionality, which includes better error tracking, improved performance, and more detailed reporting.

## Testing
The new error reporting approach has been tested to ensure that errors are correctly reported to Sentry and that the transition does not affect the application's error handling.
Using the following code to test the error being picked up by forcing it onto the window
This triggers a test error and reports it to Sentry if the environment is not production to ensure the function is working:

```ts
if (process.env.NODE_ENV !== 'production') {
	try {
		throw new Error('>>>>Test error for sentry');
	} catch (err) {
		reportError(err, 'test-feature');
	}
}
```

Screenshot 2024-08-23 at 13 32 03


Jakeii:
I think this could be easier to review if you exported the window reportError from the old module `report-error.ts`, then not so many files would need to be updated (and reviewed)!

```
  export const reportError = window.guardian.modules.sentry.reportError;
```

dskamiotis:
> I think this could be easier to review if you exported the window reportError from the old module `report-error.ts`, then not so many files would need to be updated (and reviewed)!
> 
> ```
>   export const reportError = window.guardian.modules.sentry.reportError;
> ```

Sure thing 👍

src/insert/fill-dynamic-advert-slot.ts
```diff
@@ -29,14 +29,7 @@ const fillDynamicAdSlot = (
...
 				log('commercial', errorMessage);
-				reportError(
-					Error(errorMessage),
-					{
-						feature: 'commercial',
-						slotId: adSlot.id,
-					},
-					false,
-				);
+				reportError(Error(errorMessage), 'commercial');
```
Jakeii, line 32: Does DCR raven not support sending additional metadata?

SiAdcock: Good spot. It looks like [DCR is missing support](https://github.com/guardian/dotcom-rendering/blob/03e25eed03f1af91a99989a002af3e626c333564/dotcom-rendering/src/client/sentryLoader/sentry.ts#L58) for [custom tags](https://docs.sentry.io/platforms/javascript/enriching-events/tags/) and only allows us to pass the `feature`. Let's bring this up with WebX as adding custom tag support to DCR at the capture level would probably be desirable for them too.

Example error with custom tag `slotId`:

https://the-guardian.sentry.io/issues/5666660955/?project=35463&query=is%3Aunresolved+feature%3Acommercial&referrer=issue-stream&statsPeriod=14d&stream_index=0

src/init/consented/dfp-listeners.ts
```diff
@@ -7,14 +6,8 @@ const initDfpListeners = (): Promise<void> => {
...
-		pubads.addEventListener(
-			'slotRenderEnded',
-			raven.wrap<typeof onSlotRender>(onSlotRender),
-		);
-		pubads.addEventListener(
-			'slotOnload',
-			raven.wrap<typeof onSlotLoad>(onSlotLoad),
-		);
+		pubads.addEventListener('slotRenderEnded', onSlotRender);
+		pubads.addEventListener('slotOnload', onSlotLoad);
```
SiAdcock, line 10: I expect we'd still like these errors to be reportable to Sentry. What's the best way to achieve this in an idiomatic Sentry way?

src/insert/fill-dynamic-advert-slot.ts
SiAdcock: Are we also happy to lose the [`shouldThrow`](https://github.com/guardian/commercial/blob/main/src/utils/report-error.ts#L19) flag?

src/core/messenger.ts
```diff
@@ -514,7 +510,6 @@ export const init = (
 	>,
 	errorHandler: MessengerErrorHandler,
```
SiAdcock, line 511: Do we still need to accept an `errorHandler` as a parameter here? From what I can see, [we always pass](https://github.com/search?q=org%3Aguardian+initmessenger&type=code) `reportError` from `utils/report-error`, so I'm not sure what this abstraction gives us. Is it just for testing purposes?

src/core/messenger.ts
```diff
@@ -374,9 +372,7 @@ const onMessage = async (event: MessageEvent): Promise<void> => {
 				respond(message.id, event.source, null, response);
 			})
 			.catch((ex: Error) => {
-				reportError(ex, {
-					feature: 'native-ads',
-				});
+				reportError(ex, 'native-ads');
```
SiAdcock, line 375: Note: we are losing the [`shouldThrow`](https://github.com/guardian/commercial/blob/main/src/utils/report-error.ts#L19) flag. Therefore `reportError` will no longer automatically re-throw this error, it won't stop execution and it won't report the error to browser console.

src/define/create-advert.ts
```diff
@@ -21,14 +21,7 @@ const createAdvert = (
 		log('commercial', errMsg);
 
 		if (!navigator.userAgent.includes('DuckDuckGo')) {
-			reportError(
-				new Error(errMsg),
-				{
-					feature: 'commercial',
-				},
-				false,
-				1 / 100,
```
SiAdcock, line 30: Note: previously this error was being shown to 1% of users who were served the Sentry client

src/init/consented/messenger.ts
```diff
@@ -12,7 +12,6 @@ import { init as resize } from 'core/messenger/resize';
 import { init as scroll } from 'core/messenger/scroll';
 import { init as type } from 'core/messenger/type';
 import { init as viewport } from 'core/messenger/viewport';
-import { reportError } from 'utils/report-error';
```
SiAdcock, line 15: We also need to remove the usage of this function further down

src/init/consented/dynamic-ad-slots.ts
```diff
@@ -23,14 +23,11 @@ const dynamicAdSlotModules: Modules = [
...
-		dynamicAdSlotModules.map(async ([name, init]) => {
+		dynamicAdSlotModules.map(async ([, init]) => {
 			try {
 				await init();
 			} catch (error) {
-				reportError(error, {
-					feature: 'commercial',
-					tag: name,
-				});
+				reportError(error, 'commercial');
```
SiAdcock, line 30: This error will no longer be re-thrown

src/init/consentless.ts
```diff
@@ -10,7 +10,6 @@ import { initConsentless } from 'init/consentless/prepare-ootag';
 import { reloadPageOnConsentChange } from 'init/shared/reload-page-on-consent-change';
 import { init as setAdTestCookie } from 'init/shared/set-adtest-cookie';
 import { init as setAdTestInLabelsCookie } from 'init/shared/set-adtest-in-labels-cookie';
-import { reportError } from 'utils/report-error';
```
SiAdcock, line 13: We also need to remove the usage of this function further down

src/lib/cookies.js
```diff
@@ -42,11 +42,17 @@ const addCookie = (name, value, daysToLive, isCrossSubdomain = false) => {
 	const expires = new Date();
 
 	if (!isValidCookieValue(name) || !isValidCookieValue(value)) {
-		reportError(
-			new Error(`${ERR_INVALID_COOKIE_NAME} .${name}=${value}`),
-			{},
-			false,
-		);
+		if (window.guardian?.modules?.sentry?.reportError) {
```
SiAdcock, line 45: This is the only time we're making this check. DCR goes to some lengths to ensure this function is available and stubbed before Senty is loaded. Is it necessary to also make this check here?

src/utils/report-error.spec.ts
```diff
@@ -22,34 +11,22 @@ describe('report-error', () => {
...
-			error,
-			ravenMetaData,
-		);
-	});
+	// temporarily disabled due to flakiness
+	// test('Does throw an error', () => {
+	// 	expect(() => {
+	// 		reportError(error, tags);
+	// 	}).toThrowError(error);
+	// });
```
SiAdcock, line 27: This is an expected change of behaviour because the DCR implementation doesn't automatically re-throw the error. We should either delete this test or make a decision to implement the automatic re-throw in DCR.

src/utils/report-error.spec.ts
```diff
@@ -22,34 +11,22 @@ describe('report-error', () => {
...
 
 	test('Applies a sampling rate that prevents a sample of errors being reporting', () => {
-		reportError(error, tags, false, 1 / 100);
-
-		expect(fakeRaven.captureException).not.toHaveBeenCalled();
+		reportError(error, tags);
 	});
```
SiAdcock, line 31: Again, sampling functionality is not available in DCR. Either we implement sampling there or delete this test.

src/utils/report-error.ts
```diff
@@ -1,44 +1,4 @@
...
-		raven.captureException(localError, { tags });
-	}
-	if (shouldThrow) {
-		localError.reported = true;
-		throw localError;
-	}
-};
-
-export { convertError, reportError };
+// The report error function replace the older Raven based error reporting with the new Sentry based error reporting.
```
SiAdcock, line 1: I think the first line of this comment is unnecessary. In 6 months nobody will remember what Raven was.

src/init/consented/dfp-listeners.ts
SiAdcock: Sentry has [removed the `wrap` export](https://docs.sentry.io/platforms/javascript/migration/v7-to-v8/#removal-of-wrap-export), with no replacement API. Perhaps we could recreate something similar in `utils/report-error`

```js
// not tested!
export const wrapWithErrorReporting = (fn, tags) => {
  return function () {
    try {
      fn(arguments)
    } catch (e) {
      reportError(e, tags)
    }
}
```

src/insert/fill-dynamic-advert-slot.ts
SiAdcock: We've now added support for additional metadata in guardian/dotcom-rendering#12280

Jakeii:
Still need to add the tags back to the error reports? Now that you've added support for them to the DCR `reportError` function

dskamiotis:
> Still need to add the tags back to the error reports? Now that you've added support for them to the DCR `reportError` function

done 👍

src/events/on-slot-render.ts
```diff
@@ -6,30 +6,14 @@ import { getAdvertById } from '../lib/dfp/get-advert-by-id';
...
 		reportError(
 			new Error('dfp returned an empty ad response'),
-			{
-				feature: 'commercial',
-				adUnit: adUnitPath,
-				adSlot: adSlotId,
-				adKeywords,
-			},
-			false,
+			'commercial',
```
Jakeii, line 16: still missing feature, adUnit, adSlot etc. here

src/insert/fill-dynamic-advert-slot.ts
Jakeii: @dskamiotis also here, probs want slotId

src/lib/dfp/non-refreshable-line-items.ts
```diff
@@ -42,14 +42,7 @@ export const fetchNonRefreshableLineItemIds = async (): Promise<number[]> => {
...
 	const error = Error('Failed to fetch non-refreshable line items');
-	reportError(
-		error,
-		{
-			feature: 'commercial',
-			status: String(response.status),
-		},
-		false,
-	);
+	reportError(error, 'commercial');
```
Jakeii, line 45: status here

src/utils/report-error.ts
```diff
@@ -1,44 +1,20 @@
...
-	return new Error(String(err));
+const wrapWithErrorReporting = <T>(
+	fn: ErrorReportingFunction<T>,
+	tags: Record<string, string> = {},
+): ErrorReportingFunction<T> => {
+	return function (event: T) {
+		try {
+			fn(event);
+		} catch (e) {
+			reportError(e, JSON.stringify(tags));
```
arelra, line 15: We could be a little defensive here and check if `window.guardian.modules.sentry.reportError` is defined before invoking it.

Depending on the timing of how scripts are loaded it is possible (although unlikely) that it does not get initialised before commercial boots.

arelra, line 15: I think this invocation is incorrect?

The method signature for `reportError` is 

```js
					reportError: (
						error: unknown,
						feature: string,
						tags?: Record<string, string>,
					) => void;
```

arelra: I think we should update this module to wrap `reportError` with an internal function rather than a direct reference at the module level which will only resolve once.

This will allow us to get the reference the function on each invocation and also check if it is defined.

src/events/on-slot-render.ts
```diff
@@ -20,16 +20,15 @@ const reportEmptyResponse = (
 			? event.slot.getTargeting('k')
 			: [];
 		const adKeywords = adTargetingKValues.join(', ');
-
 		reportError(
 			new Error('dfp returned an empty ad response'),
+			'commercial',
 			{
 				feature: 'commercial',
```
arelra, line 27: Can remove this as the feature is now passed as the second variable

src/init/consented.ts
```diff
@@ -121,18 +120,15 @@ const bootCommercial = async (): Promise<void> => {
...
-					EventTimer.get().mark('commercialBootStart');
-				},
-			],
+			'ga-user-timing-commercial-start',
+			function runTrackPerformance() {
+				EventTimer.get().mark('commercialStart');
+				EventTimer.get().mark('commercialBootStart');
+			},
 		],
-		tags,
```
arelra, line 134: Do we ned to pass through the tags here?

src/init/consented.ts
```diff
@@ -150,7 +146,7 @@ const bootCommercial = async (): Promise<void> => {
 		await Promise.all(promises).then(recordCommercialMetrics);
 	} catch (error) {
 		// report async errors in bootCommercial to Sentry with the commercial feature tag
-		reportError(error, tags, false);
```
arelra, line 153: tags here

src/init/consented/dynamic-ad-slots.ts
```diff
@@ -23,14 +23,11 @@ const dynamicAdSlotModules: Modules = [
...
 export const initDynamicAdSlots = async (): Promise<void> => {
 	await Promise.all(
-		dynamicAdSlotModules.map(async ([name, init]) => {
+		dynamicAdSlotModules.map(async ([, init]) => {
 			try {
 				await init();
 			} catch (error) {
-				reportError(error, {
-					feature: 'commercial',
-					tag: name,
```
arelra, line 32: tag

src/init/consented/prepare-permutive.ts
```diff
@@ -157,7 +157,7 @@ const runPermutive = (
 			page: payload,
 		});
 	} catch (err) {
-		logger(err, { feature: 'commercial' }, false);
+		logger(err, 'commercial');
```
arelra, line 160: It looks like this is actually `reportError` but passed as a parameter called `logger`.

While we're here can we keep the name `reportError` so it is less confusing.

src/init/consentless.ts
```diff
@@ -10,11 +10,10 @@ import { initConsentless } from 'init/consentless/prepare-ootag';
...
 
 const bootConsentless = async (consentState: ConsentState): Promise<void> => {
 	const consentlessModuleList = [
-		initMessenger([background, resize, type], [], reportError),
```
arelra, line 17: Should we still be passing reportError through here?

src/insert/fill-dynamic-advert-slot.ts
```diff
@@ -29,14 +29,9 @@ const fillDynamicAdSlot = (
...
-				reportError(
-					Error(errorMessage),
-					{
-						feature: 'commercial',
-						slotId: adSlot.id,
-					},
-					false,
-				);
+				reportError(Error(errorMessage), 'commercial', {
+					adSlot: adSlot.id,
```
arelra, line 33: Looks like the name of the property here should be `slotId`

src/lib/dfp/non-refreshable-line-items.spec.ts
```diff
@@ -80,11 +76,8 @@ describe('nonRefreshableLineItems', () => {
 
 		expect(reportError).toHaveBeenCalledWith(
 			new Error('Failed to fetch non-refreshable line items'),
-			{
-				feature: 'commercial',
-				status: '404',
-			},
-			false,
+			'commercial',
+			{ feature: 'commercial', status: response.status },
```
arelra, line 80: We can drop `feature` from the tags as we pass it as a string as the second parameter

src/lib/dfp/non-refreshable-line-items.ts
```diff
@@ -42,14 +42,10 @@ export const fetchNonRefreshableLineItemIds = async (): Promise<number[]> => {
...
-	reportError(
-		error,
-		{
-			feature: 'commercial',
-			status: String(response.status),
-		},
-		false,
-	);
+	reportError(error, 'commercial', {
+		feature: 'commercial',
```
arelra, line 46: Drop feature here

src/utils/report-error.ts
```diff
@@ -1,44 +1,20 @@
...
-	return new Error(String(err));
+const wrapWithErrorReporting = <T>(
+	fn: ErrorReportingFunction<T>,
+	tags: Record<string, string> = {},
+): ErrorReportingFunction<T> => {
+	return function (event: T) {
+		try {
+			fn(event);
+		} catch (e) {
+			reportError(e, 'on-slot-render ', { tags: JSON.stringify(tags) });
```
arelra, line 15: I don't think we want to be passing the feature of `on-slot-render` here as this is a generic function and not related to `on-slot-render`

arelra: Have fixed this in my commits

src/utils/robust.ts
```diff
@@ -1,54 +1,31 @@
...
-): void => {
-	const error = catchErrors(fn);
-
-	if (error) {
-		logError(name, error, tags);
-	}
-};
-
-const catchErrorsWithContext = (
+const catchErrorsAndReport = (
```
arelra, line 10: Refactored this module to combine all the functions and updated the test `robust.spec` accordingly

src/utils/robust.spec.ts
```diff
@@ -1,71 +1,58 @@
...
-import { _, catchErrorsWithContext } from './robust';
+import { reportError } from './report-error';
+import { catchErrorsAndReport } from './robust';
 import type { Modules } from './robust';
 
-const { catchAndLogError } = _;
-
-jest.mock('utils/report-error', () => ({
-	convertError: jest.fn((e: Error) => e),
+jest.mock('./report-error', () => ({
```
arelra, line 5: Refactored this module to improve mocking and test single function now exported from `robust`

src/core/messenger.ts
```diff
@@ -1,3 +1,4 @@
+import { reportError } from '../utils/report-error'; // Keep this import relative otherwise Frontend will not be able to import this module. TODO: move all imports to relative imports
```
arelra, line 1: This was a surprising issue. If Frontend imports _any_ module from commercial that uses a Webpack alias the build will fail as it can't resolve the alias.

As a follow up we should:
- remove all webpack aliases and use relative imports everywhere
- tidy up the export from `core/index.ts` as there are lots exported modules there that are not actually used by any consumers and are a hangover from when core was its own separate lib

src/init/consentless.ts
arelra: Messenger wasn't using reportError so this is fine

src/core/messenger.ts
```diff
@@ -500,24 +491,12 @@ export const unregister: UnregisterListener = (type, callback, options) => {
...
-	>,
-	errorHandler: MessengerErrorHandler,
+	listeners: Array<(register: RegisterListener) => void>,
+	persistentListeners: Array<(register: RegisterPersistentListener) => void>,
 ): void => {
-	reportError = errorHandler;
-	listeners.forEach((moduleInit) => moduleInit(register, errorHandler));
+	listeners.forEach((moduleInit) => moduleInit(register));
 	persistentListeners.forEach((moduleInit) =>
-		moduleInit(registerPersistentListener, errorHandler),
```
arelra, line 520: None of the listeners were actually using the `errorHandler` aka `reportError` so this has been removed

emma-imber (approved):
Thanks for the run through - looks great!

arelra:
Thanks for your work on this @dskamiotis ✨

PR #1546, opened 2024-08-21
Return highest priority line item if it has a priority of 4 or less

emma-imber:
## What does this change?
Adds an if statement to the `pickLineItem` function to return the highest priority line item if there is a sponsorship or takeover targeting the page.

## Why?
This will make sure that GEM respects takeover campaigns when they're eligible to serve, instead of serving another advert.

### DIFF
--- a/src/elements-manager/find-creative.ts
+++ b/src/elements-manager/find-creative.ts
@@ -19,6 +19,17 @@ const pickLineItem = (lineItems: LineItem[]) => {
 		(a, b) => 1 / a.priority - 1 / b.priority,
 	);
 
+	const highestPriorityLineItem = sortedLineItems[-1];
+
+	// Takeovers have a priority of 4 or less, so if the highest priority
+	// line item has a priority of 4 or less, we should return that
+	if (
+		highestPriorityLineItem !== undefined &&
+		highestPriorityLineItem.priority < 5
+	) {
+		return highestPriorityLineItem;
+	}
+
 	const randomNumber = Math.random() * randomMultiplier;
 
 	let accumulator = 0;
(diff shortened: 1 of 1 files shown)


Jakeii (approved):
Definitely better than before! But what happens if there's competing sponsorships of the same priority?

emma-imber:
> Definitely better than before! But what happens if there's competing sponsorships of the same priority?

Yeah interesting question! Looks like GAM handles this case by looking at the % delivery goal. If it's 100%, it's a takeover and should be delivered, so we could sort by the percentage goal as a way of choosing between equal priorities? Looks like in the case that two line items of equal priority and percentage goal compete, each would get delivered 50% of the time. I found this interesting reading for how GAM handles competing sponsorships: https://support.google.com/admanager/answer/177426?sjid=17817089396196796428-EU#:~:text=patterns%20and%20seasonality.-,Overbooked%20competing%20sponsorships,-When%20multiple%20sponsorships

emma-imber:
> > > Definitely better than before! But what happens if there's competing sponsorships of the same priority?
> > 
> > 
> > Yeah interesting question! Looks like GAM handles this case by looking at the % delivery goal. If it's 100%, it's a takeover and should be delivered, so we could sort by the percentage goal as a way of choosing between equal priorities? Looks like in the case that two line items of equal priority and percentage goal compete, each would get delivered 50% of the time. I found this interesting reading for how GAM handles competing sponsorships: [support.google.com/admanager/answer/177426?sjid=17817089396196796428-EU#:~:text=patterns and seasonality.-,Overbooked competing sponsorships,-When multiple sponsorships](https://support.google.com/admanager/answer/177426?sjid=17817089396196796428-EU#:~:text=patterns%20and%20seasonality.-,Overbooked%20competing%20sponsorships,-When%20multiple%20sponsorships)
> 
> Nice find! Ah so we should use the delivery goal too to choose a line item, will need to add it to the GEM schema
> 
> That could be a followup?

Yep sounds good! I'll add a card for it as a follow up, and merge this as it's probably still an improvement

PR #1550, opened 2024-08-27
Fix spacefinder placing ads too close together on tablet & desktop

Jakeii:
## What does this change?
Tablets should use the mobile spacefinder pass/rules as all ads are inserted inline, like they are on mobile.

I had to add specific ad count/positions to the e2e tests as although they're using the mobile rules, things will be in different positions due to text flowing differently in paragraphs etc.

The right rail pass should also be avoiding inline1, at the moment inline2 is way too close to inline1.

## Why?
There are a few instances of ads being too close together that this will fix

<table>
<tr>
<td>Tablet
<td>Desktop
<tr>
<td>Screenshot 2024-08-27 at 16 05 12
<td>Screenshot 2024-08-27 at 16 04 45
</table>

And some other tablet issues occuring due to the use of desktop rules
Screenshot 2024-08-27 at 16 04 45

This is not great for UX!

### DIFF
--- a/.changeset/bright-beans-train.md
+++ b/.changeset/bright-beans-train.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
+Fix issue with double ads on tablets
--- a/src/insert/spacefinder/article.ts
+++ b/src/insert/spacefinder/article.ts
@@ -191,11 +191,20 @@ const addDesktopRightRailAds = (
 	);
 };
 
-const addMobileInlineAds = (fillSlot: FillAdSlot): Promise<boolean> => {
+const addMobileInlineAds = (
+	fillSlot: FillAdSlot,
+	currentBreakpoint: ReturnType<typeof getCurrentBreakpoint>,
+): Promise<boolean> => {
 	const insertAds: SpacefinderWriter = async (paras) => {
 		const slots = paras.map(async (para, i) => {
-			const name = i === 0 ? 'top-above-nav' : `inline${i}`;
-			const type = i === 0 ? 'top-above-nav' : 'inline';
+			const name =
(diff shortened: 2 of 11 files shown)


src/insert/spacefinder/article.ts
```diff
@@ -191,11 +191,20 @@ const addDesktopRightRailAds = (
 	);
 };
 
-const addMobileInlineAds = (fillSlot: FillAdSlot): Promise<boolean> => {
+const addMobileInlineAds = (
```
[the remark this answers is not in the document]
Jakeii: Isn't a tablet a mobile device? 😂 So it _could_ still be an appropriate name, but as source has 'mobile' === 'phone'  I guess `addMobileAndTabletInlineAds` works

src/insert/spacefinder/rules.ts
```diff
@@ -30,11 +30,12 @@ const minDistanceBetweenInlineAds = isInHighValueSection ? 500 : 750;
 const candidateSelector = ':scope > p, [data-spacefinder-role="nested"] > p';
 
 const leftColumnOpponentSelector = ['richLink', 'thumbnail']
-	.map((role) => `[data-spacefinder-role="${role}"]`)
+	.map((role) => `:scope > [data-spacefinder-role="${role}"]`)
```
Jakeii, line 33: These need `:scope >` as they're finding nested elements to avoid at the moment

.changeset/bright-beans-train.md
emma-imber, line 1: Duplicate changeset

Jakeii: This was intentional, they'll both end up in the changelog as separate changes

PR #1551, opened 2024-08-27
Check for videos as well as images when setting inline2 min above

emma-imber:
## What does this change?
Check for videos as well as images before increasing the min above for inline2.

## Why?
Articles with no images can still have a video, which means we don't need the extra buffer on the min above for inline2. By checking if we have a video atom, we should be able to improve ad ratio a bit on articles with videos but no images.

### Before
Screenshot 2024-08-27 at 17 46 43

### After
Screenshot 2024-08-27 at 17 46 27

### DIFF
--- a/src/insert/spacefinder/rules.ts
+++ b/src/insert/spacefinder/rules.ts
@@ -20,7 +20,8 @@ const isImmersive = window.guardian.config.page.isImmersive;
 
 const hasImagesOrVideo =
 	!!window.guardian.config.page.lightboxImages?.images.length ||
-	window.guardian.config.page.hasYoutubeAtom;
+	window.guardian.config.page.hasYouTubeAtom;
+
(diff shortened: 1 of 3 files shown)


src/insert/spacefinder/rules.ts
```diff
@@ -18,7 +18,10 @@ const MOST_VIEWED_HEIGHT = 600;
 
 const isImmersive = window.guardian.config.page.isImmersive;
 
-const hasImages = !!window.guardian.config.page.lightboxImages?.images.length;
+const hasImagesOrVideo =
+	!!window.guardian.config.page.lightboxImages?.images.length ||
+	window.guardian.config.page.hasYouTubeAtom;
```
arelra, line 23: I think its possible that we have a single video embed rather than a YouTube atom but this will capture the vast majority of cases.

arelra (approved):
Nice find! ✨

PR #1556, opened 2024-08-28
Add unit tests for GEM line item targeting logic

i-hardy:
- Adds a fixture file for a set of line items
- Adds unit tests for various aspects of targeting logic including device targeting
- Fixes a bug where if a line item's custom targeting parameters included an array of items to exclude, only the first item in the array was checked
- Moves the toggle for logging item selection logic decisions higher up the call stack (to make the tests less noisy)


Jakeii (approved):
Awesome, gives much more confidence in the code! And nice work spotting the bug🕵️

PR #1559, opened 2024-08-29
Check user is a member of Guardian org for beta releases

i-hardy:
## What does this change?

This enables all members of the Guardian org to release beta versions of the commercial bundle.

## Why?

Restricting beta release permissions only to repo admins is a little too narrow- sometimes it's useful for people outside the team to be able to make quick releases.

### DIFF
--- a/.changeset/angry-books-look.md
+++ b/.changeset/angry-books-look.md
@@ -1,5 +1,0 @@
----
-'@guardian/commercial': patch
----
-
-amend permissions for beta releases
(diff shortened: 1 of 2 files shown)


.changeset/angry-books-look.md
emma-imber, line 1: I think we don't need a changeset as we don't want to bump the commercial version for this change - it's just an improvement to our repository workflow

emma-imber (approved):
Nice improvement!

PR #1560, opened 2024-08-29
Delete mediator

Jakeii:
## What does this change?
Remove mediator

## Why?
It hasn't worked for some time, searching for the coresponding listeners or emitters in frontend often doesn't even find anything, we can safely get rid of this module and it's dependency now.


arelra (approved):
✨

PR #1572, opened 2024-09-03
Remove `optimiseSpacefinderInline` test and set changes live

Jakeii:
## What does this change?
Remove `optimiseSpacefinderInline` test and set changes live

## Why?
AB test was successful, more revenue and ad ratio, no noticeable effect on UX.


arelra (approved):
🎉 great result @emma-imber 

Original PR just for reference:
https://github.com/guardian/commercial/pull/1517

PR #1585, opened 2024-09-13
Handle competing sponsorships

emma-imber:
## What does this change?
Uses delivery goal to select between competing line items with the same priority level if they have a sponsorship level priority. Line items with a higher delivery goal are prioritised, as a delivery goal of 100% is used when we're delivering a takeover.

I've tested this using a local json object by adding an extra fake line item with a lower delivery goal but same priority to make sure that the item with the highest delivery goal is selected.

### DIFF
--- a/src/elements-manager/find-creative.ts
+++ b/src/elements-manager/find-creative.ts
@@ -35,8 +35,6 @@ const pickLineItem = (lineItems: LineItem[]) => {
 				lineItem.priority === highestPriorityLineItem.priority,
 		);
 
-		console.log(competingSponsorshipLineItems);
-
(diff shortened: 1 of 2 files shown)


Jakeii (approved):
Nice, but if there's several with the same delivery goal 😛

emma-imber:
> Nice! But if there's several with the same delivery goal 😛

I feel like it's fine to randomly pick between any line items that have the same priority and delivery goal - in theory they shouldn't be set up so that that can happen anyway. Have added some more code to check for competing line items with the same delivery goal and choose between them if needed!

PR #1592, opened 2024-09-18
Update @guardian/libs for new US Framework, usnat

akinsola-guardian:
## What does this change?
This PR updates the @guardian/libs version, transitioning the US Privacy Framework from ccpa to usnat.

Including the gpp property in the Prebid config (https://docs.prebid.org/dev-docs/modules/consentManagementGpp.html)
Replacing the ccpa in the ConsentState throughout the codebase with usnat.

## Why?
Migration of US Privacy framework from ccpa to USNAT/GPP.


arelra (approved):
lgtm @akinsola-guardian 

My only remaining concern is the sequencing of rolling out this change. In particular if there might be an issue with rolling out on Frontend and DCR at different times. 

I can catch up with you separately about this.

PR #1597, opened 2024-09-24
Add new 371x660 prebid size

Jakeii:
## What does this change?
This is a new size that run through pubmatic via prebid.js in the inline2 slot on articles.

The new size is not allowed to refresh, which is handled in the size callbacks; or display after a slot is refreshed which is handled by negatively targeting the `refreshed` custom targeting param.

New line items have been set up in Ad Manager.

### DIFF
--- a/src/insert/spacefinder/article.ts
+++ b/src/insert/spacefinder/article.ts
@@ -204,7 +204,7 @@ const additionalMobileAndTabletInlineSizes = (index: number) => {
 			],
 		};
 	}
-	return {};
+	return undefined;
 };
 
 const addMobileAndTabletInlineAds = (
--- a/src/insert/spacefinder/article.ts
+++ b/src/insert/spacefinder/article.ts
@@ -191,6 +191,22 @@ const addDesktopRightRailAds = (
 	);
 };
 
+const additionalMobileAndTabletInlineSizes = (index: number) => {
+	if (index === 1) {
+		return {
+			mobile: [adSizes.portraitInterstitial],
+		};
+	} else if (index === 2) {
+		return {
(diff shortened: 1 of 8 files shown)


src/insert/spacefinder/article.ts
```diff
@@ -209,12 +225,7 @@ const addMobileAndTabletInlineAds = (
 			return fillSlot(
 				name,
 				slot,
-				// Add the mobile portrait interstitial size to inline1 and inline2
-				i == 1 || i == 2
-					? {
-							mobile: [adSizes.portraitInterstitial],
-						}
-					: undefined,
+				additionalMobileAndTabletInlineSizes(i),
```
arelra, line 228: Previously we passed undefined to `fillSlot` if there wasn't a match. I assume `{}` will not change the behaviour but it would be good to narrow the expected types if we can.

Jakeii: I would argue it's fine accepting both? `undefined` works because it's an optional parameter ([which is defaulted to `{}` eventually](https://github.com/guardian/commercial/blob/jlk/prebid-pubmatic-interscroller/src/define/Advert.ts#L137)) and `{}` satisfies the `SizeMapping` type.

We could require a `SizeMapping` to have at least 1 property which makes sense, but setting that up is a bit of a pain AFAIK? I think we may rely on empty size mappings elsewhere though.

arelra: It seems like we're using `{}` as a proxy for no sizes in which case 'undefined' is more applicable? Very minor just a suggestion!

arelra (approved):

PR #1598, opened 2024-09-25
Bump source to 8.0.0

Jakeii:
## What does this change?
Bump source to 8.0.0

Also looks like typescript had accidently been updated to 5.6 by dependabot, putting it back for now

### DIFF
--- a/src/lib/detect/detect-breakpoint.ts
+++ b/src/lib/detect/detect-breakpoint.ts
@@ -23,7 +23,7 @@ const tweakpoints = {
 	leftCol: sourceBreakpoints['leftCol'],
 };
 
-const breakpointNames: SourceBreakpoint[] = [
+const sourceBreakpointNames: SourceBreakpoint[] = [
 	'wide',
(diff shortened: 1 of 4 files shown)


Jakeii:
This appears to have out breakpoint detection code. Which is odd, because the source breakpoints sure haven't changed🤔

src/lib/detect/detect-breakpoint.ts
```diff
@@ -105,15 +105,15 @@ const updateBreakpoint = (breakpoint: SourceBreakpoint): void => {
  * using getViewPort, which utilizes window.innerWidth which causes a reflow.
  */
 const initMediaQueryListeners = () => {
-	Object.entries(sourceBreakpoints).forEach(([bp], index, bps) => {
```
Jakeii, line 108: The order of the keys in this object have changed (which is not guaranteed anyway), `sourceBreakpointNames` makes more sense as it's an array.

arelra (approved):

PR #1599, opened 2024-09-25
Add prebid sharedId module and test

Jakeii:
## What does this change?
configure the "sharedIdSystem" using the `_pubcid` cookie we used to use.

What is it? See the [prebid.js docs](https://docs.prebid.org/dev-docs/modules/userid-submodules/sharedid.html) TLDR it's a shared identifier sent to prebid partners only when the user has consented.

This is behind a 50% test so we can find out if it actually does improve ad revenue.

## Why?
partners will pay us more for ad space if we share this id.


src/experiments/tests/prebid-shared-id.ts
```diff
@@ -0,0 +1,28 @@
+import type { ABTest } from '@guardian/ab-core';
+
+export const prebidSharedId: ABTest = {
+	id: 'PrebidSharedId',
+	author: '@commercial-dev',
+	start: '2024-09-25',
+	expiry: '2024-10-30',
+	audience: 0 / 100,
```
emma-imber, line 8: Are we setting it to 50% in a later PR?

emma-imber (approved):

src/experiments/tests/prebid-shared-id.ts
Jakeii: whoops I changed `audienceOffset` by mistake! Nice spot thanks

PR #1612, opened 2024-10-07
Remove unused exports

arelra:
## What does this change?

Removes unused exports from the commercial repo.

This is a hangover from when `commercial-core` and `commercial` were two separate libraries and we would export from core into commercial. Since merging we can now remove any exported functions which were just for internal sharing and not used by other repositories.

I based the exports that we should keep on this search:
https://github.com/search?q=org%3Aguardian+%40guardian%2Fcommercial&type=code&p=1

Exports required for atoms-rendering were ignored as that is an archived project.

### DIFF
--- a/.changeset/swift-jars-drop.md
+++ b/.changeset/swift-jars-drop.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': patch
+---
+
+Removed unused exports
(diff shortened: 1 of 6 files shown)


emma-imber (approved):
🧹

PR #1616, opened 2024-10-11
Enforce relative imports when importing from core

emma-imber:
## What does this change?
Adds an eslint rule to ban non-relative imports from core and replaces all instances with relative imports.

## Why?
We had [an issue where non relative imports were causing build failures](https://github.com/guardian/commercial/pull/1541#discussion_r1786255802), so moving to relative imports is sensible as our best practice.

There are a few folders we need to do this for - I'm going to split the changes into several separate PRs for my own sanity and to make reviewing (hopefully) easier. I felt like 74 files changed just from replacing imports of core was bulky enough for 1 PR without the other updates too!

### DIFF
--- a/.changeset/stupid-gifts-grab.md
+++ b/.changeset/stupid-gifts-grab.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
+Use relative imports when importing from core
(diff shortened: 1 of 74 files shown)


arelra (approved):
📈

PR #1626, opened 2024-10-15
Ban non-relative imports from utils

emma-imber:
## What does this change?
Adds an eslint rule to ban non-relative imports from utils, and updates all imports to comply.

## Why?
Non-relative imports can cause issues with builds when being used in frontend.


arelra (approved):
✨

PR #1634, opened 2024-10-24
Add description_url to IMA ad requests

emma-imber:
## What does this change?
Adds the article URL to the youtube IMA ad request, by adding the description_url parameter. It seems pretty common practice from what I've read online to use the URL of the page that the video is loading on as the description url. Also updates tests to make sure the request URL is constructed as expected.

## Why?
This is used to determine if the domain is eligible for ads to serve. See [here](https://support.google.com/admanager/answer/1734048?hl=en#:~:text=video%20ad%20inventory.-,The%20description%20URL,-For%20Ad%20Exchange) for more info.

## Testing

I've tested on CODE, and we get ads loading before the video as expected:

Screenshot 2024-10-24 at 16 25 05

And we can see description_url getting added to the payload:

Screenshot 2024-10-24 at 16 18 26

### DIFF
--- a/src/core/targeting/youtube-ima.ts
+++ b/src/core/targeting/youtube-ima.ts
@@ -85,6 +85,9 @@ const buildImaAdTagUrl = ({
 		impl: 's',
 		vad_type: 'linear',
 		vpos: 'preroll',
+		description_url: encodeURIComponent(
+			`https://www.theguardian.com/${window.guardian.config.page.pageId}`,
+		),
 		/**
 		 * cust_params string is encoded
 		 * cust_params values are also encoded so they will get double encoded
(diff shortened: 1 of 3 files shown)


src/core/targeting/youtube-ima.ts
```diff
@@ -85,6 +85,9 @@ const buildImaAdTagUrl = ({
 		impl: 's',
 		vad_type: 'linear',
 		vpos: 'preroll',
+		description_url: encodeURIComponent(
+			`https://www.theguardian.com/${window.guardian.config.page.pageId}`,
```
arelra, line 89: I wonder if we can combine `window.guardian.config.page.host` to get the URL domain from config

```suggestion
			`${window.guardian.config.page.host}/${window.guardian.config.page.pageId}`,
```

arelra (approved):

PR #1635, opened 2024-10-25
Add comment to explain what prebidPermutiveAudience switch is for

emma-imber:
## What does this change?
Adds a comment to describe what the prebidPermutiveAudience switch is used for.

## Why?
It wasn't immediately clear looking at the code what the difference between the permutive switches was - hopefully this makes things clearer in the future.


deedeeh (approved):
Thank you for making it clearer to everyone 👍🏼

PR #1641, opened 2024-10-31
Don't define static top-above-nav on consentless mobile DCR

emma-imber:
## What does this change?
Filter out the static top-above-nav slot on mobile DCR when defining consentless fixed slots.

## Why?
At the moment, two top-above-nav slots get defined - the static one, and the dynamically inserted mobile top-above-nav. The styling and ad label logic is currently being applied to the static slot, not the dynamic one, causing display issues. Filtering out the static slot so that we don't have duplicate definitions mirrors the logic for consented ads, and makes sure the label and extra classes are applied as expected.

### MPU
Screenshot 2024-10-31 at 15 45 08

### Fabric
Screenshot 2024-10-31 at 15 40 25

### MPU
Screenshot 2024-10-31 at 15 45 44

### Fabric
Screenshot 2024-10-31 at 15 39 29

### DIFF
--- a/.changeset/eight-tools-press.md
+++ b/.changeset/eight-tools-press.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
+Filter out static top-above-nav on consentless mobile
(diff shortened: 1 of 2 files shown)


deedeeh (approved):
Looks good Emma!

PR #1654, opened 2024-11-11
Setup test for new prebid analytics endpoint

Jakeii:
## What does this change?
Adds a second endpoint for prebid analytics that is behind a 0% test.

## Why?
To get a small amount of data in the new table to compare with the current data.

### DIFF
--- a/src/lib/header-bidding/prebid/prebid.ts
+++ b/src/lib/header-bidding/prebid/prebid.ts
@@ -8,6 +8,8 @@ import { PREBID_TIMEOUT } from '../../../core/constants/prebid-timeout';
 import { EventTimer } from '../../../core/event-timer';
 import type { PageTargeting } from '../../../core/targeting/build-page-targeting';
 import type { Advert } from '../../../define/Advert';
+import { isUserInVariant } from '../../../experiments/ab';
+import { newHeaderBiddingEndpoint } from '../../../experiments/tests/new-header-bidding-endpoint';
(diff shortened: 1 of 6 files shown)


src/lib/header-bidding/prebid/prebid.ts
```diff
@@ -423,6 +427,11 @@ const initialise = (
 				options: {
 					ajaxUrl: window.guardian.config.page.ajaxUrl ?? '',
 					pv: window.guardian.ophan.pageViewId,
+					enableV2Endpoint: isUserInVariant(
+						newHeaderBiddingEndpoint,
+						'variant',
+					),
+					ajaxUrlV2: `//performance-events.code.dev-guardianapis.com/header-bidding`,
```
deedeeh, line 434: Could you expand on this endpoint please? How did you form it in this way?

Jakeii: `performance-events.code.dev-guardianapis.com` (and the prod version `performance-events.guardianapis.com`) is setup in fastly already for commercial-metrics and core-web-vitals, I've added the `header-bidding` path within fastly to send data to the datalake. Perhaps I can go over it in a dev time?

deedeeh: Oh ok! Yes that would be great please 🙂

deedeeh (approved):

PR #1657, opened 2024-11-12
Moving changesets to dev dependencies

rowannekabalan:
This dependency has turned up in the dependency tree on Frontend as it's in the dependencies list, however we think this is a dev dependency so we have moved it accordingly.

### DIFF
--- a/.changeset/lemon-nails-drop.md
+++ b/.changeset/lemon-nails-drop.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': patch
+---
+
+Move changesets to dev dependency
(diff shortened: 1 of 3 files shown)


domlander:
Changesets is a dev dependency in other Guardian repo's: https://github.com/search?q=org%3Aguardian+%22%40changesets%2Fcli%22%3A&type=code

emma-imber (approved):
Looks good. Just an FYI that you'll need to add a changeset if you want to release this change, or it'll just be included next time we bump the version - whatever works for you!

PR #1661, opened 2024-11-18
Use region specific prebid bundles

emma-imber:
## What does this change?
Uses region specific bundles of Prebid to reduce the amount of javascript that we ship to the user.

Draft for testing

### Before
Prebid chunk loaded in UK before:
Screenshot 2024-11-18 at 18 10 13

### After
Prebid chunk loaded in UK after:
Screenshot 2024-11-18 at 18 11 15


dskamiotis (dismissed):

emma-imber:
> apologies for confusion, have tested locally just out of interest, with the `pnpm run build:prod`, and in my `bundle-analyzer-report` still getting the original size of 427kb, might I be missing a step?

Hey Demetrios - this PR is still a draft, and not ready for review yet.

In reply to your question about the prebid size - that's because the import is behind an AB test, so the build just contains the current version of prebid. If you were to run this locally, opt in to the test, use a VPN, and then check the version of prebid getting loaded in the network tab, you would see a smaller version being imported.

arelra (approved):
🎉

Jakeii (approved):

mchv:
@emma-imber 💎 Great job on this, looking forward for the outcome

PR #1679, opened 2024-11-22
Show ads in right column of football live page

SiAdcock:
## What does this change?

Allow ads to show in the right column on the [football live page](https://www.theguardian.com/football/live). 

## Why?

8ca948df7b68e24dbd666765e9457f6b

Follow on from #1379


Jakeii (approved):
Cheers Simon! 🤑

PR #1685, opened 2024-11-26
Unskip liveblog-ad-limit test

arelra:
## What does this change?

Update `liveblog-ad-limit` e2e test to reduce flakiness.

On inspection it looks like its possible for `window.mockLiveUpdate` to be undefined until the `Liveness` island is hydrated and the function is set on the window.

Adding a wait on `window.mockLiveUpdate` to be defined should remove this corner case.


deedeeh (approved):
I'm glad we won't have that issue anymore. Thank you Ravi for doing it ✨

PR #1692, opened 2024-12-04
remove a9 block bidder add for fronts

dskamiotis:
## What does this change?
This PR reintroduces the ability for GumGum (interscroller ads) to serve on Fronts after being previously blocked. 

Changes to A9 module:
1. **Context-Based Filtering:**
 - Filters ad units based on the page context:
   - Network front: Includes only the ad unit with slot ID `dfp-ad--inline1--mobile`.
   - Section front: Includes only the ad unit with slot ID `dfp-ad--top-above-nav`.
   - Non-front pages: Includes all ad units.
   - For pages that are both articles and fronts, treats them as non-front pages and includes all ad units.

#### 2. Bidder Blocking:
- Adds a params object to each ad unit with a blockedBidders property:
Network front: Blocks GumGum for dfp-ad--inline1--mobile.
Section front: Blocks GumGum for dfp-ad--top-above-nav.

#### 3. Ad Unit Updates:
- Maps over ad units to include the new params object with blockedBidders, ensuring proper configuration based on context and bidder blocking.

## Why?
GumGum interscrollers were blocked in a previous ticket (linked below) but generated significant revenue before the block. Commercial has approved reintroducing them under these controlled conditions to balance user experience and revenue.

## Notes
- "Fronts" refer to prominent pages, including the Network Front (e.g., homepage) and Section Fronts (e.g., category or topic-specific pages).
- The `inline1` and `top-above-nav` ad slots were chosen based on their positions to optimize both user engagement and ad performance.

Related PR: [Add blocked bidder within Amazon A9 for SSP GumGum when on Fronts](https://github.com/guardian/commercial/pull/1647)

## Screenshots

| Network Front   | 
|------------|
| Screenshot 2024-12-04 at 14 04 29) | 

| Section Front   | 
|------------|
| Screenshot 2024-12-04 at 14 02 30 | 

| Article: Non Front  page  | 
|------------|
| Screenshot 2024-12-06 at 18 02 08 |

### DIFF
--- a/src/lib/header-bidding/a9/a9.ts
+++ b/src/lib/header-bidding/a9/a9.ts
@@ -64,11 +64,23 @@ const requestBids = async (
 		return requestQueue;
 	}
 
+	const section = window.guardian.config.page.section;
+
+	const isNetworkFront =
+		section === 'uk' || section === 'us' || section === 'au';
+
+	const filteredAdUnits = adUnits.filter((adUnit) => {
+		if (isNetworkFront) {
+			return adUnit.slotID === 'dfp-ad--inline1--mobile';
+		}
+		return adUnit.slotID === 'dfp-ad--top-above-nav';
+	});
+
 	requestQueue = requestQueue
 		.then(
 			() =>
 				new Promise<void>((resolve) => {
-					window.apstag?.fetchBids({ slots: adUnits }, () => {
+					window.apstag?.fetchBids({ slots: filteredAdUnits }, () => {
 						window.googletag.cmd.push(() => {
 							window.apstag?.setDisplayBids();
 							resolve();
--- a/src/lib/header-bidding/a9/a9.ts
+++ b/src/lib/header-bidding/a9/a9.ts
@@ -92,23 +92,26 @@ const requestBids = async (
 	 * @param adUnits - The array of ad units to be filtered.
 	 * @returns The filtered array of ad units based on the page context.
 	 */
-	const filteredAdUnits = adUnits.filter((adUnit) => {
-		if (isNetworkFront) {
-			return adUnit.slotID === 'dfp-ad--inline1--mobile';
+	const updatedAdUnits = adUnits.map((adUnit) => {
+		let blockedBidders: string[] = [];
+
+		if (isNetworkFront && adUnit.slotID === 'dfp-ad--inline1--mobile') {
+			blockedBidders = ['1lsxjb4']; // Block GumGum for network front condition
+		} else if (
+			isSectionFront &&
+			adUnit.slotID === 'dfp-ad--top-above-nav'
+		) {
+			blockedBidders = ['1lsxjb4']; // Block GumGum for section front condition
 		}
-		if (isSectionFront) {
-			return adUnit.slotID === 'dfp-ad--top-above-nav';
-		}
-		if (!window.guardian.config.page.isFront) {
-			return true;
-		}
+
+		return { ...adUnit, params: { blockedBidders } };
 	});
 
 	requestQueue = requestQueue
 		.then(
 			() =>
 				new Promise<void>((resolve) => {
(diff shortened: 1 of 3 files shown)


src/lib/header-bidding/a9/a9.ts
```diff
@@ -70,11 +64,23 @@ const requestBids = async (
 		return requestQueue;
 	}
 
+	const section = window.guardian.config.page.section;
+
+	const isNetworkFront =
+		section === 'uk' || section === 'us' || section === 'au';
```
deedeeh, line 70: I am trying to check in the Trello ticket if those info are there but just double checking you got them from Alex? I mean are those the only three regions that we are showing a9 in Fronts?

src/lib/header-bidding/a9/a9.ts
```diff
@@ -70,11 +64,23 @@ const requestBids = async (
...
+
+	const filteredAdUnits = adUnits.filter((adUnit) => {
```
Jakeii, line 72: We still want a9 to work in the other slots, just nut gumgum afaik.

src/lib/header-bidding/a9/a9.ts
Jakeii: missing europe here

src/lib/header-bidding/a9/a9.ts
Amouzle: Yep, Jake's right. I see the issue here, we want:
- A9 to run everywhere
- But prevent GumGum running in certain slots, when it serves on section fronts

As I type this is realise it's quite the messy rule want to chat it through @dskamiotis?

src/lib/header-bidding/a9/a9.ts
dskamiotis: @deedeeh thats right `inline-1` slots only render on mobile, assuming it was an omission, and double checked with @Amouzle.

src/lib/header-bidding/a9/a9.ts
dskamiotis: thanks @Amouzle I have an idea/solution, but will chat it through to confirm if thats ok

src/lib/header-bidding/a9/a9.ts
dskamiotis: > missing europe here

@Jakeii assuming `international` as well as `europe`?

src/lib/header-bidding/a9/a9.ts
Jakeii: This still prevents a9 from running for all the other slots on fronts, it's also preventing all bidders not just GumGum?

src/lib/header-bidding/a9/a9.ts
```diff
@@ -70,11 +64,51 @@ const requestBids = async (
...
+	 *
+	 * @param adUnits - The array of ad units to be filtered.
+	 * @returns The filtered array of ad units based on the page context.
+	 */
+	const filteredAdUnits = adUnits.filter((adUnit) => {
+		if (isNetworkFront) {
+			return adUnit.slotID === 'dfp-ad--inline1--mobile';
+		}
+		if (isSectionFront) {
+			return adUnit.slotID === 'dfp-ad--top-above-nav';
```
emma-imber, line 100: Is this meant to be 'dfp-ad--top-above-nav--mobile'?

src/lib/header-bidding/a9/a9.ts
dskamiotis: I thought we only wanted specif slots for the network and section fronts?
On Network Front === inline1 (as we would like it further down the page)

On Section Fronts === top-above-nav
and all other slots for non fronts?

Jakeii: We want only GumGum to be blocked in the other slots, the code you've got prevents any A9 bidder not just GumGum.
As @Amouzle we still want A9 working everywhere.

PR #1697, opened 2024-12-10
Delete lib modules

arelra:
## What does this change?

Minor tidy up to remove some lib modules. 

Inlined or removed entirely where appropriate


src/core/messenger/scroll.spec.ts
```diff
@@ -91,8 +90,8 @@ describe.skip('Cross-frame messenger: scroll', () => {
 			constructor(callback: ObsCallback) {
 				onIntersect = callback;
 				return Object.freeze({
-					observe: noop,
-					unobserve: noop,
+					observe: () => {},
```
deedeeh, line 93: Does it mean here we can have a callback but before it was using function noop?

emma-imber (approved):
Nice 🧹

src/core/messenger/scroll.spec.ts
arelra: I think this is okay as this file is a test. I think its just creating a mock `IntersectionObserver` and we're stubbing some of the interface we're not concerned with in context of the test.

PR #1707, opened 2024-12-20
Delete more libs

arelra:
## What does this change?

When migrating out of Frontend we carried over lib modules even where we use a small number of functions from the overall module.

This PR inlines functions where possible which allows us to delete more of the old libs and their dependency chain.

Also some completely unused modules were deleted.

## Why?

Reduce LOC


Jakeii (approved):
Nice finds! 🧹

PR #1719, opened 2025-01-03
Switch to SWC Compiler

Jakeii:
## What does this change?
Migrate to use https://swc.rs for transpiling and minification instead of typescript and babel.

swc is widely used including by [guardian/dotcom-rendering](https://github.com/guardian/dotcom-rendering)

## Why?
SWC is much faster, and usually produces smaller bundle sizes, although in our case it's a very minor reduction in bundle size.

#### Build times
Dev
```sh
# before
webpack 5.97.1 compiled successfully in 11699 ms
# after 
webpack 5.97.1 compiled successfully in 1270 ms
```

Prod
```sh
# before
webpack 5.97.1 compiled with 1 warning in 14376 ms
# after
webpack 5.97.1 compiled with 1 warning in 2038 ms
```

#### Bundle sizes
Including all prebid bundles
| Before | After |
|--------|-------|
| 2.07 MB (699.92 KB gzipped) | 20.06 MB (698.48 KB gzipped) |

Including just uk prebid bundle (More demonstrative as only one bundle is ever served to a user)
| Before | After |
|--------|-------|
| 612.76 KB (197.72 KB gzipped) | 603.09 KB (196.91 KB gzipped) |


src/init/consented/prepare-prebid.ts
```diff
@@ -35,7 +35,7 @@ const prebidVersion = () => {
 const loadPrebid = async (framework: ConsentFramework): Promise<void> => {
 	if (shouldLoadPrebid()) {
 		await import(
-			/* webpackChunkName: "Prebid.js" */
+			/* webpackChunkName: "[request]" */
```
Jakeii, line 38: this gives us nicer names for the localised prebid chunks

PR #1721, opened 2025-01-06
Libs tidy up pt.3

arelra:
## What does this change?

More libs tidy up:

- Delete unused geolocation functions
- Inline and simplify get-ad-iframe
- Inline spacefinder utils

## Why?

Reduce singularly referenced modules and delete unused code

Previous:
https://github.com/guardian/commercial/pull/1707
https://github.com/guardian/commercial/pull/1697


src/events/render-advert.ts
```diff
@@ -179,6 +178,21 @@ const addContainerClass = (adSlotNode: HTMLElement, isRendered: boolean) => {
 		});
 };
 
+/**
+ * Check if the ad slot has a corresponding iframe to indicte the ad has rendered.
+ * @param adSlot
+ * @returns
+ */
+const hasIframe = (adSlot: HTMLElement): Promise<boolean> =>
```
arelra, line 186: This is a simplified version of the current function. All the IE specific code has been removed and the promise return value is now a boolean:

https://github.com/guardian/commercial/blob/cdefafe37158b6ca290a181ea3a3cc222d94ae5f/src/lib/dfp/get-ad-iframe.ts#L5-L44

PR #1722, opened 2025-01-06
Remove some old code related to supporting legacy browsers

Jakeii:
## What does this change?
Removes the `$$` module, it doesn't provide any utility these days.
Delete an old polyfill for passive event listeners, browser support is now at >97% so safe to get rid?

Quite sure `article-aside-adverts.ts` could be deleted also, but have carded up to check on that.

## Why?
Less/tidier code.


src/core/messenger/scroll.ts
```diff
@@ -91,7 +90,7 @@ const addScrollListener = (
 	respond: Respond,
 ): void => {
 	if (iframeCounter === 0) {
-		addEventListener(w, 'scroll', onScroll, {
+		w.addEventListener('scroll', onScroll, {
```
arelra, line 93: While we're here could we replace `w` with a direct reference to `window` in this module? It's a little surprising to see it aliased.

arelra (approved):
Nice! 📉

PR #1729, opened 2025-01-09
Libs tidy up part 4 - merge utils

arelra:
## What does this change?

Due to the prior migration of the commercial codebase out of Frontend and the merging of the previous commercial-core and commercial-bundle repos we have multiple lib and util type directories in the codebase e.g.

```
- core
  - lib
- lib
- utils
```

This PR moves all modules currently in `utils` into `lib, renames and creates appropriate directories wherever required.

The commits list the discrete changes.

Merging core is a bigger pieces of work as it exports functions for clients and will be tackled in a subsequent PR.

## Why?

Less surprising to find libs in different places across the codebase.


deedeeh (approved):
Looks good and spot on the renaming of the `editionToCountryCodeMap` variable 👍🏼

PR #1730, opened 2025-01-09
Revert "Add amIUsed calls to see if modules are still needed"

emma-imber:
Reverts guardian/commercial#1726

We've got a few hours of data now, and having queried BigQuery, `high-merch.ts` is still being used by galleries, but there are no entries for `article-aside-adverts.ts`, so we can delete that in a follow up PR.

### DIFF
--- a/src/insert/article-aside-adverts.ts
+++ b/src/insert/article-aside-adverts.ts
@@ -1,4 +1,3 @@
-import { amIUsed } from '../utils/am-i-used';
 import fastdom from '../utils/fastdom-promise';
 
 const minArticleHeight = 1300;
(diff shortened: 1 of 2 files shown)


arelra (approved):

Jakeii:
Could we also add a comment to `high-merch.ts` that it'll be redundant when gallaries are migrated to DCR?

emma-imber:
> Could we also add a comment to `high-merch.ts` that it'll be redundant when gallaries are migrated to DCR?

I've added a comment as part of this follow up PR: https://github.com/guardian/commercial/pull/1733

PR #1733, opened 2025-01-09
Delete article-aside-adverts as it's no longer used

emma-imber:
## What does this change?
Deletes `article-aside-adverts.ts` as it's no longer used. Also adds a comment to the `high-merch.ts` file to remind us to revisit whether it's still needed after the galleries migration.

## Why?
After adding `amIUsed` to check if these files are still needed, `high-merch.ts` is still used, but `article-aside-adverts.ts` is not, so we can delete it.


arelra (approved):
Nice! ✨

Jakeii (approved):

PR #1736, opened 2025-01-10
Add an error message if we can't pull DCR container image

deedeeh:
## What does this change?

This PR adds an error message to the Github Actions logs if we can't pull the DCR container image and the step is failed which means we need to rebuild DCR main to create a new image.

## Why?

We rarely fail this step because we can't pull the image from DCR due to no images https://github.com/guardian/commercial/actions/runs/11627393704/job/32380639973. It's hard to debug when the issue isn't ongoing so an error message will be sufficient until the issue happens again then we can debug.

### DIFF
--- a/.github/workflows/e2e-playwright.yml
+++ b/.github/workflows/e2e-playwright.yml
@@ -53,6 +53,7 @@ jobs:
       # This makes the commercial dev server available from inside the container
       # Note that GHA provides a service container feature, but it does not support this argument
       - name: Start DCR in a container
+        id: start-dcr-container
         run: |
           /usr/bin/docker run -d \
             --network host \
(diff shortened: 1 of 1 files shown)


dskamiotis (approved):
This makes sense, would you like to add a changeset to this? or will it be part of a larger piece of work?

deedeeh:
> This makes sense, would you like to add a changeset to this? or will it be part of a larger piece of work?

I believe no need for a changeset in this PR as it's a CI change so it's not affecting DCR or Frontend. @arelra please correct me if I am wrong.

arelra:
> > This makes sense, would you like to add a changeset to this? or will it be part of a larger piece of work?
> 
> I believe no need for a changeset in this PR as it's a CI change so it's not affecting DCR or Frontend. @arelra please correct me if I am wrong. 

That's right @deedeeh no changeset required as it's an internal change and doesn't change the public behaviour of the bundle or library.

arelra (approved):
This will be very helpful to diagnose the problem if it happens again ✨

PR #1737, opened 2025-01-10
0% test bundling prebid here in commercial, instead of importing pre-bundled prebid

Jakeii:
## What does this change?
Import the prebid source module rather than the built bundle

Our 3 custom/modded modules can be moved here, the modifying of the adapters (adding aliases) can be done without copying them.

I've moved the `guAnalyticsModule` here as is, I timeboxed a conversion to typescript, but it's a little complex, I think it can be converted in a subsequent PR? I think it needs to be rewritten to be honest.

I've put this behind a 0% test so that it can be tested on theguardian.com, as we don't get much prebid activity locally or on CODE.

## Why?
It will allow us to use our own bundling tools, makes prebid upgrades much easier and removes the need to maintain our own prebid fork `@guardian/prebid`


webpack.config.mjs
```diff
@@ -60,6 +82,14 @@ const config = {
 					},
 				],
 			},
+			{
+				test: /.js$/,
+				include: /prebid\.js/,
+				use: {
+					loader: 'babel-loader',
+					options: prebidBabelOptions,
```
Jakeii, line 90: prebid has some odd babel configuration so we need to use their babel config, this is recommended in the prebid docs.

PR #1752, opened 2025-01-21
Remove octokit dependency

cemms1:
## What does this change?

Removes `@octokit/core` from the dependencies list

## Why?

It's not used anywhere, and even if it was it probs should be a dev dependency


deedeeh (approved):
🧹

PR #1764, opened 2025-01-24
Refactor commercial boot process

emma-imber:
## What does this change?
Splits out the `loadModules`, `recordCommercialMetrics` and `bootCommercial` functions into a separate `commercial-boot-utils` file for tidiness. Also splits `consented.ts` into two files - `ad-free.ts` and `consented-advertising.ts`. This (in my opinion) makes the ad free logic easier to find and clearer to read. By doing this, we also get rid of the idea of "base" and "extra" modules, which were confusingly named.

Also removes the naming of the modules and no longer loads using the `catchErrorsAndReport` function. This also means we can delete the `robust` module as this was the only use of it.

## Why?
Having explained this code several times over the last few weeks to different people, I have noticed a few parts that don't make immediate sense when you try to explain them (eg the naming of base vs extra modules). I think this little refactor is much more readable, and focuses the different bundle entry point files on the modules that they load.

Also, by having an ad-free specific file, we should be shipping less code to end users who don't see ads. I am also in favour of having the ad free check in the commercial file instead of the consented file, as I think it makes the purpose of each file much clearer to understand.

Would like to hear other people's thoughts on if this is an improvement or not!


Jakeii:
Looks much nicer to me!

Jakeii:
The whole array of module names with init functions and `catchErrorsAndReport` looks to be just to help with error reporting?

I can't ever recall looking for errors in sentry using these, could we get rid and simplify further?

src/commercial.ts
```diff
@@ -18,13 +18,18 @@ void (async () => {
 		!commercialFeatures.adFree
 	) {
 		void import(
-			/* webpackChunkName: "consentless" */
-			'./init/consentless'
+			/* webpackChunkName: "consentless-advertising" */
+			'./init/consentless-advertising'
```
arelra, line 22: This is very minor so feel free to ignore :) What do we gain anything by adding `-advertising` to all the chunks? Is it to help searching in dev tools?

src/init/consented-advertising.ts
```diff
@@ -0,0 +1,56 @@
...
+import { init as initThirdPartyTags } from './consented/third-party-tags';
+import { init as initTrackGpcSignal } from './consented/track-gpc-signal';
+import { init as initTrackScrollDepth } from './consented/track-scroll-depth';
+import { reloadPageOnConsentChange } from './shared/reload-page-on-consent-change';
+import { init as setAdTestCookie } from './shared/set-adtest-cookie';
+import { init as setAdTestInLabelsCookie } from './shared/set-adtest-in-labels-cookie';
+
+// all modules needed for commercial code and ads to run
+const commercialModules: Modules = [
+	['cm-adFreeSlotRemoveFronts', adFreeSlotRemove],
```
arelra, line 25: While we're here... do we ned the `cm-` prefix to all the modules any longer. I suspect its something we carried over from Frontend. Could be a follow up.

arelra (approved):
Nice refactor ✨ !

src/commercial.ts
emma-imber: It was only to make the distinction between code that loads an ad and the ad free code really explicit to anyone looking at it for the first time. I'm not set on it at all - happy to go with the majority opinion!

emma-imber:
> The whole array of module names with init functions and `catchErrorsAndReport` looks to be just to help with error reporting?
> 
> I can't ever recall looking for errors in sentry using these, could we get rid and simplify further?

Yeah I don't think I've ever used this either! It's also the only bit of code using catchErrorsAndReport, so we could delete the robust file if we got rid

src/init/consented-advertising.ts
emma-imber: Yeah good point - I think this ties in with Jake's point about reporting. If we're only using the module names for error reporting, maybe we can remove them altogether

src/init/consented-advertising.ts
```diff
@@ -0,0 +1,56 @@
...
+	initTrackGpcSignal,
+	initMessenger,
+	setAdTestCookie,
+	setAdTestInLabelsCookie,
+	reloadPageOnConsentChange,
+	preparePrebid,
+	initDfpListeners,
+	// Permutive init code must run before google tag enableServices()
+	// The permutive lib however is loaded async with the third party tags
+	() => initPermutive().then(prepareGoogletag),
```
i-hardy, line 40: Maybe this could be part of a separate bit of work to tidy up the Permutive/Googletag code, but it would be nice if this could be a module unto itself like the others rather than a random hard-coded function and comment breaking up the array

src/lib/init-utils.ts
```diff
@@ -0,0 +1,59 @@
...
+		return Promise.allSettled(modules.map((module) => module())).then(
+			recordCommercialMetrics,
+		);
+	} catch (error) {
+		// report async errors in bootCommercial to Sentry with the commercial feature tag
+		reportError(error, 'commercial', tags);
+	}
+};
+
+export { bootCommercial };
```
i-hardy, line 59: I'm a bit allergic to everything being called 'init' in any case, but given the export maybe this file could be named something like 'commercial-boot-utils'? I think it would be a little clearer. Breaking these out from the main file is a huge improvement though!

src/commercial.ts
```diff
@@ -18,13 +18,18 @@ void (async () => {
 		!commercialFeatures.adFree
```
i-hardy, line 18: Could this condition be moved out into a named function that describes this particular state?

Eg.

```suggestion
		function shouldRunConsentless(consentState): boolean {
				return window.guardian.config.switches.optOutAdvertising &&
				consentState.tcfv2 &&
				!getConsentFor('googletag', consentState) &&
				!commercialFeatures.adFree;
		}
		...
		if (shouldRunConsentless(consentState) {
```

It would improve readability and make it clearer what all these conditions actually add up towards.

src/init/consented-advertising.ts
emma-imber: I'm going to address this in a separate PR, but yes, agreed!

PR #1651, opened 2024-11-08
adds manual gpid to prebid ad unit and launch to 99% of users

dskamiotis:
## What does this change?
The GPID is being added to enable the successful sending of GPID to prebid vendors. This implementation uses the [GPT Pre-Auction module](https://docs.prebid.org/dev-docs/modules/gpt-pre-auction.html), as recommended by the [Prebid Docs](https://docs.prebid.org/features/pbAdSlot.html) if your project is using GAM/GPT.

## Why?
We are adding the GPID to prebid as part of the OpenRTB framework, there isa field in the impressionExtension object called imp.ext.gpid. This will allow uis to conform to the protocol by sending a standardised GPID across all SPP's.
The Guardian’s ad server (GAM) to Prebid needs to follow certain standards to ensure that Prebid can interpret the information correctly and that all exchanges involved in the auction can handle and propagate it.

This includes:
- updating the `PrebidAdUnit` class to include a method to dynamically generate a GPID for each ad unit
- using the `Unique ad slots` method outlined in the [docs](https://docs.prebid.org/features/pbAdSlot.html):

- Including the GPID to 99% of users - users in variant with `shouldIncludeGpid`
```
pbjs.addAdUnits({
    code: '/1111/homepage-leftnav',
    ortb2Imp: {
        ext: {
            gpid: "/1111/homepage-leftnav",
        data: {
                pbadslot: "/1111/homepage-leftnav"
            }
        }
    },
    mediaTypes: ...
    bids: ...
});
```
- Added Testing - 1% control and 1% variant
- Control - continues usuing existing protocol data structure
- Variant - adds the new structure:
```js
ortb2Imp: {
        ext: {
            gpid: "/1111/homepage-leftnav",
        data: {
                pbadslot: "/1111/homepage-leftnav"
            }
        }
 ```

| OpenRTB in professor prebid - Control - `#ab-gpidPrebid=control`|
|--------|
| Screenshot 2024-11-14 at 09 33 43 |

| OpenRTB in professor prebid - Variant  - `#ab-gpidPrebid=variant`|
|--------|
| Screenshot 2024-11-08 at 21 23 44 |

PR #1772, opened 2025-01-29
remove test for GPID in prebid

dskamiotis:
## What does this change?
This PR removes the GPID in prebid config test, and exposes it to all users ([related PR](https://github.com/guardian/commercial/pull/1651))

## Why?
The GPID is being added to enable the successful sending of GPID to prebid vendors. This implementation uses the [GPT Pre-Auction module](https://docs.prebid.org/dev-docs/modules/gpt-pre-auction.html), as recommended by the [Prebid Docs](https://docs.prebid.org/features/pbAdSlot.html) if your project is using GAM/GPT.

We are happy to expose the GPID in the prbid config, given a 2% uplift.

### Tested: 
Locally with `#ab-GpidPrebidAdUnits=control`

Since the `control` portion was outside of the test (1%), they did not recieve the GPID in the prebid paylopad.
Here we see with the opting into the the test as control the client does get the GPID value, proving the test has been successfully removed. 

http://localhost:3030/Article/http:/localhost:9000/sport/2023/nov/22/ronnie-osullivan-warns-he-will-quit-snooker-if-he-cannot-play-in-china#ab-GpidPrebidAdUnits=control
Screenshot 2025-01-30 at 09 32 50


src/experiments/ab-tests.ts
```diff
@@ -1,5 +1,4 @@
 import type { ABTest } from '@guardian/ab-core';
-import { gpidPrebidAdUnits } from './tests/gpid-prebid';
```
emma-imber, line 2: You'll also need to delete the test file you were importing from here

Jakeii (approved):

PR #1773, opened 2025-01-30
Add The Trade Desk to prebid

Jakeii:
## What does this change?
Adds new config for new prebid bidder The Trade Desk.

The line-items have already been setup. And I've tested that the new module is requesting bids.

This is behind a 0% test for the commercial and the trade desk folks to test also.

PR #1781, opened 2025-02-05
Remove The Trade Desk test and set live to all users

Jakeii:
## Why?
The test added in #1773 has been reviewed and everyone is happy, so we can launch.


Jakeii:
> LGTM
> 
> Some of the others seem to have switches associated with them - would we want a switch to control this or are we happy with just code?

There's a switch check for all bidders here https://github.com/guardian/commercial/pull/1781/files#diff-c700ce5a4ebaf41a57442961ac50d39a2b202c557699b0d6b192d34437c023d9L697

Looks like some of the other bidders had a redundant switch check in their `shouldInclude` function

PR #1782, opened 2025-02-05
Move the initPermutive call

emma-imber:
## What does this change?
Moves the `initPermutive()` call from `consented-advertising` to be a dependency of `fillStaticAdvertSlots`.

## Why?
At the moment, `initPermutive()` has to happen before `prepareGoogletag`, so it blocks the loading of the googletag script. By moving it to be a dependency of `fillStaticAdvertSlots`, we can load the googletag script a little earlier.

The [documentation states](https://support.permutive.com/hc/en-us/articles/360011779239-Deploying-the-Permutive-JavaScript-Tag#:~:text=The%20JavaScript%20required%20for%20this%20is%20shown%20below.%20This%20script%20should%20be%20placed%20after%C2%A0googletag%C2%A0%C2%A0is%20defined%20but%20before%20a%20call%20is%20made%20to%C2%A0googletag.enableServices()) that the `initPermutive()` code must run before `googletag.enableServices()` is called. `googletag.enableServices()` is called by `displayAds` and `displayLazyAds`, both of which are only called by `fillStaticAdvertSlots`, so as long as initPermutive runs before that, we should be all good.

### DIFF
--- a/.changeset/unlucky-lizards-hide.md
+++ b/.changeset/unlucky-lizards-hide.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial': minor
+---
+
+Move the initPermutive call so that the googletag script can be loaded a little earlier
--- a/src/init/consented-advertising.ts
+++ b/src/init/consented-advertising.ts
@@ -9,7 +9,6 @@ import { init as initIpsosMori } from './consented/ipsos-mori';
 import { init as initMessenger } from './consented/messenger';
 import { init as prepareA9 } from './consented/prepare-a9';
 import { init as prepareGoogletag } from './consented/prepare-googletag';
-import { initPermutive } from './consented/prepare-permutive';
 import { init as preparePrebid } from './consented/prepare-prebid';
 import { removeDisabledSlots as closeDisabledSlots } from './consented/remove-slots';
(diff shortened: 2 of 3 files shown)


dskamiotis (approved):
This looks sensible to me, especially made clearer given the refactoring work you did in the PR: https://github.com/guardian/commercial/pull/1764

Would be interesting to see the metrics we can use to test (assuming in CODE as you mentioned?)

src/experiments/tests/move-permutive-segmentation.ts
```diff
@@ -0,0 +1,29 @@
+import type { ABTest } from '@guardian/ab-core';
+
+export const movePermutiveSegmentation: ABTest = {
+	id: 'MovePermutiveSegmentation',
+	author: '@commercial-dev',
+	start: '2025-01-14',
+	expiry: '2025-02-28',
+	audience: 0 / 100,
+	audienceOffset: 20 / 100,
```
dskamiotis, line 9: I know its a 0% test, Wondered what this does?

emma-imber: The offset is to make sure that if we set it live (eg not 0%) that the test audience wouldn't overlap with that of your test

PR #1786, opened 2025-02-06
Replace commas with semicolons in Permutive setup with googletag

cemms1:
## What does this change?

- Replaces commas with semicolons
- Moves eslint disable comments above the line they refer to

## Why?

Was it a mistake that the permutive setup commands ended with commas instead of semicolons?

From when permutive was added (https://github.com/guardian/frontend/commit/09675cc7df4b50cb64f56ed74a9864fd3183980c#diff-3553e1025635aeb071b2bfe8d9b01ee48d9d27adf6597350bf315f5dae03648b) it's always had commas instead of semicolons 🤔 

This is unlikely to affect anything, since these commands probably still run fine with commas instead of semicolons, but this brings us closer to the way the docs recommend: https://developer.permutive.com/docs/ad-manager-dfp#javascript-changes

### DIFF
--- a/src/init/consented/prepare-permutive.ts
+++ b/src/init/consented/prepare-permutive.ts
@@ -222,20 +222,21 @@ const initPermutiveSegmentation = () => {
 		'359ba275-5edd-4756-84f8-21a24369ce0b',
 		{},
 	);
+
 	/* eslint-enable */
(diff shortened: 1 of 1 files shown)


dskamiotis (approved):
This makes sense, easier to read, not sure of why commas in the first place!

Jakeii (approved):

PR #1791, opened 2025-02-12
Remove Improve Digital Prebid adaptor

cemms1:
## What does this change?

Removes the code for Improve Digital

## Why?

We've had the switch turned off in frontend for a while so this hasn't been running.
There's appetite to reduce the vendor list in Sourcepoint and this is a candidate for removal


emma-imber (approved):
Nice!

PR #1811, opened 2025-02-20
Improve E2E testing of prebid

Jakeii:
## What does this change?
Expand our e2e tests to better verify prebid is running correctly.

checks include:
* That the prebid.js script is loaded
* That bids are requested for our bidders
* That no bidderErrors occur, excluding timeouts as these are fairly common and unlikely to be because of our own code.

## Why?
Confidence when making changes to prebid

### DIFF
--- a/playwright/tests/prebid.spec.ts
+++ b/playwright/tests/prebid.spec.ts
@@ -1,0 +1,75 @@
+import { expect, test } from '@playwright/test';
+import type { PrebidAuctionInitEvent } from '../../src/lib/header-bidding/prebid-types';
+import { articles } from '../fixtures/pages';
+import { cmpAcceptAll } from '../lib/cmp';
+import { loadPage } from '../lib/load-page';
+
+const testPage = articles[0];
+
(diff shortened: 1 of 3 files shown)


playwright/tests/prebid.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+			'oxd',
+			'and',
+			'pubmatic',
+			'ix',
+			'adyoulike',
+			'ozone',
+			'xhb',
+			'criteo',
+			'ttd',
+			'rubicon',
```
Amouzle, line 70: Do you need to specific this is UK, to avoid issues for US/AUS IPs?

arelra (approved):
Great to see these tests expanded, nice work! ✨

playwright/tests/prebid.spec.ts
Jakeii: For future travellers, we discussed in standup and will cover this in our canaries as we don't have multi-region testing in our CI.

PR #1821, opened 2025-02-25
Remove top-above-nav at tablet breakpoint on consentless

i-hardy:
## What does this change?

This removes the top-above-nav slot at the tablet breakpoint in the consentless flow.

## Why?

Since we cannot pass ad sizes to OptOut in the same way as we do Google, the top-above-nave ad in this slot was served at desktop breakpoint size, causing horizontal scroll. The proportion of unconsented users on a tablet browser is extremely small (and likely to become smaller with consent or pay), so simply removing the slot as we do on mobile is a simpler solution than either trying to implement tablet top-above-nav ads in OptOut, or adding styling to cut off oversized ads with possible knock-on effects on consented advertising.

**Before**
Screenshot 2025-02-25 at 14 06 45

**After**
Screenshot 2025-02-25 at 14 06 54

### DIFF
--- a/src/init/consentless/init-fixed-slots.ts
+++ b/src/init/consentless/init-fixed-slots.ts
@@ -2,12 +2,18 @@ import { getCurrentBreakpoint } from '../../lib/detect/detect-breakpoint';
 import { removeDisabledSlots } from '../consented/remove-slots';
 import { defineSlot } from './define-slot';
 
+const isTabletOrMobile = () => {
+	const breakpoint = getCurrentBreakpoint();
+	return breakpoint === 'mobile' || breakpoint === 'tablet';
+};
+
 const initFixedSlots = async (): Promise<void> => {
 	await removeDisabledSlots();
 
-	const isDCRMobile =
-		window.guardian.config.isDotcomRendering &&
-		getCurrentBreakpoint() === 'mobile';
+	// We remove top-above-nav on both tablet and mobile for consentless
+	// due to issues with ad sizing in OptOut at the tablet breakpoint
+	const shouldHideTopAboveNav =
+		window.guardian.config.isDotcomRendering && isTabletOrMobile();
 
 	const adverts = [
 		...document.querySelectorAll<HTMLElement>(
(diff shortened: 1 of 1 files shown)


src/init/consentless/init-fixed-slots.ts
```diff
@@ -2,12 +2,18 @@ import { getCurrentBreakpoint } from '../../lib/detect/detect-breakpoint';
...
+
 const initFixedSlots = async (): Promise<void> => {
 	await removeDisabledSlots();
 
-	const isDCRMobile =
-		window.guardian.config.isDotcomRendering &&
-		getCurrentBreakpoint() === 'mobile';
+	// We remove top-above-nav on both tablet and mobile for consentless
+	// due to issues with ad sizing in OptOut at the tablet breakpoint
+	const shouldHideTopAboveNav =
```
deedeeh, line 15: Don't you think `isDCRMobileOrTablet` might be a bit more obvious than `shouldHideTopAboveNav`?

src/init/consentless/init-fixed-slots.ts
```diff
@@ -17,9 +23,19 @@ const initFixedSlots = async (): Promise<void> => {
 		// we need to not init top-above-nav on mobile view in DCR
```
deedeeh, line 23:
```suggestion
		// we need to not init top-above-nav on mobile and tablet views in DCR
```

src/init/consentless/init-fixed-slots.ts
```diff
@@ -17,9 +23,19 @@ const initFixedSlots = async (): Promise<void> => {
...
-			(adSlot) => !(isDCRMobile && adSlot.id === 'dfp-ad--top-above-nav'),
+			(adSlot) =>
+				!(
+					shouldHideTopAboveNav &&
+					adSlot.id === 'dfp-ad--top-above-nav'
+				),
 		);
 
+	if (getCurrentBreakpoint() === 'tablet') {
+		// We need to explicitly remove the element as there are no styles to hide it on tablet
```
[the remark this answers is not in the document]
i-hardy: It happens in DCR because we hide top-above-nav on mobile for _both_ consented and consentless users. In this case we only want to hide the slot on the consentless flow (due to us being unable to serve appropriately sized ads via OptOut), and we don't know at the point of CSS render whether the user is consented or not.

While we could hide the element via CSS rather than removing it, we would still only be able to do so somewhere in this code where we know the user has opted out, and with the same effect of layout shift where the space for the ad disappears.

PR #1797, opened 2025-02-13
Check individual vendor consent for Prebid adaptors

cemms1:
## What does this change?

Check individual consent for each Prebid adaptor in the `requestBids` Prebid function so that we don't request bids for ad slots from bidders the user has not consented to 

This PR no longer checks for consent in the `setupPrebid` function since this config needs to be updated if consent changes during a session. This should be done separately

## Why?

[Trello ticket](https://trello.com/c/kz462thB/3225-add-individual-consent-checks-to-each-prebid-adapter)

We want to check the consent state for each bid adaptor rather than relying on the bid adaptor to check consent themselves. This will enable us to remove the global Prebid custom vendor check, reducing the vendor list and allowing more granular control over which adaptors can bid on a page.

> [!NOTE]
> This PR needs https://github.com/guardian/csnx/pull/1939 to be released first

PR #1825, opened 2025-02-27
Add consent helper to make it easier to check consent for TCF vendors safely

cemms1:
## What does this change?

Adds `hasConsentFor` function to wrap the `@guardian/libs` `getConsentFor` function that also takes into account the non TCF frameworks

Uses this in the Prebid `utils` file helpers

There are quite a few instances where we want to use the same logic so we can start to use this function in other places in this repo if successful

## Why?

There's a possible mistake in https://github.com/guardian/commercial/pull/1797 because it only checks consent for vendors in TCF regions, without taking into account whether it needs to or not

### DIFF
--- a/src/lib/consent.ts
+++ b/src/lib/consent.ts
@@ -1,0 +1,25 @@
+import {
+	type ConsentState,
+	getConsentFor,
+	type VendorName,
+} from '@guardian/libs';
+
+/**
+ * Helper function to check consent for a known TCF vendor that also
+ * takes into account global consent frameworks (USNAT & AUS)
+ * */
+export const hasConsentFor = (
+	vendorName: VendorName,
+	consentState: ConsentState,
+): boolean => {
+	switch (consentState.framework) {
+		case 'tcfv2':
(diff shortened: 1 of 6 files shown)


src/lib/consent.ts
```diff
@@ -0,0 +1,25 @@
...
+	type ConsentState,
+	getConsentFor,
+	type VendorName,
+} from '@guardian/libs';
+
+/**
+ * Helper function to check consent for a known TCF vendor that also
+ * takes into account global consent frameworks (USNAT & AUS)
+ * */
+export const hasConsentFor = (
```
arelra, line 11: I'm not sure if this is exactly what you want, but the cmp already does the appropriate checks for each regional framework

https://github.com/guardian/csnx/blob/47c02c76f65a53d638718e5777cec52341a8f398/libs/%40guardian/libs/src/consent-management-platform/getConsentFor.ts

arelra: I see you found it 👍

arelra: Easily done!

PR #1839, opened 2025-03-04
Dina/prebid bidcache ab test

deedeeh:


src/lib/header-bidding/prebid/prebid.ts
```diff
@@ -295,6 +297,11 @@ const timeoutBuffer = 400;
  */
 const bidderTimeout = PREBID_TIMEOUT;
 
+/**
+ * useBidCache is false by default then we will set it to true when switch in on and user is in variant
+ */
+const useBidCache = false;
+
```
Jakeii, line 304: could the `shouldIncludePrebidBidCache` check go here?

deedeeh: I am not really sure why we do that from L368-L375 but is there another way to do that? You mean otherwise use `window.guardian.pbjs.useBidCache` instead of assigning them to a variable `pbjsConfig`?

Jakeii: Charlotte's suggestion is what I was thinking!

deedeeh: @Jakeii I have no idea why I thought about something else or maybe I misread your comment! Thank you @cemms1 for the call to clearly explain that. Done in https://github.com/guardian/commercial/pull/1839/commits/574a3151ae51da6dcfc3d0a13da92d38ee7cb6ba

PR #1846, opened 2025-03-05
Fix Prebid tests

cemms1:
## What does this change?

- Un-skips Prebid tests in `prebid.spec`
- Removes unnecessary `mockConsentState` since it's not currently used in `prebid.ts`
- Removes unnecessary Permutive tests that refer to logic that now lives outside of this file. replaces these tests with new ones that reflect the current logic

## Why?

Keeping test coverage of Prebid up to date after a mistake removing it


emma-imber (approved):
🧹

PR #1847, opened 2025-03-05
Handle DuckDuckGo Mobile errors 🦆

emma-imber:
## What does this change?
Filters out `loadScript` and `createAdvert` errors that are caused by the DuckDuckGo Mobile browser to reduce noise on Sentry. I've also updated the comments from the ones I made a few years back to explain where the string values we look for come from.

## Why?
DuckDuckGo strikes again. Found an error in Sentry with these stats:

Screenshot 2025-03-05 at 16 42 19
Screenshot 2025-03-05 at 16 42 23

DuckDuckGo is massively overrepresented in Sentry errors because it blocks ads by default, so we filter out the desktop browsers to reduce noise in Sentry. The mobile browser is now causing similar issues, so it's worth filtering these out too.

### DIFF
--- a/src/define/create-advert.ts
+++ b/src/define/create-advert.ts
@@ -20,7 +20,13 @@ const createAdvert = (
 
 		log('commercial', errMsg);
 
-		if (!navigator.userAgent.includes('DuckDuckGo')) {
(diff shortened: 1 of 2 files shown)


deedeeh (approved):
Nice to reduce noise in Sentry!

PR #1848, opened 2025-03-05
Add keywords to ortb2.site.ext.data in Prebid config

emma-imber:
## What does this change?
Adds page keywords to `ortb2.site.ext.data` in the Prebid bid config. Example of how this looks:

Screenshot 2025-03-05 at 17 51 29

## Why?
Magnite and Index access keywords through `ortb2.site.ext.data`, and these are the bidders that we want to receive keywords. We're not adding keywords to `ortb2.site` because there is no revenue benefit.


deedeeh (approved):
Looks Good Emma 🚀

PR #1850, opened 2025-03-06
Check consent before setting Prebid bidder config

cemms1:
## What does this change?

- Checks consent before setting bidder config when initialising Prebid
  - Adds tests for setting config apart from calling the `window.pbjs.setBidderConfig` which is difficult to mock 
- Refactors various `shouldInclude` functions into one function with a switch statement
  - Updates tests for expected behaviour for `shouldInclude` functionality

## Why?

This gives us more granular control over which Prebid bidders can run for various consent choices and will eventually allow us to remove the fake global Prebid vendor from our vendor list in SP

### DIFF
--- a/src/lib/header-bidding/utils.ts
+++ b/src/lib/header-bidding/utils.ts
@@ -198,7 +198,6 @@ export const shouldIncludeBidder =
 					!isInUsOrCa()
 				);
 			case 'ozone':
-				// TODO: Check is we want regional restrictions on where we load the ozoneBidAdapter
 				return (
 					isSwitchedOn('prebidOzone') &&
 					getConsentFor('ozone', consentState) &&


emma-imber (approved):
This is definitely nicer! 👏 Love to see the bidders in alphabetical order now

src/lib/header-bidding/utils.ts
```diff
@@ -161,63 +161,80 @@ export const getRandomIntInclusive = (
...
+			case 'kargo':
+				return isSwitchedOn('prebidKargo') && isInUsa();
+			case 'oxd':
+				return (
+					isSwitchedOn('prebidOpenx') &&
+					getConsentFor('openX', consentState) &&
+					!isInUsOrCa()
+				);
+			case 'ozone':
+				// TODO: Check is we want regional restrictions on where we load the ozoneBidAdapter
```
Jakeii, line 201:
```suggestion
				// TODO: Check if we want regional restrictions on where we load the ozoneBidAdapter
```

Jakeii (approved):
Great stuff!

PR #1865, opened 2025-03-13
Update peer dependency ranges for core-web-vitals and identity-auth packages

cemms1:
## What does this change?

Updates `peerDependencies` to match the versions listed in `dependencies` or `devDependencies`

Also adds changeset to bump commercial package as a major release due to major changes in the peer dependencies

## Why?

We merged some Dependabot PRs to bump package versions but this did not automatically update the peer dependency ranges for these packages.


emma-imber (approved):
Good spot!

PR #1867, opened 2025-03-14
Fix prebid ozone/index/criteo custom price bucketing

Jakeii:
## What does this change?
Restore width/height price bucketing for ozone and index

This can be done by overriding the `hb_pb` targeting for these bidders rather than modifying the prebid source code.
https://docs.prebid.org/dev-docs/publisher-api-reference/bidderSettings.html

Added test cases for this.

## Why?
This was accidently removed when I [deprecated our custom prebid](https://github.com/guardian/commercial/pull/1737) `@guardian/prebid` (it wasn't in the readme but I should have checked!)

We had added `guCustomPriceBuckets` long ago in https://github.com/guardian/Prebid.js/pull/126 which allowed us to set the priceBuckets based on the creative size.

Now this can be done without needing to modify the prebid.js source code as mentioned above.

The background on this is that we have different a price bucket for these bidders to reduce the line-item count.

Without this, for certain edge cases (now in the unit test) it would be possible for the price buckets to not exist in GAM for a winning bid from ozone/index and not serve from GAM when it should have.

### DIFF
--- a/src/lib/header-bidding/prebid/prebid.ts
+++ b/src/lib/header-bidding/prebid/prebid.ts
@@ -262,14 +262,6 @@ declare global {
 				bidders: BidderCode[];
 				config: {
 					customPriceBucket?: PrebidPriceGranularity;
-					/**
-					 * This is a custom property that has been added to our fork of prebid.js
-					 * to select a price bucket based on the width and height of the slot.
-					 */
-					guCustomPriceBucket?: (bid: {
-						width: number;
-						height: number;
-					}) => PrebidPriceGranularity | undefined;
 				};
 			}) => void;
 			getConfig: (item?: string) => PbjsConfig & {
@@ -465,17 +457,14 @@ const initialise = (window: Window, consentState: ConsentState): void => {
 			adserverTargeting: [
 				{
 					key: 'hb_pb',
-					val({ bidder, width, height, cpm, pbCg }) {
-						if (bidder === 'ozone') {
-							return overridePriceBucket(
-								'ozone',
-								width,
-								height,
-								cpm,
-								pbCg,
-							);
-						}
-						return pbCg;
+					val({ width, height, cpm, pbCg }) {
+						return overridePriceBucket(
+							'ozone',
+							width,
+							height,
+							cpm,
+							pbCg,
+						);
 					},
 				},
 			],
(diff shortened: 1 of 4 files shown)


src/lib/header-bidding/prebid/prebid.ts
```diff
@@ -450,45 +461,48 @@ const initialise = (window: Window, consentState: ConsentState): void => {
...
+						if (bidder === 'ozone') {
+							return overridePriceBucket(
+								'ozone',
+								width,
+								height,
+								cpm,
+								pbCg,
+							);
+						}
+						return pbCg;
```
Jakeii, line 478: `pbCg` is the custom price from the main `customPriceBuckets` config option

emma-imber (approved):
Good catch, and nice to have unit tests for this!

src/lib/header-bidding/prebid/prebid.ts
```diff
@@ -448,59 +455,62 @@ const initialise = (window: Window, consentState: ConsentState): void => {
...
-						'commercial',
-						`Custom Prebid - Ozone price bucket for size (${width},${height}):`,
-						ozoneGranularity,
-					);
-					return ozoneGranularity;
+		window.pbjs.bidderSettings.ozone = {
+			adserverTargeting: [
+				{
+					key: 'hb_pb',
+					val({ width, height, cpm, pbCg }) {
```
arelra, line 482: While this works but I find this syntax a little 'exotic'.

Might this be a little less surprising?

```suggestion
					val: ({ width, height, cpm, pbCg }) => {
```

src/lib/header-bidding/prebid/prebid.ts
```diff
@@ -162,35 +157,47 @@ type EnableAnalyticsConfig = {
...
+	bidderCode: string;
+	[x: string]: unknown;
+};
+
 // bidResponse expected types. Check with advertisers
 type XaxisBidResponse = {
-	appnexus: {
-		buyerMemberId: string;
+	appnexus?: {
+		buyerMemberId?: string;
```
arelra, line 177: Are these properties now optional?

arelra (approved):
Great find @Jakeii and a big improvement to the custom price bucketing.

Thanks for taking the time to add comprehensive tests too.

src/lib/header-bidding/prebid/prebid.ts
Jakeii: They were only used wrapped with `Partial`, this is more explicit I think

PR #1871, opened 2025-03-17
Fix the trade desk test expiry

Jakeii:
## What does this change?
Fix expiry date for the test in #1864 :S

## Why?
It's not 2024 anymore


deedeeh (approved):
We definitely need validation in AB testing!

PR #1878, opened 2025-03-18
Improve ophan record of test events

Jakeii:
## What does this change?
Refactor `ophanRecord` function

## Why?
Sentry is reporting that `window.guardian.ophan` is sometimes undefined here, perhaps not loaded yet or could be blocked so changing it to be optional.

We're also wrapping the ophan functions just for logging purposes that we don't really need, we can tidy it up.

I've also added a simple queue so that all events will be sent if ophan becomes available the next time it tries to record.


emma-imber (approved):
Makes sense!

PR #1879, opened 2025-03-18
Make sure firstvisit targeting is added for opt out ads

emma-imber:
## What does this change?
Checks consent state for `googletag` instead of `consentState.canTarget` before adding the `firstvisit` custom targeting value for opt out ads.

## Why?
This is because we check the consent for `googletag` before launching opt out ads, not `canTarget`, so by checking the same consent permissions, we can make sure the targeting is reliably added. I think following consent or pay, there's some misalignment between these consent permissions, which means in the UK the `firstvisit` targeting value is not currently being sent.

## Before
Screenshot 2025-03-18 at 18 16 15

## After
Screenshot 2025-03-18 at 18 17 43

### DIFF
--- a/src/lib/targeting/build-page-targeting.ts
+++ b/src/lib/targeting/build-page-targeting.ts
@@ -1,6 +1,6 @@
 import type { Participations } from '@guardian/ab-core';
 import type { ConsentState, CountryCode } from '@guardian/libs';
-import { cmp, getCookie, isString } from '@guardian/libs';
+import { cmp, getConsentFor, getCookie, isString } from '@guardian/libs';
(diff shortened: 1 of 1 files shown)


deedeeh (commented):
Looks good Emma 👍 I just want to ask about the screenshots as I can't see any differences between the payload before and after.

emma-imber:
> Looks good Emma 👍 I just want to ask about the screenshots as I can't see any differences between the payload before and after.

If you look between the 'edition' and 'k' entries, firstvisit should be in the 2nd screenshot and not the first

deedeeh (approved):

PR #1895, opened 2025-03-25
Remove non-Okta auth

cemms1:
## What does this change?

Removes non-Okta (cookies) style auth from the commercial codebase

## Why?

The Okta feature switch has been removed (https://github.com/guardian/frontend/pull/27770) so we are now fully switched over and there is no need for the legacy cookie-based auth sign in

See also https://github.com/guardian/dotcom-rendering/pull/13537

### DIFF
--- a/playwright/tests/consent.spec.ts
+++ b/playwright/tests/consent.spec.ts
@@ -56,32 +56,35 @@ test.describe('tcfv2 consent', () => {
 		await adSlotsArePresent(page);
 	});
 
-	test.skip(`Login as subscriber, reject all, load Opt Out, ad slots are not present on multiple page loads`, async ({
-		page,
-		// context,
-	}) => {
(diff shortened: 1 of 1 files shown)


src/lib/identity/api.ts
```diff
@@ -3,212 +3,53 @@ import type {
...
-	}
-
-	return {
-		headers: {
-			Authorization: `Bearer ${authStatus.accessToken.accessToken}`,
-			'X-GU-IS-OAUTH': 'true',
-		},
-	};
-};
+const isUserLoggedIn = (): Promise<boolean> => getIdentityAuth().isSignedIn();
```
Jakeii, line 35: I know it's not something you've changed, but does this hit the API every time it's called?

Jakeii (approved):
Nice, lots of deletion!

PR #1898, opened 2025-03-26
Add GumGum winning bid logging for Amazon advertisements

dskamiotis:
## What does this change?
This PR introduces functionality to log GumGum winning bids for Amazon advertisements. 
The implementation detects specific Amazon advertisement IDs and fetch bid responses (amznp ID) associated with GumGum, and sends this data to a logging endpoint. 
The logic is encapsulated in a newly created module: `gumgum-winning-bid.ts`

#### Detection:
- The renderAdvert function checks if:
- The `advertiserId` matches the Amazon advertiser ID for GumGum (4751525411).
- The amznp ID in `window.guardian.a9WinningBids` equals 1lsxjb4.
#### Logging:
- If the conditions are met, the `logGumGumWinningBid` function is called with:
- The ad slot ID (advert.node.id).
- The advertiser ID (slotRenderEndedEvent.advertiserId).
#### Endpoint:
- The logging endpoint dynamically switches between development and production environments:
- Development: //logs.code.dev-guardianapis.com/log
- Production: //logs.guardianapis.com/log
#### Cleanup:
- A `beforeunload` event listener clears `window.guardian.a9WinningBids` to prevent stale data from persisting across page navigations.

## Why?
Tracking GumGum Performance: This change enables tracking of GumGum winning bids for Amazon advertisements, providing valuable insights into ad performance.
Centralised Logging: By creating a dedicated module for GumGum logging, the logic is reusable and easier to maintain.

## Testing variant
- using `http://localhost:3030/Article/http:/localhost:9000/sport/2023/nov/22/ronnie-osullivan-warns-he-will-quit-snooker-if-he-cannot-play-in-china#ab-A9BidResponseWinner=variant`
- specifically tricky adding `#ab-A9BidResponseWinner=variant` (this is the test definition description in the commercial src not the switch name in frontend.

### DIFF
--- a/src/events/render-advert.ts
+++ b/src/events/render-advert.ts
@@ -2,6 +2,7 @@ import type { Advert } from '../define/Advert';
 import { adSizes } from '../lib/ad-sizes';
 import { reportError } from '../lib/error/report-error';
 import fastdom from '../lib/fastdom-promise';
+import { logGumGumWinningBid } from '../lib/gumgum-winning-bid';
 import { renderAdvertLabel } from './render-advert-label';
 
 /**
@@ -200,6 +201,26 @@ const renderAdvert = (
 	advert: Advert,
 	slotRenderEndedEvent: googletag.events.SlotRenderEndedEvent,
 ): Promise<boolean> => {
+	const isA9GumGum =
+		(window.guardian.a9WinningBids?.[0]?.['amznp'] ?? '') === '1lsxjb4';
+
+	if (
+		slotRenderEndedEvent.advertiserId != null &&
+		slotRenderEndedEvent.advertiserId === 4751525411 &&
+		isA9GumGum
+	) {
+		const adSlotId = advert.node.id;
+		logGumGumWinningBid(
+			adSlotId,
+			slotRenderEndedEvent.advertiserId.toString(),
+		);
+	}
+
+	window.addEventListener('beforeunload', () => {
+		console.log(`User is clearing data in the browser`);
+		window.guardian.a9WinningBids = [];
+	});
+
 	addContentClass(advert.node);
 	return hasIframe(advert.node)
 		.then((isRendered) => {
--- a/src/events/render-advert.ts
+++ b/src/events/render-advert.ts
@@ -201,8 +201,11 @@ const renderAdvert = (
 	advert: Advert,
 	slotRenderEndedEvent: googletag.events.SlotRenderEndedEvent,
 ): Promise<boolean> => {
-	const isA9GumGum =
-		(window.guardian.a9WinningBids?.[0]?.['amznp'] ?? '') === '1lsxjb4';
+	const matchingAd = window.guardian.a9WinningBids?.find(
+		(bidResponse) => bidResponse.slotID == advert.id,
+	);
+
+	const isA9GumGum = matchingAd?.amznp === '1lsxjb4';
 
 	if (
 		slotRenderEndedEvent.advertiserId != null &&
--- a/src/events/render-advert.ts
+++ b/src/events/render-advert.ts
@@ -218,11 +218,7 @@ const renderAdvert = (
 
 		const isA9GumGum = matchingAd?.amznp === '1lsxjb4';
 
-		if (
-			slotRenderEndedEvent.advertiserId != null &&
-			slotRenderEndedEvent.advertiserId === 4751525411 &&
-			isA9GumGum
-		) {
+		if (slotRenderEndedEvent.advertiserId === 4751525411 && isA9GumGum) {
 			const adSlotId = advert.node.id;
 			logGumGumWinningBid(
 				adSlotId,
--- a/src/events/render-advert.ts
+++ b/src/events/render-advert.ts
@@ -205,7 +205,7 @@ const renderAdvert = (
 ): Promise<boolean> => {
 	const isInA9BidResponseWinnerTest = isUserInVariant(
 		a9BidResponseWinner,
-		'control',
+		'variant',
 	);
 
 	if (
--- a/src/events/render-advert.ts
+++ b/src/events/render-advert.ts
@@ -219,11 +219,6 @@ const renderAdvert = (
 		);
 	}
 
-	window.addEventListener('beforeunload', () => {
-		console.log(`User is clearing data in the browser`);
-		window.guardian.a9WinningBids = [];
-	});
-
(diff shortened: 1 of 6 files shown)


src/events/render-advert.ts
```diff
@@ -200,6 +201,26 @@ const renderAdvert = (
 	advert: Advert,
 	slotRenderEndedEvent: googletag.events.SlotRenderEndedEvent,
 ): Promise<boolean> => {
+	const isA9GumGum =
+		(window.guardian.a9WinningBids?.[0]?.['amznp'] ?? '') === '1lsxjb4';
```
Jakeii, line 205: You need find the bid response for this specific ad slot, `0` may not necessarily be this slot

dskamiotis: I was looking at the email from APS where they state:

"This amznp value denotes which bidder this bid is from, and you can use your Bidder Mapping sheet to see which specific bidder this maps to (this is available under Demand > Manage Demand in your UI)."

So from that I infered it was the winning bid, therefore I thought the 0'th being the only response in the array?
image

@Jakeii

Jakeii: right, but fetch bids requests ads for multiple slots (in our case, not in the screenshot)

dskamiotis: I see we are getting an array of bidResponse objects, I have dealt with this by matching each 'advert` being processed by the `render-advert` function to be matched to the array of bidResponses.

src/events/render-advert.ts
```diff
@@ -200,6 +201,29 @@ const renderAdvert = (
...
+		logGumGumWinningBid(
+			adSlotId,
+			slotRenderEndedEvent.advertiserId.toString(),
+		);
+	}
+
+	window.addEventListener('beforeunload', () => {
+		console.log(`User is clearing data in the browser`);
+		window.guardian.a9WinningBids = [];
+	});
```
Jakeii, line 225: what's this for?

src/lib/header-bidding/a9/a9.ts
```diff
@@ -74,7 +74,9 @@ const requestBids = async (
 		.then(
 			() =>
 				new Promise<void>((resolve) => {
-					window.apstag?.fetchBids({ slots: adUnits }, () => {
+					window.apstag?.fetchBids({ slots: adUnits }, (res) => {
+						console.log('Bids Response:', res);
+						window.guardian.a9WinningBids = res;
```
Jakeii, line 79: I think this can go in the commercial namespace `window.guardian.commercial`

Jakeii:
Just a couple small things, but otherwise looking good now👌

src/events/render-advert.ts
Jakeii: the window object doesn't persist across page loads.

dskamiotis: ok great so safe to remove this Im assuming 👍

src/types/global.ts
```diff
@@ -479,6 +495,7 @@ declare global {
 					) => void;
 				};
 			};
+			a9WinningBids?: FetchBidResponse[];
```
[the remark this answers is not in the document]
dskamiotis: nice spot 👍 that was the initial structure since changed

src/lib/header-bidding/a9/a9.ts
```diff
@@ -74,12 +76,29 @@ const requestBids = async (
...
+							if (
+								isInA9BidResponseWinnerTest &&
+								window.guardian.config.switches
+									.a9BidResponseWinner
+							) {
+								window.guardian.commercial =
+									window.guardian.commercial ?? {};
+								window.guardian.commercial.a9WinningBids =
+									bidResponse;
+							}
```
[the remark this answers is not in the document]
dskamiotis: The window object is used here to store state that needs to persist across later in the render-advert function. 
It keeps track of whether the winning bid from APS (Amazon Publisher Services) was GumGum. 
This state is later used in the render-advert function to confirm that the rendered ad matches the winning bid and that the advertiserId corresponds to Amazon. Then we have a match and are certain the rendered advert is indeed Gumgum.

src/lib/gumgum-winning-bid.ts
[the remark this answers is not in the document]
dskamiotis: I think we decided to send the data to fastly to be analysed by D&I, giving us the data we need to understand how often it serves more than once on the same pageview.

Unfortunately this data is not available in DFP logs, which is linked to page views,  that only shows it’s a Amazon Ad that’s won; and the Amazon APS logs which does show that gumgum won, doesn’t show pageview id.

src/lib/header-bidding/a9/a9.ts
```diff
@@ -74,12 +76,29 @@ const requestBids = async (
...
-							window.apstag?.setDisplayBids();
-							resolve();
-						});
-					});
+					window.apstag?.fetchBids(
+						{ slots: adUnits },
+						(bidResponse) => {
+							const isInA9BidResponseWinnerTest = isUserInVariant(
+								a9BidResponseWinner,
+								'control',
```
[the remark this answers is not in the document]
dskamiotis: yes that makes sense, its a 0% test but makes it more readable.

src/events/render-advert.ts
```diff
@@ -200,6 +203,34 @@ const renderAdvert = (
 	advert: Advert,
 	slotRenderEndedEvent: googletag.events.SlotRenderEndedEvent,
 ): Promise<boolean> => {
+	const isInA9BidResponseWinnerTest = isUserInVariant(
+		a9BidResponseWinner,
+		'control',
```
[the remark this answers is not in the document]
dskamiotis: again here especially for when we add a percent portion for the actual test

src/lib/header-bidding/a9/a9.ts
```diff
@@ -74,12 +76,29 @@ const requestBids = async (
...
+							);
+							if (
+								isInA9BidResponseWinnerTest &&
+								window.guardian.config.switches
+									.a9BidResponseWinner
+							) {
+								window.guardian.commercial =
+									window.guardian.commercial ?? {};
+								window.guardian.commercial.a9WinningBids =
+									bidResponse;
```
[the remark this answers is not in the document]
dskamiotis: aha I see that a neat approach from nullish coalescing operator to assignment op

PR #1900, opened 2025-03-28
Convert prebid analytics adapter typescript

Jakeii:
## What does this change?
Convert our custom analytics module to typescript, it had been copied over from our now deprecated prebid.js fork.

After this I have a follow up refactor in https://github.com/guardian/commercial/pull/1903 that improves the code a bit and adds tests.

## Why?
Typescript is the Bestscript


src/lib/header-bidding/prebid/price-config.ts
```diff
@@ -176,6 +176,14 @@ export const overridePriceBucket = (
 		bidder === 'criteo'
 			? criteoPriceGranularity
 			: getPriceGranularityForSize(bidder, width, height);
+
+	if (!priceGranularity) {
```
Jakeii, line 180: I'm not sure why this wasn't a problem before, but now it can be undefined

src/types/prebid.ts
```diff
@@ -0,0 +1,115 @@
+interface AnalyticsAdapterConfig {
```
Jakeii, line 1: I think we could move all prebid types here in a followup.

src/lib/header-bidding/prebid/modules/analyticsAdapter.ts
```diff
@@ -1,73 +1,32 @@
...
 const VERSION = 9;
 
-const analyticsType = 'endpoint';
-const SENDALL_ON = {};
-
-// Look there: http://jsben.ch/qhIE6
-SENDALL_ON[EVENTS.AUCTION_END] = true;
-SENDALL_ON[EVENTS.BID_WON] = true;
-
-let analyticsAdapter = Object.assign(adapter({ analyticsType }), {
```
Jakeii, line 24: There was a lot of "using things before they were defined" eslint errors so things have moved around a bit.

emma-imber (approved):

PR #1903, opened 2025-03-31
Refactor analyticsAdapter

Jakeii:
I'll layout the desired behaviour as it's quite hard to decipher from the old code.

1. Analytics adapter is initialised by extending the base analytics adapter
2. override the `handler` function for our custom analytics
3. queue up all events that we want to log
4. if a `auctionEnd` or `bidWon` event are received send the event with the rest of the queued events as a payload and clear the queue

Th rewrite modernises the code and adds unit tests and an e2e test.


src/lib/header-bidding/prebid/modules/analyticsAdapter.ts
```diff
@@ -91,235 +69,255 @@ function logEvents(events: EventData[]): void {
...
-	setSafely(event, 'aid', args.auctionId);
-	setSafely(event, 'bid', args.requestId);
-	return [event];
-};
-
-const trackAuctionInit: Handler = (adapter, args) => {
-	if (adapter.context) {
-		adapter.context.auctionTimeStart = Date.now();
+// Remove any properties that are undefined or null
+function createEvent(event: RawEventData): EventData {
```
Jakeii, line 77: This replaces `setSafely`

src/lib/header-bidding/prebid/modules/analyticsAdapter.ts
```diff
@@ -91,235 +69,255 @@ function logEvents(events: EventData[]): void {
...
-		setSafely(event, 'did', args.adId);
-		setSafely(event, 'cid', args.creativeId);
-		setSafely(event, 'sz', args.size);
-		setSafely(event, 'ttr', args.timeToRespond);
-		setSafely(event, 'lid', args.dealId);
+	return cleanedEvent as EventData;
+}
+
+// Handlers for each event type, these create a loggable event from prebid event
+const handlers = {
```
Jakeii, line 98: having the handlers like this gives us a few benefits, we can ensure that the function name matches the event and eliminates the switch statement

src/lib/header-bidding/prebid/modules/analyticsAdapter.ts
```diff
@@ -91,235 +69,255 @@ function logEvents(events: EventData[]): void {
...
+		const event = createEvent({
+			ev: 'nobid',
+			n: args.bidder ?? args.bidderCode,
+			bid: args.bidId ?? args.requestId,
+			sid: args.adUnitCode,
+			aid: args.auctionId,
+			ttr: duration,
+		});
 
-class AnalyticsQueue {
```
Jakeii, line 185: this class served no purpose really a simple array does the job.

src/lib/header-bidding/prebid/modules/analyticsAdapter.ts
```diff
@@ -91,235 +69,255 @@ function logEvents(events: EventData[]): void {
...
+				payload,
+			},
+		);
 	}
-}
+	return payload;
+};
 
 const analyticsAdapter = Object.assign(adapter({ analyticsType: 'endpoint' }), {
+	sendPayload: async (
```
Jakeii, line 223: the only reason this function is in the adapter is so it can be mocked for testing.

src/lib/header-bidding/prebid/modules/analyticsAdapter.ts
```diff
@@ -91,235 +69,255 @@ function logEvents(events: EventData[]): void {
...
 		}
 	},
 });
 
-analyticsAdapter.context = {};
-
 analyticsAdapter.originEnableAnalytics = analyticsAdapter.enableAnalytics;
 
 analyticsAdapter.enableAnalytics = (config: AnalyticsConfig): void => {
-	if (!config.options.ajaxUrl) {
```
Jakeii, line 266: typescript will ensure these are set

src/lib/header-bidding/prebid/modules/analyticsAdapter.ts
```diff
@@ -91,235 +69,255 @@ function logEvents(events: EventData[]): void {
...
 		pv: config.options.pv,
-		queue: new AnalyticsQueue(),
 	};
+
 	if (analyticsAdapter.originEnableAnalytics) {
 		analyticsAdapter.originEnableAnalytics(config);
 	}
 };
 
-analyticsAdapter.ajaxCall = function ajaxCall(data: string): void {
```
Jakeii, line 284: have switched to fetch, now called in `sendPayload`

playwright/tests/prebid.spec.ts
```diff
@@ -100,4 +101,28 @@ test.describe('Prebid', () => {
 
 		expect(bidderErrors).toHaveLength(0);
 	});
+
+	test('analytics should be called', async ({ page }) => {
```
emma-imber, line 105: Nice to see a test!

src/lib/header-bidding/prebid-types.ts
```diff
@@ -172,32 +169,176 @@ export type PrebidBidderErrorEvent = {
...
+		clickUrl?: string;
+	};
+}
+
+type TrackParams = {
+	eventType: string;
+	args: BidArgs;
+};
+
+const eventKeys = [
```
emma-imber, line 243: Would be nice to potentially have comments explaining what these stand for, as I can imagine that coming up in future, but could be a follow up!

src/lib/header-bidding/prebid/prebid.ts
```diff
@@ -513,15 +512,17 @@ const initialise = (window: Window, consentState: ConsentState): void => {
...
 				provider: 'gu',
 				options: {
-					ajaxUrl: window.guardian.config.page.isDev
-						? `//performance-events.code.dev-guardianapis.com/header-bidding`
-						: `//performance-events.guardianapis.com/header-bidding`,
-					pv: window.guardian.ophan?.pageViewId,
+					url:
+						window.guardian.config.page.isDev ||
+						window.location.hostname.includes('localhost')
+							? `//performance-events.code.dev-guardianapis.com/header-bidding`
```
emma-imber, line 523: Would be nice if we could reuse `headerBiddingAnalyticsUrl` here to avoid having these urls and logic defined in multiple places

src/lib/header-bidding/prebid/modules/analyticsAdapter.ts
```diff
@@ -91,235 +68,255 @@ function logEvents(events: EventData[]): void {
...
+					events,
+				},
+			);
+		}
+	},
 	track({ eventType, args }: { eventType: string; args: BidArgs }): void {
 		if (!analyticsAdapter.context) {
+			// this should never happen
+			reportError(
+				new Error('context is not defined, prebid event not be logged'),
```
emma-imber, line 261:
```suggestion
				new Error('context is not defined, prebid event not being logged'),
```

emma-imber (approved):
Definitely easier to read!

src/lib/header-bidding/prebid-types.ts
Jakeii: Good shout, there's a comment in the data platform models that I could steal them from actually!

PR #1914, opened 2025-04-07
Remove frontend crossword ads logic

emma-imber:
## What does this change?
Removes the client side ad insertion logic for the crossword mobile banner ad, since DCR handles this server side. Also removes the crossword-banner ad definition, since this slot doesn't exist in the DCR crossword implementation.

## Why?
DCR crosswords are now at 100%, so we can do a little bit of tidying to remove any remaining logic that's specifically for crossword ads on frontend.


Jakeii (approved):
🧹

PR #1925, opened 2025-04-10
Unify handling of the empty ad size and actually empty slots

i-hardy:
## What does this change?

This changes our handling of the 2x2 'empty' ad size to treat them in the same way as we do slots where nothing is returned from GAM/OptOut, and changes the collapsing of any kind of empty slot to use `display: none` rather than removing elements from the DOM. It also adds Playwright tests for this code path.

## Why?

We decided after some investigation that standardising around collapsing slots with `display: none` was more desirable for increased visibility. We also required tests for slot collapsing behaviour to help prevent a repeat of a regression in which slots for which no ad could be provided were not collapsed and remained as blank space on the page, but due to the differences in how genuinely empty ads and the 'empty' size were handled, it was not possible to test for how empty slots got collapsed.

This PR standardises slot collapsing behaviour across the board, makes it testable, and removes code duplication.

### DIFF
--- a/playwright/tests/collapse-slot.spec.ts
+++ b/playwright/tests/collapse-slot.spec.ts
@@ -7,9 +7,7 @@ import { loadPage } from '../lib/load-page';
 const { path } = articleWithCollapsedSlots;
 
 test.describe('Ad slot removal', () => {
-	test(`Empty ad slots should be removed from the DOM`, async ({
-		page,
-	}) => {
+	test(`Empty ad slots should be removed from the DOM`, async ({ page }) => {
 		const gamResponse = waitForGAMResponseForSlot(page, 'top-above-nav');
 
 		await loadPage(page, path);
--- a/playwright/tests/collapse-slot.spec.ts
+++ b/playwright/tests/collapse-slot.spec.ts
@@ -7,7 +7,9 @@ import { loadPage } from '../lib/load-page';
 const { path } = articleWithCollapsedSlots;
 
 test.describe('Ad slot removal', () => {
-	test(`Empty ad slots should be removed from the DOM`, async ({ page }) => {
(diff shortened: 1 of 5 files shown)


src/insert/comments-expanded-advert.ts
```diff
@@ -121,13 +121,8 @@ const removeMobileCommentsExpandedAds = (): Promise<void> => {
...
 			if (advert) {
-				window.googletag.destroySlots([advert.slot]);
+				emptyAdvert(advert);
 			}
-			node.remove();
-			dfpEnv.adverts.delete(node.id);
-			dfpEnv.advertsToLoad = dfpEnv.advertsToLoad.filter(
-				(_) => _ !== advert,
-			);
 		}),
```
GHaberis, line 126: Nice, it looks tidier now we've removed the `dfpEnv` logic from here as we use it via `emptyAdvert` now 👍

GHaberis (approved):
Looks good to me 👍

playwright/tests/collapse-slot.spec.ts
```diff
@@ -0,0 +1,55 @@
+import { expect, test } from '@playwright/test';
+import { articleWithCollapsedSlots } from '../fixtures/pages/articles';
+import { cmpAcceptAll } from '../lib/cmp';
+import { assertHeader, waitForGAMResponseForSlot } from '../lib/gam';
+import { loadPage } from '../lib/load-page';
+
+const { path } = articleWithCollapsedSlots;
+
+test.describe('Ad slot removal', () => {
+	test(`Empty ad slots should be removed from the DOM`, async ({ page }) => {
```
mchv, line 10: Maybe a small thing, but should the test now not be described as `Empty ad slots should be NOT removed from the DOM, but be invisible`

PR #1927, opened 2025-04-11
Add event type to invalid analytics payload error

emma-imber:
## What does this change?
Passes in the type of the first prebid event into the invalid analytics payload error to help with debugging. This is the data used by the code to check the validity of the payload, so it should help us see what the issue could be.

## Why?
At the moment the Sentry logs aren't giving much helpful info as the payload isn't readable in the error. See the screenshot of the payload data in the Sentry error below:

Screenshot 2025-04-11 at 14 16 57


deedeeh (approved):
It would be good to know the actual error so great Emma 👌🏼

PR #1941, opened 2025-04-22
 Prevent overwriting of A9 winning bids across multiple slots

dskamiotis:
## What does this change?
This PR updates the logic in a9.ts to prevent overwriting A9 winning bids when multiple slots are processed. Instead of replacing the existing a9WinningBids array with new bid responses, this change appends the new bids to the array using push(...bidResponse). This ensures that all winning bid objects persist across ad refreshes or when new slots come into view.
## Why?

Previously, each time fetchBids was called, the a9WinningBids array was overwritten with the latest bid response. This caused winning bids to be overwritten, as previous winning bids were not retained. By appending new bids to the array, we can keep track of all winning bid objects, ensuring accurate matching of bids to their respective slots during the rendering process.


src/lib/header-bidding/a9/a9.ts
```diff
@@ -85,8 +85,9 @@ const requestBids = async (
 							);
 							if (isInA9BidResponseWinnerTest) {
 								window.guardian.commercial ??= {};
-								window.guardian.commercial.a9WinningBids =
-									bidResponse;
+								window.guardian.commercial.a9WinningBids?.push(
+									...bidResponse,
+								);
```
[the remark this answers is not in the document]
Jakeii: > Do we definitely want to spread `bidResponse` rather than just pushing it to the array?

I think we do, it's an array of responses from the bidders, will make looking up the response we want easier later?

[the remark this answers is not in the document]
dskamiotis: Yes agreed, I believe each response from APS is a `bidResponse: FetchBidResponse[]`

src/lib/header-bidding/a9/a9.ts
```diff
@@ -85,8 +85,10 @@ const requestBids = async (
 							);
 							if (isInA9BidResponseWinnerTest) {
 								window.guardian.commercial ??= {};
-								window.guardian.commercial.a9WinningBids =
-									bidResponse;
+								window.guardian.commercial.a9WinningBids = [];
```
SiAdcock, line 88: I think resets the property every time we get a bid response, so we'd be back to square one? We only want to set `a9WinningBids` to a new array if it doesn't already exist. Apologies if I'm missing something.

dskamiotis: No you're absolutely right I was meant to do it like the object above on line 87!

dskamiotis: I have added window object screenshots in the description to help

src/lib/header-bidding/a9/a9.spec.ts
```diff
@@ -104,3 +110,110 @@ describe('initialise', () => {
 		expect(window.apstag?.init).toHaveBeenCalled();
 	});
```
dskamiotis, line 111: Since adding my separate tests for this file, it is now producing this check failure for the tests not part of my changes.
I have tried type asserting and creating a mock type but to no avail. I'm not entirely sure what the issue is here.

src/lib/header-bidding/a9/a9.spec.ts
```diff
@@ -104,3 +109,113 @@ describe('initialise', () => {
...
+		expect(window.guardian.commercial?.a9WinningBids).toMatchObject(
+			expect.arrayContaining([...mockBidresponse, ...mockBidresponse]),
+		);
+	});
+	it('should not add a9WinningBids to the window.commercial object if no bidResponse', () => {
+		mockOnConsentChange(tcfv2WithConsentMock);
+		mockGetConsentFor(true);
+		const result = a9.logA9BidResponse([]);
+		expect(result).toEqual(undefined);
+	});
```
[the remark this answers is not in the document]
dskamiotis: aha this makes total sense... its passing but not for the reasons we thought.
have updated and added new test 👍

PR #1948, opened 2025-04-24
try running dependabot daily

Jakeii:
## Why?
In theory this will allow package updates to trickle in, rather than a big wave once a week.


deedeeh (commented):
Looks good Jake 👍🏼 I just have a question do you think we want to open 10 pull requests daily or that number could be changed to less maybe?

deedeeh (approved):

PR #1955, opened 2025-04-29
Remove old Prebid references

cemms1:
## What does this change?

Removes references to our Guardian forked version of Prebid.js from this repo

## Why?

We missed some references to Prebid when moving over from our forked version to use the npm package directly

See https://github.com/search?q=org%3Aguardian+repo%3Aguardian%2Fcommercial+guardian%2Fprebid++NOT+is%3Aarchived&type=code

### DIFF
--- a/docs/header-bidding/readme.md
+++ b/docs/header-bidding/readme.md
@@ -201,19 +201,19 @@ Here is an example header bidding creative script which includes a parameter for
 
 You may have noticed that there are separate line items for each bid price.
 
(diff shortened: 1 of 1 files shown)


scripts/bump_commercial.sh
```diff
@@ -21,8 +21,6 @@ git pull origin main
 git checkout -b "$BRANCH_NAME"
```
Jakeii, line 21: could props delete this script all together!

Jakeii (approved):

PR #1966, opened 2025-05-06
Update `inlineOpponentSelector` rule to include nested images

deedeeh:
## What does this change?

This PR updates the inlineOpponentSelector rule in Spacefinder to avoid having an ad just after an image. There is a rule already for images which is included under `data-spacefinder-role=inline` but in this case it's nested so we had to update the rule to include the `inline` role when it's nested.

## Screenshots

| Before      | After      |
| ----------- | ---------- |
| Screenshot 2025-05-06 at 16 33 10 | Screenshot 2025-05-06 at 16 34 22 |
| Screenshot 2025-05-06 at 16 34 49 | Screenshot 2025-05-06 at 16 35 10 | 

## Why?

This will prevent having an ad straightaway after an image.

### DIFF
--- a/src/insert/spacefinder/rules.ts
+++ b/src/insert/spacefinder/rules.ts
@@ -52,7 +52,10 @@ const leftColumnOpponentSelector = ['richLink', 'thumbnail']
 const rightColumnOpponentSelector =
 	':scope > [data-spacefinder-role="immersive"]';
 const inlineOpponentSelector = ['inline', 'supporting', 'showcase', 'halfWidth']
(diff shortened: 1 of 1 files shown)


deedeeh:
> @deedeeh is it possible to get a screenshot or two here? To help my understanding of the change

I had a huge feeling that I will be asked to do that and it's entirely my fault that I haven't done that during the debugging/fix as I will need to point frontend to CAPI-CODE again to do that and it was busy for most of the day 😅. I will let you know once I am done with the screenshots.

Jakeii (approved):

PR #2000, opened 2025-05-22
Improve video progress reporting

Jakeii:
## What does this change?
Improve how video progress is reported, used for video interscrollers and fabric-video in the future.

Use the slot-rendered event to save the creativeId and creativeTemplateId on the corresponding `advert` class instance for later use.

The advert class instance can be laster retrieved by the slot id, this is a bit cleaner than the current use of the `googletag.pubads()` API

Rename the video interscroller lib and refactor to store a map of slotIds to progress which is then reported when the page is inactive.

## Why?
We need to start using this code for other use cases like fabric-video and fabric-video-xl, it's not impossible that several of these could appear on the page at once.


src/lib/video-progress-reporting.ts
```diff
@@ -0,0 +1,58 @@
...
+			return;
+		}
+		const { creativeId, lineItemId } = advert;
+
+		const progress = value ?? 0;
+
+		void fetch(endpoint, {
+			method: 'POST',
+			body: JSON.stringify({
+				label: 'commercial.videoProgress',
```
Jakeii, line 26: I've updated the report query with the new label

src/lib/video-progress-reporting.ts
```diff
@@ -0,0 +1,58 @@
...
+			return;
+		}
+		const { creativeId, lineItemId } = advert;
+
+		const progress = value ?? 0;
+
+		void fetch(endpoint, {
+			method: 'POST',
+			body: JSON.stringify({
+				label: 'commercial.interscroller.videoProgress',
```
Jakeii, line 26: changing this will break any current video reports so it'll have to keep interscroller in the label!

src/lib/messenger/background.ts
```diff
@@ -282,21 +259,21 @@ const setupBackground = async (
...
+				const advert = getAdvertById(adSlot.id);
 
-				if (creativeInfo) {
-					initVideoProgressReporting(
-						creativeInfo.creativeId,
-						creativeInfo.lineItemId,
-					);
-				}
+				const shouldReportVideoProgress =
+					advert?.creativeTemplateId === 11885667;
```
i-hardy, line 265: Maybe make this ID a constant to help explain what it is?

i-hardy (approved):
One small suggestion, otherwise looks good 👍

src/lib/messenger/background.ts
Jakeii: Good shout

PR #2012, opened 2025-05-28
Set up listener for Opinary poll events to send to Permutive

cemms1:
## What does this change?

Adds event listener for Opinary polls to send response data to Permutive

## Why?

We're integrating Opinary via custom GAM creatives so need to send the survey response data


GHaberis (approved):
Looks good 👍

PR #2032, opened 2025-06-12
Add `PrebidAdUnit` 0% test

deedeeh:
## What does this change?

This PR is grouping slots such as inlines excluding `inline1` and all `fronts-banner` to allow the highest benefits of bid caching in Prebid. The setup we have at the moment each `adUnitCode` in Prebid has a different value (`dfp-ad--inline-2`, `dfp-ad--inline-3`, etc...) but because it's used by bidCache we need to have similar value for `adUnitCode`. For more info check [cache bids](https://docs.prebid.org/dev-docs/faq.html#does-prebidjs-cache-bids).

The changes:
- New property `prebidAdUnit` in `Advert` class to be used in `PrebidAdUnit` class.
- `CustomSlotMatching` sets the slot name to it's original slotId before setting the targeting and send it to GAM.
- The changes will be applied if `useBidCache` switch is on and the user in variant otherwise the current setup renders.
- 0% test is created 

Unit tests will be good to have but I might do it in a separate PR but let me know if you have different opinion.

Tested locally and forced a bid with the use of `debugging` module in Prebid and it works for inline2+ and fronts-banners. I also checked the request parser to check the data sent to GAM and I am able to see the original slot name e.g. `dfp-ad--inline-2` plus the targeting `hb_bidder`.

## Why?

Increase the chances of having cached bids on multiple slots with same sizes which might result in revenue.

### DIFF
--- a/src/define/Advert.ts
+++ b/src/define/Advert.ts
@@ -127,8 +127,7 @@ const isSizeMappingEmpty = (sizeMapping: SizeMapping): boolean => {
 const getPrebidAdUnit = (advertId: string): string => {
 	if (advertId.includes('fronts-banner')) {
 		return 'banner';
-	} else if (advertId.match(/inline(?:[2-9]|1[0-9])/)) {
-		// Matches inline2 to inline19
+	} else if (advertId.includes('inline') && !advertId.includes('inline1')) {
 		return 'inline';
 	}
 	return advertId;
--- a/src/define/Advert.ts
+++ b/src/define/Advert.ts
@@ -124,8 +124,8 @@ const isSizeMappingEmpty = (sizeMapping: SizeMapping): boolean => {
 };
 
 //We are grouping all fronts-banner slots and subsequent inline slots except inline1 to get the most use of prebid bid caching
-const getGroupedSlot = (advertId: string): string => {
-	if (advertId.match(/fronts-banner/)) {
+const getPrebidAdUnit = (advertId: string): string => {
+	if (advertId.includes('fronts-banner')) {
 		return 'banner';
 	} else if (advertId.match(/inline(?:[2-9]|1[0-9])/)) {
 		// Matches inline2 to inline19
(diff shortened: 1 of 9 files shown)


src/define/Advert.ts
```diff
@@ -123,9 +123,21 @@ const isSizeMappingEmpty = (sizeMapping: SizeMapping): boolean => {
 	);
 };
 
+//We are grouping all fronts-banner slots and subsequent inline slots except inline1 to get the most use of prebid bid caching
+const getGroupedSlot = (advertId: string): string => {
+	if (advertId.match(/fronts-banner/)) {
```
Jakeii, line 128: not sure if regex's are necessary, `includes` might be better? Like https://github.com/guardian/commercial/blob/8bb34702d16943891069ac9930adbd392116e6d4/src/define/Advert.ts#L75

src/define/Advert.ts
```diff
@@ -123,9 +123,21 @@ const isSizeMappingEmpty = (sizeMapping: SizeMapping): boolean => {
...
+		// Matches inline2 to inline19
+		return 'inline';
+	}
+	return advertId;
+};
+
 class Advert {
 	id: string;
 	node: HTMLElement;
+	groupedSlot: string;
```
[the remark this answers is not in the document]
deedeeh: Yes we can change it but does that mean you want me to change it for the test name as well?

src/define/Advert.ts
Jakeii: It could be includes inline and !includes inline1? But up to you

deedeeh: d6867772d1f5cfc4749e6e523801485fa2c26cca 👍🏼

src/lib/header-bidding/prebid/prebid.ts
```diff
@@ -225,21 +226,30 @@ class PrebidAdUnit {
...
+					this.gpid,
+					consentState,
+				)
+			: bids(
+					advert.id,
+					slot.sizes,
+					pageTargeting,
+					this.gpid,
+					consentState,
+				);
```
Jakeii, line 243: Bit of a nitpick, feel free to ignore 😬
```suggestion
		this.bids = bids(
					shouldIncludePrebidAdUnit
						? advert.prebidAdUnit
						: advert.id,
					slot.sizes,
					pageTargeting,
					this.gpid,
					consentState,
				);
```

PR #2045, opened 2025-06-18
Add 1% variant for 8 days to PrebidMultibid test

deedeeh:
## What does this change?

This PR increases the `PrebidMultibid` test from 0% to 1% variant so in total 2% and it should take around 8 days to collect data. 

## Why?

We want to retest multibid after the changes that has been done in the setup https://github.com/guardian/commercial/pull/2008 to be confident we don't have any negative impact before we make further changes to group slots and test it separately.


deedeeh:
> Nice, very sensible!
> 
> Do you think we should set the expiry date a couple of days later just in case we haven't collected enough data?

I've given extra couple of days but yes maybe it's better to allow it more time as there is no harm to do that instead of having another PR to extend due to not having enough data. I will do that now.

PR #2050, opened 2025-06-20
Update build steps to support generating 2 bundles to test prebid.js updates

GHaberis:
## What does this change?

This PR introduce changes to the build steps to support generating 2 "commercial.js" bundles, one uses the existing  `prebid.js` dependency,  and another which swaps it out at build time for a version of our choosing, aka `prebidTest.js`. 

The idea is we can then set-up a server side AB test in `Frontend` so it switches the bundle referenced on the server depending on a user's participation in the test

Before settling on this approach I tested a few other approaches:

1. Client-side A/B test. This would require making imports of `prebid.js` dynamic and conditionally importing either the current or new prebid.js dependency. I felt this was too much of a deviation on the code running in the browser, and had the potential to create unknown side effects.
2. Server-side A/B test w/ multiple entry points. Instead of using multiple configs (as I have done here) we could have a single config with multiple entry points, I hoped this would mean the plugin logic could remain untouched. Unfortunately I couldn't target rules to entry points, there are multiple ways to target Webpack rules against imports but sadly I couldn't find anyway of targeting a rule conditionally for an entry point so abandoned this option. 

### Done
-  [x] Import `prebid.js@v9.46` as alias `prebidTest.js` dependency
-  [x] Set-up `webpack.config.prebidTest.mjs` to alias imports of `prebid.js` to `prebidTest.js`, also update existing `prebid.js` aliases and use the Babel config for `prebidTest.js` (aka `prebid.js@v9.46`).
-  [x] Update `webpack.config.dev.mjs` to read `process.env.BUILD_PREBID_TEST` and use appropriate config to generate dev bundle during dev. 
- [x] Update `webpack.config.prod.mjs` to export an array of Webpack configs rather than 1, the 2nd config is for the "Prebid Test".
- [x] Update `pnpm build:prod` to execute  a new`webpack.build.prod.mjs` script rather than using the `webpack` cli command and pointing it towards the `webpack.config.prod.mjs` config. I did this as it allowed me to execute the cloudformation/prout logic after _both_ bundles had been compiled. As plugins they would only execute after an individual config/bundle had been compiled, and they'd only have access to that config/bundle's compilation info. Updating the plugins to handle this felt brittle.
- [x] Refactor the cloudformation/prout plugins to be modules imported by the new `webpack.build.prod.mjs` script.

### To do (in follow up PRs?)
- webpack.config.prebidTest.mjs` is pretty much a copy of webpack.config.mjs`, I could probably extracxt common config and use webpack `merge` to compose with the non-common config.
- Improve DevX of working on webpack scripts, so we have linting, prettier and type safety?

### Testing
- [x] Confirm commercial bundle with `prebid.js@v9.27` runs locally using DCR
- [x] Confirm commercial bundle with `prebid.js@v9.46` runs locally using DCR
- [x] Confirm generated output for the 'Dev' and 'Prod' commands is correct, including cloudformation/prout files.
- [ ] Deploy to CODE and test 

### Note
A known limitation would be if the `prebid.js` update we're testing actually changes the interface we use in the source code, I've raised this and the likelihood of them introducing a breaking change is low.


Jakeii:
Sorry, I'm catching up after being away, I don't quite follow the reason a client side test is a bad idea? [pbjs](https://github.com/guardian/commercial/blob/main/src/lib/header-bidding/prebid/pbjs.ts) is [dynamically imported already](https://github.com/guardian/commercial/blob/main/src/init/consented/prepare-prebid.ts#L21), so it could be duplicated into a variant with the new version and conditionally imported?

GHaberis:
> Sorry, I'm catching up after being away, I don't quite follow the reason a client side test is a bad idea? [pbjs](https://github.com/guardian/commercial/blob/main/src/lib/header-bidding/prebid/pbjs.ts) is [dynamically imported already](https://github.com/guardian/commercial/blob/main/src/init/consented/prepare-prebid.ts#L21), so it could be duplicated into a variant with the new version and conditionally imported?

@Jakeii I started with looking at this option, but it wasn't as simple as copying `pbjs.ts` and conditionally importing the appropriate version. Any `prebid.js` imports further down the import tree would also require copying or being converted into dynamic imports. Perhaps there aren't too many of these? Maybe just the `./modules`? I ran into a few issues that I eventually pivoted to a server-side test. 

Here's the WIP branch I started looking at this: https://github.com/guardian/commercial/compare/main...gh-prebid-946. I'd be happy to revisit together?

GHaberis:
Closing this is we went with a client-side solution in the end: https://github.com/guardian/commercial/pull/2057

PR #2068, opened 2025-07-03
Split  out core modules into seperate package

Jakeii:
## What does this change?
Move all modules that are exported as a library to a new workspace `core`, some dependant modules also needed to be moved.

bundle files can still use the modules of course, but refer to them using the full package name `@guardian/commercial/some-module`

All the modules in core are now exported, as I think it makes sense for consistency if you can import `@guardian/commercial/xyz` from the bundle it should be available for other consumers.

I've used [conditional exports](https://nodejs.org/api/packages.html#conditional-exports) and aliases to tell webpack, jest, eslint that they should use the source typescript rather than the built files, so that imports update and the dev server updates without needing to rebuild core locally.

I've made use of [pnpm catelogs](https://pnpm.io/catalogs) so that we don't need to update packages in multiple places.

I've tested `@guardian/commercial` in DCR and Frontend.

And tested the bundle works as expected locally and in CODE.

## Why?
To make it more clear that changing these files will affect the `@guardian/commercial` library and require version bumps for consumers.


tsconfig.json
```diff
@@ -18,15 +18,22 @@
...
+		"moduleResolution": "bundler",
 		"noEmit": true,
 		"noUnusedLocals": true,
 		"resolveJsonModule": true,
 		"skipLibCheck": true,
 		"strict": true,
 		"target": "ESNEXT",
-		"noUncheckedIndexedAccess": true
+		"noUncheckedIndexedAccess": true,
+		"customConditions": ["development"]
```
Jakeii, line 30: This tells typescript to use the `development` field defined in package.json `exports` which for `@guardian/commercial` points to the source typescript.

tsconfig.json
```diff
@@ -18,15 +18,22 @@
 			"ES2020",
 			"ESNext"
 		],
-		"module": "ESNEXT",
-		"moduleResolution": "node",
+		"module": "preserve",
+		"moduleResolution": "bundler",
```
Jakeii, line 22: This is similar to `nodenext` in that it obeys package.json `exports` except that it doesn't require extensions in imports.

webpack.config.mjs
```diff
@@ -85,6 +85,7 @@ const config = {
 			),
 		},
 		extensions: ['.js', '.ts', '.tsx', '.jsx'],
+		conditionNames: ['development', 'import', 'default'],
```
Jakeii, line 88: This tells webpack to use the development field defined in package.json exports which for @guardian/commercial points to the source typescript.

src/types/global.ts
```diff
@@ -1,14 +1,7 @@
 import type { CustomClaims } from '@guardian/identity-auth';
```
Jakeii, line 1: Only moved the types needed by core from here

core/src/targeting/build-page-targeting.spec.ts
```diff
@@ -4,8 +4,8 @@ import type {
 	USNATConsentState,
 } from '@guardian/libs';
 import { cmp as cmp_, setCookie, storage } from '@guardian/libs';
-import { getAuthStatus as getAuthStatus_ } from '../../lib/identity/api';
-import type { AuthStatus } from '../../lib/identity/api';
+// import { getAuthStatus as getAuthStatus_ } from '../../lib/identity/api';
+// import type { AuthStatus } from '../../lib/identity/api';
```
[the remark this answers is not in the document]
Jakeii: Opps, the mock here turned out to be unnecessary and saved my having to move the identity stuff to core!

PR #2072, opened 2025-07-08
Add `consentManagementGpp` module

Jakeii:
## Why?
Since we switched to GPP in the US we haven't been passing consent to our prebid vendors for users there.

This wasn't a risk to us as we don't launch prebid without the requisite consent, but likely means our ad-tech partners may believe we’re not adequately collecting and/or passing consent on to them

Before:
Screenshot 2025-07-08 at 14 24 31

After:
Screenshot 2025-07-08 at 14 23 59

### DIFF
--- a/src/lib/header-bidding/prebid/pbjs-v9.46.0.ts
+++ b/src/lib/header-bidding/prebid/pbjs-v9.46.0.ts
@@ -2,7 +2,7 @@ import pbjs from 'prebid-v9.46.0.js';
 
 import 'prebid-v9.46.0.js/modules/adyoulikeBidAdapter';
 import 'prebid-v9.46.0.js/modules/consentManagementTcf';
-import 'prebid-v9.46.0.js/modules/consentManagementUsp';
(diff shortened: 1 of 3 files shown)


Jakeii:
> I think we might also need to update here
> 
> https://github.com/guardian/commercial/blob/7b8b9a668e58ef64d49033bc60ed9a969b68a65c/src/lib/header-bidding/prebid/prebid.ts#L47-L52
> 
> And possibly [here](https://github.com/guardian/commercial/blob/7b8b9a668e58ef64d49033bc60ed9a969b68a65c/src/lib/header-bidding/prebid/prebid.spec.ts#L252-L255)?

I don't think so, Usp is still used in Australia I think, and Gpp is already there https://github.com/guardian/commercial/blob/7b8b9a668e58ef64d49033bc60ed9a969b68a65c/src/lib/header-bidding/prebid/prebid.ts#L62

But that also means we need to keep the Usp module too...

PR #2077, opened 2025-07-10
Fix core build path

Jakeii:
## What does this change?
Fix Build Core artifact path

## Why?
The build path was incorrect causing changesets release not to work


dskamiotis (approved):
👍

PR #2079, opened 2025-07-10
Include `rtdModule` in Prebid set up

cemms1:
## What does this change?

- Includes `rtdModule` in our Prebid set up
- Reorganises the `pbjs.ts` file into sections to make it easier to recognise what each module is for

## Why?

We haven't been syncing data to Permutive recently so need to investigate why this is.

We've also started seeing a console warning in the browser: `pbjs.ts:24 fun-hooks: referenced 'realTimeData' but it was never created`

We should be including the `rtdModule` as part of our `permutiveRtdProvider` set up in Prebid, according to [the docs](https://docs.prebid.org/dev-docs/modules/permutiveRtdProvider.html)

| Before | After |
| ----- | ----- |
| ![before][] | ![after][] |

[before]: https://github.com/user-attachments/assets/23ada43f-ae04-4955-b050-5cd64eb8a044
[after]: https://github.com/user-attachments/assets/b158a9df-01ae-4f7a-8a9c-b657805f1618

### DIFF
--- a/src/lib/header-bidding/prebid/pbjs-v9.46.0.ts
+++ b/src/lib/header-bidding/prebid/pbjs-v9.46.0.ts
@@ -1,23 +1,31 @@
 import pbjs from 'prebid-v9.46.0.js';
 
-import 'prebid-v9.46.0.js/modules/adyoulikeBidAdapter';
+// Consent management
 import 'prebid-v9.46.0.js/modules/consentManagementTcf';
 import 'prebid-v9.46.0.js/modules/consentManagementUsp';
 import 'prebid-v9.46.0.js/modules/consentManagementGpp';
+
+// Bid adapters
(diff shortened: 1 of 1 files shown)


Jakeii:
> We did used to have this in the old frontend setup (see [guardian/frontend#23810](https://github.com/guardian/frontend/pull/23810), [guardian/Prebid.js#117](https://github.com/guardian/Prebid.js/pull/117)) so must have missed this when migrating to commercial or when starting to use Prebid.js directly rather than from our fork of it
> 
> EDIT: I think something might have gone wrong here as it was added _and_ removed in the same PR [guardian/Prebid.js@2328002](https://github.com/guardian/Prebid.js/commit/2328002d969308ef0f10ea8a94e9757711bba193)

So weird! So the `rtdModule` *was* never there??

PR #2144, opened 2025-08-05
Refine Admiral tag launch conditions

cemms1:
## What does this change?

- Moves Admiral `shouldRun` logic to its own file
- Includes additional criteria for when it can run:
  - Prevent Admiral running if `shouldHideAdverts`, `shouldHideReaderRevenue` or `isSensitive` are true for the page
  - Prevent Admiral running on `paid-content` sponsorship types (includes Guardian Labs and Hosted Content) 
  - Prevent Admiral running for certain sections ('about', 'info', 'membership', 'help', 'guardian-live-australia', 'gnm-archive')

## Why?

As part of the work ahead of launching the Admiral AB test, we want to more tightly control when the script can run and avoid the ad block recovery screen being shown in certain scenarios

### DIFF
--- a/bundle/src/lib/third-party-tags/admiral-adblocker.ts
+++ b/bundle/src/lib/third-party-tags/admiral-adblocker.ts
@@ -20,6 +20,7 @@ const baseAjaxUrl =
  * - Should not run for certain sections
  */
 const shouldRun =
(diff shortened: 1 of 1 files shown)


GHaberis (approved):
Looks good to me 👍

PR #2152, opened 2025-08-08
Remove prebid bid cache code

deedeeh:
## What does this change?

This PR removes Prebid bid cahe feature and all the code related to it such as the prebid grouped slots. Related switches removal work https://github.com/guardian/frontend/pull/28151

## Why?

There is no uplift in revenue after multiple tests to reach the highest potential of bid caching but it seems it doesn't work efficiently unless we integrate the multibid module with prebid bid cache but our GAM setup doesn't allow us to work with multibid.


bundle/src/lib/header-bidding/prebid/prebid.spec.ts
```diff
@@ -189,7 +181,6 @@ describe('initialise', () => {
 		jest.mocked(shouldIncludeBidder).mockReturnValue(
 			jest.fn().mockReturnValue(true),
 		);
-		jest.mocked(isUserInVariant).mockReturnValue(false);
```
deedeeh, line 192: I removed this as it seems like `prebid-id5` isn't in test anymore and I was using the `isUserInVariant` for grouped slots but it isn't needed anymore so I removed it from the imports.

PR #2156, opened 2025-08-11
Put the prebid946 to 5% of the global audience

GHaberis:
Increase the audience from 0 to 5%.


GHaberis:
I've updated the start and end dates, it wasn't clear whether an offset was necessary tbh, however there are 3 client-side AB tests running (some of which have been running for years) that cover the entire audience so an audience share with another test is unavoidable.

PR #2169, opened 2025-08-20
Revert "Add `tagid` and new placements to TripleLift bidder"

deedeeh:
Reverts guardian/commercial#2168


deedeeh:
It seems like the issue wasn't due to this PR being merged so I will close it. It was due to this change https://github.com/guardian/dotcom-rendering/pull/14412/files that's why we have all the Playwright tests failing.

PR #2176, opened 2025-08-22
Playwright tests use DCR POST endpoints

arelra:
## What does this change?

WebX want to remove the GET endpoints from DCR's PROD server:
https://github.com/guardian/dotcom-rendering/pull/14383

This has been done for DCR here:
https://github.com/guardian/dotcom-rendering/pull/14421

This PR switches the Commercial Playwright tests from using DCR's PROD server GET endpoints to the POST endpoints.

This primarily involves reworking the `load-page.ts` utility to switch the GET request to a POST request, fetching the JSON payload from PROD Frontend itself to POST to DCR rather relying on DCR's dev server to do this.

This change also does some cleaning up and improvements to the e2e tests:

- allow local fixtures to be passed to `load-page`
- allow overriding of config and switches to `load-page`
- remove configurable 'STAGE' to run the e2e tests against. I don't think we ever used this to point tests at CODE or PROD and removing makes the test setup simpler.
- remove the fixtures server. We used this for a couple of tests but we can now override the POST request to DCR with a fixture. This simplifies the dev server and test setup.
- add a `locators.ts` helper to centralise visibility checks and waiting behaviour
- improve the stability of `liveblog-ad-limit.spec.ts` as it was flakey when running locally and on CI
- add a CI step to skip mandb updates to mitigate occasional slow updates to mandb when installing packages

## Why?

Reduce DCR's PROD server surface area.

Tidy up.

## Testing

I've tested this against a DCR container image with the GET endpoints removed.


bundle/playwright/fixtures/pages/articles.ts
```diff
@@ -11,22 +8,13 @@ const stage = getStage();
  **/
 const articles = [
 	{
-		path: getTestUrl({
-			stage,
-			path: '/politics/2022/feb/10/keir-starmer-says-stop-the-war-coalition-gives-help-to-authoritarians-like-putin',
-		}),
```
arelra, line 17: Remove the concept of 'stage' i.e. dev, code or prod. Hence the `getTestUrl` is no longer required

i-hardy (approved):
This looks great, the changes to `loadPage` are a big boost to code clarity as well as adding functionality.

Jakeii:
Love the improvements you did here! Is any of the logic worth libing as it's shared with DCR's playwright tests?

Although we want to re-raise the idea of moving into DCR at some point... So perhaps redundant!

PR #2195, opened 2025-09-04
Create copies of modules for prebid-v9.46.0

GHaberis:
## What does this change?

This PR creates a v9.46.0 copy of every module that imports from [prebid.js/src](http://prebid.js/src), [prebid.js/libraries](http://prebid.js/libraries), [prebid.js/adapters](http://prebid.js/adapters). This creates copies of the following modules:

- /commercial/bundle/src/lib/header-bidding/prebid/modules/appnexusBidAdapter.ts
- /commercial/bundle/src/lib/header-bidding/prebid/modules/openxBidAdapter.ts
- /commercial/bundle/src/lib/header-bidding/prebid/modules/analyticsAdapter.ts

So we now have 

- /commercial/bundle/src/lib/header-bidding/prebid/modules/appnexusBidAdapter-v9.46.0.ts
- /commercial/bundle/src/lib/header-bidding/prebid/modules/openxBidAdapter-v9.46.0.ts
- /commercial/bundle/src/lib/header-bidding/prebid/modules/analyticsAdapter-v9.46.0.ts

In each copy any `prebid.js/*` import has been updated to use `prebid-v9.46.0.js/*`, this means the `'prebid-v9.46.0.js/*` aliases configured in our [webpack config](https://github.com/guardian/commercial/blob/main/bundle/webpack.config.mjs#L73) will work as intended and point to the v9.46.0 of the prebid library.

The module /commercialbundle/src/lib/header-bidding/prebid/pbjs-v9.46.0.ts has been updated to now imports these copies.

## Why?

The `prebid-v9.46.0.js/*` aliases in https://github.com/guardian/commercial/blob/main/bundle/webpack.config.mjs  do not pull in the code from node_modules/prebid-v9.46.0.js as intended. Even when we’re in the variant we’re still using the code from node_modules/prebid.js, this meant certain code such as the analyticsAdapter was not working in the variant. More info in [Trello](https://trello.com/c/yZUJWGcY/3562-investigate-why-prebid946-variant-page-views-arent-in-prebidanalytics-table).

### DIFF
--- a/bundle/src/lib/header-bidding/prebid/modules/analyticsAdapter-v9.46.0.ts
+++ b/bundle/src/lib/header-bidding/prebid/modules/analyticsAdapter-v9.46.0.ts
@@ -1,0 +1,308 @@
+// see http://docs.prebid.org/dev-docs/integrate-with-the-prebid-analytics-api.html
+import { log } from '@guardian/libs';
+import adapter from 'prebid-v9.46.0.js/libraries/analyticsAdapter/AnalyticsAdapter.js';
+import adapterManager from 'prebid-v9.46.0.js/src/adapterManager.js';
+import { fetch } from 'prebid-v9.46.0.js/src/ajax.js';
+import { EVENTS } from 'prebid-v9.46.0.js/src/constants.js';
+import { reportError } from '../../../../lib/error/report-error';
+import {
+	type BidArgs,
+	type EventData,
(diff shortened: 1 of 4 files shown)


GHaberis:
I've tested on `CODE` and it looks like it's fixed the issue 🤞

PR #2205, opened 2025-09-08
Reword step which handles reporting of failure for starting the DCR container

cemms1:
## What does this change?

Rewords the name of the step which handles the failure scenario in the previous step, "Start DCR in a container"

## Why?

It looked a bit like an actual failure due to the wording of the step, rather than a job that was skipped due to a lack of failure

Screenshot 2025-09-08 at 11 53 12


deedeeh (approved):
Thank you for doing this @cemms1 ✨

PR #2221, opened 2025-09-12
verifying fetch email from identity lib

dskamiotis:
## What does this change?

As part of our tech review doc for a 3rd Party Integration for Commercial Dev (ID5), we are investigating the use of the 3rd party JavaScript SDK. 
A piece of info we're considering sending the 3rd party via their SDK is a SHA-256 hash of a user's email address, 
We are checking if we can access this via the identity lib in the commercial bundle.

**Email Normalisation**: Per ID5 documentation requirements, emails are normalized before hashing (lowercase, trimmed, Gmail-specific rules) to ensure consistent hash generation across different systems and user input variations.
[ID5 Normalisation docs](https://wiki.id5.io/docs/passing-partner-data-to-id5#normalizing-hashed-inputs)

We'll only do this for signed-in users where we have access to the email address and consent has been granted

## Why?
This is somewhat of a spike to test whether we are able to retrieve a user's email address in the client when they're signed in.
In order to properly test this, it will need to be deployed to CODE.

**Screenshot**

| Before | After |
|----------|----------|
| Screenshot 2025-09-15 at 16 40 50   | Screenshot 2025-09-15 at 16 42 07  |


ashishpuliyel (changes requested):
Have I understood correctly that the hashing happens client-side? I would like for us to do the hashing server-side if at all possible.

GHaberis:
@dskamiotis do we need this PR anymore? I believe it's been superseded by https://github.com/guardian/commercial/pull/2282 (open) and https://github.com/guardian/commercial/pull/2275 (merged)?

PR #2225, opened 2025-09-18
add inventoryGroupId to criteo bid config

dskamiotis:


bundle/src/lib/header-bidding/prebid-types.ts
```diff
@@ -79,10 +79,11 @@ type PrebidOpenXParams = {
 
 type PrebidCriteoParams =
 	| {
-			networkId: number;
+			zoneId?: number;
```
GHaberis, line 82: Not sure this should be optional, an object of type `PrebidCriteoParams` could be either an object with `zoneId` or an object with `networkId` and an optional `pubId`. This type with `zoneId?` implies it could be an empty object? Might be worth testing to verify that.

PR #2250, opened 2025-10-14
Add new AB test framework tests to page targeting

Jakeii:
## Why?
Add to page targeting so that new ab tests appear in GAM data in the datalake, for example impressions.


cemms1 (commented):
Not a biggie but I think we used the term "beta" for the new AB testing framework to distinguish from the old without using the relative term "new"
See https://github.com/search?q=org%3Aguardian+betaab&type=code

cemms1 (approved):
Meant to click approve when I commented!

PR #2255, opened 2025-10-16
Feat: Timed Queue and SignIn Gate Intercept

on-ye:
## What does this change?

- Listens for the signin gate to be displayed, ands stops ad refreshes from happening
- As the event loop can't be paused in the browser, all refresh events are passed through a queue which can be paused and drained
- Once the signin gate modal is opened (listened to via an event listener: https://github.com/guardian/dotcom-rendering/pull/14605/files#diff-cc933475d587696c53144cb65852989ce22bf2b1fa13b10bdee3268f8966963bR308-R318), any subsequent refreshes are placed in a queue which get drained once the signin gate is closed

Note:
The TimedQueue doesn't handle interrupts at the moment. Assuming a queue is draining with a job scheduled to execute in 20 seconds for example, the job should be cancelled if the user triggers the queue to be paused. This behaviour can be implemented at a later date, but for the current functionality won't be needed as the signin gate will only appear once per session, thus a pause and drain will only happen once per session

## Why?

According to Google Ads, ads must be on screen for at least 30seconds before being refreshed.

## How to test

- Add logs to the `loadAdvert` function in bundle/src/display/load-advert.ts (https://github.com/guardian/commercial/pull/2255/files#diff-0241a8a89cd894db2efe453d93765405f4e40336ad9e5872aff8d1c622d57bb5R34). These logs should be before at the top of the function to acknowledge the function being called, and within to queue to determine when it's being drained
- Visit an article: http://localhost:3030/Article/https://www.theguardian.com/cities/2019/sep/13/reclaimed-lakes-and-giant-airports-how-mexico-city-might-have-looked
- In the console you should see the logs where the `loadAdvert` is being called and the callback being run
- Call `document.dispatchEvent(new CustomEvent("modal:open", { detail: { modalType: "sign-in-gate-v2" } }))` to trigger the pause functionality
- Scroll up and down to have the `loadAdvert` being called by other ad slots, but you should not see the callback being run
- Call `document.dispatchEvent(new CustomEvent("modal:close", { detail: { modalType: "sign-in-gate-v2" } }))` to trigger the drain functionality
- You should see the logs of the `loadAdvert` callbacks being run, indicating that once the modal is closed, all queued loadAdverts will be run

Apply the above steps to the `refreshAdvert` function, but trigger refresh advert from `https://github.com/guardian/commercial/tree/main/bundle/src/init/consented/dfp-listeners.ts`, and comment out the refresh logic within the callback to prevent errors e.g.

```ts
// bundle/src/init/consented/dfp-listeners.ts
import type { Advert } from '@guardian/commercial-core/types';

const initDfpListeners = (): Promise<void> => {
	let i = 0;
	setInterval(() => {
		refreshAdvert({
			id: String(++i)
		} as Advert)
	}, 1000);
    // ... rest of function
};
```

```ts
// bundle/src/display/load-advert.ts
export const refreshAdvert = (advert: Advert): void => {
	// advert.size contains the effective size being displayed prior to refreshing
	adQueue.add(() => {
		console.log('Refreshing advert', advert.id);
	});
};
```

### DIFF
--- a/bundle/src/lib/timed-queue.ts
+++ b/bundle/src/lib/timed-queue.ts
@@ -43,7 +43,10 @@ export function TimedQueue<T extends GenericJob>() {
 			if (pipeline.length || timeWhenPaused) {
 				// if there are pending jobs or if the queue is paused,
 				// add to the queue to be run when drained
-				const pipelineJob = createPipelineJob(job, executeImmediatelyAfterPause);
+				const pipelineJob = createPipelineJob(
+					job,
+					executeImmediatelyAfterPause,
+				);
 				pipeline.push(pipelineJob);
 			} else {
 				// run immediately if not paused and no pending jobs


i-hardy (approved):
This looks great 👍

Jakeii:
>According to Google Ads, ads must be on screen for at least 30seconds before being refreshed.

👀 Should refreshing also be added to the queue if they're scrolled out of view?

Jakeii:
> > Cool solution! But does this warrant a queue? There's never going to be more than a handful of refreshes queued isn't there?
> > Couldn't you clear the any refresh timeouts when the modal is displayed, save the timestamp when it was cleared, as a attribute on the adslot class perhaps, then recreate it when the modal is hidden with duration `30s - ${however much time it got through initially}`.
> 
> I think I possibly could, but seems a bit messier to me. We're calling refresh on the timeouts and invoking it from other functions that take place in the background. Some of these events are "fire-and-forget" from what I can tell like the `passbackRefresh` function so if we don't capture it we won't get any refreshes. We also call the `loadAds` function which needs to be invoked once the dialog is closed, so I don't know how to store and execute all these separate events once the user is done outside of a queue. Happy to jump on a call to talk about it though.
> 

Yeah a call might be good, we do have the `AdSlot` class to store things related to an ad instance, like for example a promise for when running prebid for the slot, as it runs at different times on different devices.

> > > According to Google Ads, ads must be on screen for at least 30seconds before being refreshed.
> > 
> > 
> > 👀 Should refreshing also be added to the queue if they're scrolled out of view?
> 
> Good question, it may be something I need to look at but at the moment the slot will refresh because bids would have been made already once the modal is closed, and also if the slot is ready - I'm not sure if it counts being ready when out of view???

I was thinking unrelated to the modal! For if an ad loads, then is scrolled out of view for 30s then scrolled back into view it'll refresh without really being *in view* for 30s. But no ones complained about this behaviour thusfar...

PR #2265, opened 2025-10-21
Delete carrot slot code

Jakeii:
## What does this change?
Delete any code related to the `carrot` slot

Breaking changeset as this is removing something from the exported ad sizes.

## Why?

The slot has been a long standing pain point, and we finally have permission to kill it!


on-ye (approved):
Basically CMD+F "carrot" right? 😄

PR #2269, opened 2025-10-22
adds an email hash function in core commercial

dskamiotis:
## What does this change?
A hash email function added which takes a string and uses the node `crypto` library.
The function is located in the core package to keep the logic centralised and prevent duplication or accidental changes.

This centralisation is important for future use, such as the upcoming `ID5 Data Deletion Request Lambda with Baton Integration`.
Consistent hashing ensures deletion requests can accurately identify users.

As per technical review and ID5 documentation, we are not normalising emails before hashing.

## Why?
As part of our ID5 integration, for users who give consent at the CMP level, we aim to increase the value of our ad inventory.

Hashing the email creates a strong, deterministic identifier for users, even if they switch browsers or devices. This improves match quality across SSPs and partners.

### DIFF
--- a/core/src/email-hash.spec.ts
+++ b/core/src/email-hash.spec.ts
@@ -1,39 +1,27 @@
-import { hashEmailForId5 } from './email-hash';
+import { normalisedEmail } from './email-hash';
 
 describe('test hashing of email', () => {
 	it('should return correctly hashed email', () => {
 		const mockEmail = 'guardianUser1@gmail.com';
-		const hashedEmail = hashEmailForId5(mockEmail);
-		expect(hashedEmail).toEqual(
-			'2b848438a88e73be0ee84c721ce1347dc891f2b7ff827d08406e462b45b6ce9e',
-		);
+		const hashedEmail = normalisedEmail(mockEmail);
+		expect(hashedEmail).toEqual('guardianuser1@gmail.com');
 	});
-	it('should return correctly hashed email with "." (ASCII code 46)', () => {
-		const mockEmail = 'guardian.user@gmail.com';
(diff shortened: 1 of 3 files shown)


core/src/email-hash.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+			'64c75391118cd72d85992cb7ef14d4a395e592f309e02c83d6186d2c8447fab7',
+		);
+	});
+	it('should hash emials in a case sensitive way', () => {
+		const mockEmail = 'GuardianUserLocal@gmail.com';
+		const hashedEmail = hashEmailForId5(mockEmail);
+		expect(hashedEmail).toEqual(
+			'1522956557177ce97b2663328a0991ee12b664446eb9c9b275758b0d5ab0a215',
+		);
+	});
```
dskamiotis: Im likely going to change these tests, with parameterised tests with `it.each`

core/src/email-hash.ts
```diff
@@ -0,0 +1,5 @@
+import SHA256 from 'crypto-js/sha256';
```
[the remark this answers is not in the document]
on-ye: Interesting, the crypto module is available on the browser and in node so not sure why that would cause a build problem

core/package.json
```diff
@@ -47,18 +47,20 @@
 	"dependencies": {
 		"@guardian/ab-core": "8.0.1",
 		"@guardian/libs": "26.0.0",
-		"@types/googletag": "~3.3.0"
+		"@types/googletag": "~3.3.0",
+		"crypto-js": "4.2.0"
```
on-ye: we can remove this package now

PR #2292, opened 2025-11-07
Feat: prebid 10

on-ye:
This PR introduces prebid v10.11.0 alongside our existing prebid v9.x, while providing a switch on initialisation to choose between which version to load at runtime.

**Note:**
A small caveat was found where the logging of a `bidwon` event isn't triggered, however the implementation details have been matched to v9 to see if in production this is a problem. If so, a separate PR will be created to rectify this apparent problem.

### DIFF
--- a/bundle/src/init/consented/prepare-prebid.ts
+++ b/bundle/src/init/consented/prepare-prebid.ts
@@ -2,6 +2,7 @@ import { isInCanada } from '@guardian/commercial-core/geo/geo-utils';
 import type { ConsentState } from '@guardian/libs';
 import { log, onConsent } from '@guardian/libs';
 import { once } from 'lodash-es';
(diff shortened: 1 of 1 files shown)


bundle/src/init/consented/prepare-prebid.ts
```diff
@@ -70,7 +82,10 @@ export const setupPrebidOnce: () => Promise<void> = once(setupPrebid);
  * https://docs.prebid.org/overview/intro.html
  * @returns Promise
  */
-export const init = (): Promise<void> => setupPrebidOnce();
+export const init = (shouldLoadNewVersion?: boolean): Promise<void> => {
+	loadNewVersion = !!shouldLoadNewVersion;
```
Jakeii: Is there any reason this couldn't be passed as a prop?

Jakeii (approved):

PR #2315, opened 2025-12-01
Integrate Trade Desk UID2 into prebid

dskamiotis:
## What does this change?

Adds UID2 support to Prebid, following the same pattern as ID5:  
- **US-only restriction**: Checks `consentState.framework === 'usnat'` to ensure UID2 is only added for users physically located in the USA (via IP-based geolocation)
- **Consent validation**: Uses `getConsentFor('theTradeDesk', consentState)` to verify user has granted consent (checks `doNotSell: false`)
- **Email hashing**: Refactored `hashEmail()` to accept a `hashType` parameter ('id5' | 'uid2'):
  - `'id5'` → hex encoding (existing behavior unchanged)
  - `'uid2'` → base64 encoding (required by UID2 spec)
- **Non-invasive integration**: Adds UID2 to the existing `userIds` array without affecting other ID providers
- **Staged rollout**: 90% of eligible US users get UID2, 10% split into control/variant for A/B testing

## Why?

Enable UID2 **for US users only** while:
- ✅ **Respecting regional privacy laws**: Only integrates for users in USA with explicit consent
- ✅ **Ensuring compliance**: Framework detection uses server-side IP geolocation (cannot be spoofed by edition dropdown)
- ✅ **Maintaining code quality**: Refactored hash function is reusable and type-safe
- ✅ **Allowing controlled rollout**: A/B test structure enables measurement of UID2 impact

Why both checks are needed:

framework === 'usnat' → User is in USA (IP-detected by SourcePoint CMP)
getConsentFor() → User has NOT opted out (doNotSell: false)

Verification:

SourcePoint header x-sp-geo-usp: US-[STATE] confirms IP-based geolocation
Non-US users (EU/AU) are blocked by framework check, regardless of consent status
Framework cannot be changed via edition dropdown (server-side only)

### Testing
✅ Tested with VPN across US/EU/AU regions
✅ Verified framework detection via network headers
✅ Confirmed consent validation for opt-in/opt-out scenarios
✅ All existing tests pass (ID5 integration unchanged)

### Screenshots
https://m.code.dev-theguardian.com/
| USA consent (test grp :control) | non USA consent test grp :variant)|
|--------------| ------------------|
 |Screenshot 2025-12-03 at 14 38 40 | Screenshot 2025-12-03 at 15 00 58  | 
|Screenshot 2025-12-03 at 15 06 06 |  Screenshot 2025-12-03 at 15 03 03|

### DIFF
--- a/.changeset/tough-dryers-send.md
+++ b/.changeset/tough-dryers-send.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial-core': major
+---
+
+BREAKING: Rename hashEmail to hashEmailForClient and change signature
--- a/core/src/email-hash.spec.ts
+++ b/core/src/email-hash.spec.ts
@@ -1,6 +1,6 @@
 import { hashEmailForClient } from './email-hash';
 
-describe('hashEmail: Id5 client', () => {
+describe('hashEmail: ID5 client', () => {
 	it('normalises and hashes an uppercase email', async () => {
 		const hashedEmail = await hashEmailForClient(
 			'testGuardianUser@gmail.com',
@@ -36,7 +36,7 @@ describe('hashEmail: Id5 client', () => {
 	});
 });
 
-describe('hashEmail: Uid2 client', () => {
+describe('hashEmail: UID2 client', () => {
 	it('normalises and hashes an uppercase email', async () => {
 		const hashedEmail = await hashEmailForClient(
 			'user@example.com',


bundle/src/lib/header-bidding/prebid/prebid.ts
```diff
@@ -410,6 +416,35 @@ const initialise = async (
 		}
 	}
 
+	if (getConsentFor('uid2', consentState)) {
```
dskamiotis: I need to update in the `libs` repo is require to add the `uid2` vendor name.

node_modules/.pnpm/@guardian+libs@26.0.1_@guardian+ophan-tracker-js@2.6.1_tslib@2.8.1_typescript@5.9.3/node_modules/@guardian/libs/dist/consent-management-platform/vendors.d.ts

GHaberis: @dskamiotis as discussed you shouldn't need to add the vendor to the list, as:

1) It should already be present, I can see the "The UK Trade Desk Ltd" is in the current vendor list.
2) The Identity & Trust team are responsible for managing the vendor list, so if it wasn't present we would need to work with them to get it added.

GHaberis: We only want to integrate `uid2` in the USA, to check US users consent state you can read `consentState.usnat.doNotSell` - if this is `true` we should **not** integrate. 

I'm not sure what `consentState.usnat.doNotSell` resolves to for users not in the US, I'd recommened testing this. 

We don't want to integrate _unless_ the user is in the US and `consentState.usnat.doNotSell` is `false`.

bundle/src/lib/header-bidding/prebid/prebid.ts
```diff
@@ -392,9 +393,11 @@ const initialise = async (
 			email &&
 			!isUserInTestGroup('commercial-user-module-ID5', 'variant')
 		) {
-			const hashedEmail = await hashEmail(email);
+			const hashedId5Email = await hashEmail(email, 'id5');
```
on-ye: note: this const is scoped within the if block so you don't have to rename it `hashId5Email` as it won't conflict with the `hashedUid2email` below if it was called `hashedEmail` too

PR #2316, opened 2025-12-01
fix: optionally chain googletag functions

on-ye:
## What does this change?

Optionally chain `googletag.setConfig` / `googletag.getConfig` and the slot definitions

## Why?

We've noticed a number of FE errors being reported when working with googletag. The `googletag.getConfig` and `googletag.setConfig` functions may not exist if googletag has been shimmed by an adblocker, so to avoid causing page errors we'll optionally chain these instances.

## Notes
- This is a minimally invasive PR. More extensive attempt had been made to optionally chain _all_ instances of `googletag.[method]`, which was not clean, so we'll start with what is explicitly being reported in the error logs
- There is some difficulty in understanding the downstream effects of not failing these features explicitly, such as how ad metrics are impacted by silent failures considering ad slot functionality continues running. These metrics will need to be observed for some time after these changes are merged.
- There is a possibility to wrap all `googletag` features into a `withGoogleTag((googletag) => ...)` type callback to ensure we don't execute these functions until it exists, however this poses its own challenges that I won't go into detail here

### DIFF
--- a/bundle/src/define/define-slot.ts
+++ b/bundle/src/define/define-slot.ts
@@ -121,7 +121,11 @@ const isEligibleForOutstream = (slotTarget: string) =>
 	(slotTarget === 'inline1' || slotTarget === 'top-above-nav');
 
 const allowSafeFrameToExpand = (slot: googletag.Slot) => {
-	slot.setConfig({
+	/*
+		eslint-disable-next-line @typescript-eslint/no-unnecessary-condition --
+		the slot.setConfig function may not exist if googletag has been shimmed by an adblocker
+	*/
+	slot.setConfig?.({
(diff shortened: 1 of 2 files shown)


Jakeii (approved):
Approving, but I think we should revisit at some point.

Looks like the [firefox shim](https://github.com/mozilla-firefox/firefox/blob/main/browser/extensions/webcompat/shims/google-publisher-tags.js) has not been updated with this new config syntax we started using in https://github.com/guardian/commercial/pull/2271   

We could try and detect the shim and not load googletag at all or try and detect old versions as the shim is on a very old `2021050601`, could even raise a PR to update the shim perhaps

PR #2327, opened 2025-12-03
fix: slot getConfig/setConfig for shimmed browsers

on-ye:
## What does this change?

Extends the optional chaining of `slot.getConfig` and `slot.setConfig` for ad slots, first started here: #2316 

## Why?

Reduce error reporting noise for adblock shimmed browsers, namely firefox

### DIFF
--- a/bundle/src/define/define-slot.ts
+++ b/bundle/src/define/define-slot.ts
@@ -202,24 +202,6 @@ const defineSlot = (
 
 	const slotReady = initSlotIas(id, slot, sizes);
 
-	void slotReady.then(() => {
-		EventTimer.get().mark('defineSlotEnd', slotTarget);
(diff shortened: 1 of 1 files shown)


bundle/src/define/define-slot.ts
```diff
@@ -209,13 +209,29 @@ const defineSlot = (
 		// wait until IAS has initialised before checking teads eligibility
 		const isTeadsEligible = isEligibleForTeads(id);
 
-		slot.setConfig({
+		/*
+			eslint-disable-next-line @typescript-eslint/no-unnecessary-condition --
+			the slot.setConfig function may not exist if googletag has been shimmed by an adblocker
+		*/
+		slot.setConfig?.({
```
i-hardy: Could we move this below the if statement to avoid the need for a conditional operator here?

PR #2341, opened 2025-12-10
feat: load prebid 9 in place of 10

on-ye:
## What does this change?

Using prebid v9 configuration over prebid v10

## Why?

Testing why the `bidWon` event is significantly higher in v9 than in v10, by eliminating difference in the test setup as a factor


GHaberis:
@on-ye so the plan is to deliver the same experience (Prebid v9) to the Control and Variant to see whether the `bidWon` event levels out?

PR #2344, opened 2025-12-11
Update Dependabot config for cooldowns and semver version update types

cemms1:
## What does this change?

- Allow any version update type for NPM packages rather than only allowing Dependabot to raise PRs for minor and patch versions.
- Specify cooldown periods for both NPM and github-actions ecosystems
- Remove `prebid` group since we never want to update Prebid via Dependabot - we almost always want to AB test this change

- _Drive-by fix: remove `.prettierrc` file. We already specify to use the `@guardian/prettier` config in the package.json file so we don't also need to have a `.prettierrc` file_

## Why?

Tidying up Dependabot config to stay on top of dependency updates without compromising security by upgrading too early

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,9 +19,6 @@ updates:
       guardian:
         patterns:
           - '@guardian/*'
-      prebid:
-        patterns:
-          - 'prebid.js'
       types:
         patterns:
           - '@types/*'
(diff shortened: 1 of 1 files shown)


Jakeii (approved):
Can't believe we never thought to remove the prebid entry before!

cemms1:
> Can't believe we never thought to remove the prebid entry before!

Me neither 🤦

PR #2352, opened 2025-12-22
refactor: update prebid setup files

on-ye:
## What does this change?

Before engaging in the integration of LiveRamp RampID, I'd like to refactor, consolidate and separate functionality for various configuration options of the prebid setup.

## Why?

Easier separation of concerns for prebid setup, ensuring the future integration of LiveRamp and other modules is isolated suitably

### DIFF
--- a/bundle/src/lib/header-bidding/prebid/bidders/criteo.ts
+++ b/bundle/src/lib/header-bidding/prebid/bidders/criteo.ts
@@ -1,22 +1,15 @@
 import { overridePriceBucket } from '../price-config';
+import type { BidderSettings } from '../types';
 
-export const configureCriteoBidderSettings = (): void => {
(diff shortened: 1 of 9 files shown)


GHaberis (approved):
Ok I’ve been through this and it looks good. I can see you’ve broken down the large `header-bidding/prebid/prebid.ts` file into smaller modules, and replaced it with `header-bidding/prebid/index.ts` which imports these smaller modules. You’ve split out individual bidder and ID handler config into distinct modules and grouped under respective “bidder” and “id-handlers” directories. I’ve cross checked the code and it looks like a lift and shift of logic into the smaller modules. 

I think it’s good, it’s easier to find the appropriate code for bidders and id-providers now, rather than searching the long file.

PR #2376, opened 2026-01-20
Proposals for Advert Event listener APIs

Jakeii:
## What does this change?
Copy the architecture guide from dotcom-rendering, and introduce proposals for 3 features

1. More comprehensive advert events API
2. a "queue" similar to googletag's `googletag.cmd` for running code dependant on commercial, outside of commerical
3. Export from `commercial-core` a way to use the API from other code on the page

## Why?
See individual proposals, but in summary this would allow external code to hook into our advert lifecycle events. We can also make our own lifecycle dependant code much more consistent and hopefully easier to understand.

### DIFF
--- a/docs/architecture/proposals/003-global-advert-event-listeners.md
+++ b/docs/architecture/proposals/003-global-advert-event-listeners.md
@@ -12,12 +12,11 @@ Export a function that allows other code to register event listeners for Advert 
 
 ```ts
 import { onAdvertEvent } from '@guardian/commercial';
-onAdvertEvent('dfp--top-above-nav', 'loaded', (advert) => {
+onAdvertEvent('loaded', (advert) => {
 	console.log(`Advert ${advert.id} has loaded`);
 });
 
 onAdvertEvent(
-	'dfp--top-above-nav',
 	'rendered',
 	(advert) => {
 		console.log(`Advert ${advert.id} has rendered`);
@@ -25,11 +24,36 @@ onAdvertEvent(
 	{ once: true },
 );
 // onceAdvertEvent sounds weird, so using options instead
+
+// Example of listening to multiple events
+let startTime;
+onAdEvent(['ready', 'rendered'], (advert) => {
+	if (!advert.id !== 'dfp--top-above-nav') {
+		return;
+	}
+
+	switch (advert.state) {
+		case 'ready':
+			startTime = new Date().now();
+			break;
+		case 'rendered':
+			console.log('time to render:', new Date().now() - startTime);
+			break;
+	}
+});
+
+// Example of listening to all events
+onAdEvent((advert) => {
+	if (!advert.id !== 'dfp--top-above-nav') {
+		return;
+	}
+
+	console.log(`Advert ${advert.id} is:${advert.state}`);
+});
 ```
 
-The `onAdvertEvent` function would take three parameters:
+The `onAdvertEvent` function would take 2-3 parameters:
 
-- `advertId`: The ID of the advert to listen for events on.
 - `event`: The lifecycle event to listen for (e.g., 'ready', 'preparing', 'fetching', 'loading', 'rendered').
 - `callback`: A function to be called when the event occurs, receiving the Advert instance as a parameter.
   The registered callback will be invoked whenever the specified event occurs for the given Advert instance.


docs/architecture/proposals/001-advert-events.md
```diff
@@ -0,0 +1,62 @@
...
+});
+```
+
+The `on` method could be called multiple times for the same advert and event, if refreshes occur. And takes two parameters:
+
+- `event`: The lifecycle event to listen for.
+- `callback`: A function to be called when the event occurs, receiving the Advert instance as a parameter.
+  The registered callback will be invoked whenever the specified event occurs for the Advert instance.
+
+A `once` method could also be added to register a one-time listener that is removed after being invoked once.
```
on-ye: I agree with the premise of being able to listen to ad events, however this class is only used by one function (`createAdvert`) which then gets loaded in the `dfpEnv.advertsToLoad` function, therefore these events won't be listened to?

The only way that I could see this working is if we invert the control flow of our code e.g. `onSlotRender`, `onSlotLoad` and contain them within the `Advert` class, then registering them someway with the `dfpEnv.adverts` object to bridge the gap between google ads control logic and other ad associated logic (Though that feels like the wrong way round?)

docs/architecture/proposals/003-global-advert-event-listeners.md
```diff
@@ -0,0 +1,43 @@
...
+onAdvertEvent(
+	'dfp--top-above-nav',
+	'rendered',
+	(advert) => {
+		console.log(`Advert ${advert.id} has rendered`);
+	},
+	{ once: true },
+);
+// onceAdvertEvent sounds weird, so using options instead
+```
```
on-ye: This is great. I'd change the API slightly like so:

```tsx
import { onAdEvent } from '@guardian/commercial';

onAdEvent('loaded', (advert) => {
	console.log(`Advert ${advert.id} has loaded`);
}, params);
```

- Allow listeners to perform actions on _any_ ad if desired
- Performance Optimisation: The params could be used to filter on ads at the point of registration if desired, rather than filtering based on the advertId during the callback execution
- Minor: maybe `onAdEvent` is a more succinct name

```tsx
onAdEvent('loaded', (advert) => {
	if (advert.id === "dfp--top-above-nav") {
		console.log(`Advert ${advert.id} has loaded`);
	}
}, {
  once: true
});
```

- Maybe we could enable listening to multiple events for things like:
```tsx
let startTime;
onAdEvent(['ready', 'rendered'], (advert) => {
	if (!advert.id !== "dfp--top-above-nav")  {
		return;
	}

	switch (advert.state) {
		case "ready":
			startTime = new Date().now();
			break;
		case "rendered":
			console.log('time to render:', new Date().now() - startTime);
			break; 
	}
});
```

- Or all events:
```tsx
onAdEvent((advert) => {
	if (!advert.id !== "dfp--top-above-nav")  {
		return;
	}

	console.log(`Advert ${advert.id} is:` advert.state`);
});
```

These options would make the event system flexible, reduce imperative code flow and make unit testing easier against a single function rather than multiple function instances.

docs/architecture/proposals/001-advert-events.md
```diff
@@ -0,0 +1,62 @@
...
+But in [`on-slot-render.ts`](../../../bundle/src/init/consented/dfp-listeners.ts) we listen for events from googletag.
+
+And in [`lazy-load.ts`](../../../bundle/src/display/lazy-load.ts) we check the `isRendered` boolean to determine if the advert has already been rendered, so we can refresh instead of load.
+
+**Note:**
+I know we have some potential plans to refactor some of this code in the future, I still think having a clear system for advert lifecycle events would be beneficial particularly for those we don't control like when an advert is scrolled into view or when an advert is fetched from GAM.
+
+## Proposal
+
+Add an event system to the Advert class that allows subscribing to lifecycle events. And add a `status` attribute to track the current state of the Advert.
```
dskamiotis: This feels like a really positive step forward. Having an explicit and shared notion of advert state would make the lifecycle much easier to reason about.

At the moment a lot of this state is effectively held in developers’ heads across different files and mechanisms, which is hard to scale as the codebase grows. Making this explicit via lifecycle events and status feels like a strong improvement in clarity and maintainability.

docs/architecture/proposals/003-global-advert-event-listeners.md
```diff
@@ -0,0 +1,43 @@
...
+
+- `advertId`: The ID of the advert to listen for events on.
+- `event`: The lifecycle event to listen for (e.g., 'ready', 'preparing', 'fetching', 'loading', 'rendered').
+- `callback`: A function to be called when the event occurs, receiving the Advert instance as a parameter.
+  The registered callback will be invoked whenever the specified event occurs for the given Advert instance.
+- `options` (optional): An object that can contain additional options, such as:
+    - `once`: A boolean indicating whether the listener should called on the first invocation only.
+
+Internally this would use the [Advert lifecycle events API proposal](01-advert-events.md) and [queue proposal](02-commercial-loaded-queue.md) proposals to ensure that listeners can be registered even before commercial has initialised.
+
```
dskamiotis: This feels like a natural extension of the lifecycle events proposal and a nice way to make advert state observable outside the commercial bundle.

docs/architecture/proposals/002-commercial-loaded-queue.md
```diff
@@ -0,0 +1,31 @@
...
+
+If a function is pushed after commercial has already initialised, it should be executed as soon as possible.
+
+This is an established pattern used by other libraries, for example Googletag uses `window.googletag.cmd` for this purpose, opt out advertising also uses a similar pattern with `window.ootag.queue`.
+
+```ts
+// External code
+window.guardian.commercial.queue = window.guardian.commercial.queue || [];
+window.guardian.commercial.queue.push(() => {
+	// Code that depends on commercial being initialised
```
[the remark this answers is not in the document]
GHaberis: We mention above this would be useful to "run code that depends on adverts being defined" . Would we therefore need arguments to indicate that the code on the queue is dependent on something - eg. a particular advert or a particular event? If so, it feels like there's overlap with the library changes proposed in `003-global-advert-event-listeners.md` - which is a nice proposal IMO!

If however it's simply to run some code when commercial has initialised then maybe this is a simple enough approach.

PR #2413, opened 2026-02-16
Feat/prebid v10

on-ye:
## What does this change?

This attempts to reintegrate `prebid v10.23.0` alongside `prebid v9.x` via an A/B Test:
- The types between v9 and v10 are slightly different however can not easily live alongside each other due to the global `PrebidJS` so the types for prebid v10 have been used in most places, with notes added to remove prebid v9 types once the upgrade is proved successful
- A much cleaner and leaner integration of v10 module declarations
- A separation of upgrade docs between v9 and v10

## Why?

Upgrading a critical package of ours as well as ensuring we safely transition from one version to the other


dskamiotis (commented):
Noting most changes are:
- File moves and new module structure (modules-v10.11.0 → modules-v10.23.0)
- Typing updates (Event, GuAnalyticsAdapter, UserId)
- Added sendPayload.ts for analytics POSTing

A few questions for for my undertanding/clarity:
- Is sendPayload.ts entirely new behaviour, or refactor for types?
- Should analytics v10 always run, or should it be gated behind the AB test?
- Just confirming pbjs.processQueue() is safe to run immediately on import, even if multiple bundles exist on a page?

PR #2420, opened 2026-02-18
add EventTimer marks to measure timing of user ID retrieval

dskamiotis:
## What does this change?

This PR adds EventTimer marks to monitor the lifecycle of fetching User IDs, from the start of the process to the successful resolution of all User IDs.

Events added:
- getUserIdsStart
- getUserIdsEnd

## Why?
As part of the work to allow User IDs to be declared asynchronously in the commercial bundle, we want to monitor the current timing of the User ID fetching process. 
This will help us better understand performance and identify potential bottlenecks.

### DIFF
--- a/bundle/src/lib/header-bidding/prebid/id-handlers/index.ts
+++ b/bundle/src/lib/header-bidding/prebid/id-handlers/index.ts
@@ -1,3 +1,4 @@
+import { EventTimer } from '@guardian/commercial-core/event-timer';
 import { type ConsentState, getConsentFor } from '@guardian/libs';
 import { getEmail } from '../../../identity/api';
 import { isSwitchedOn } from '../../utils';
@@ -10,6 +11,8 @@ import { getUserIdForTradeDesk } from './tradedesk';
 export const getUserSyncSettings = async (
 	consentState: ConsentState,
 ): Promise<UserSync> => {
+	EventTimer.get().mark('getUserIdsStart');
+
 	const userEmail = await getEmail();
 	const fetchId5UserId =
 		getConsentFor('id5', consentState) && getUserIdForId5(userEmail);
@@ -37,6 +40,7 @@ export const getUserSyncSettings = async (
 			return Array.isArray(idModule) ? idModule : [idModule];
 		});
 	});
+	EventTimer.get().mark('getUserIdsEnd');
 
 	const userSync: UserSync = isSwitchedOn('prebidUserSync')
 		? {


GHaberis:
> I could be wrong, I think the improvement Onye carded up won't actually affect the timing of user id fetching itself, but of the overall time it takes prebid to run, by making it non-blocking, I don't think this is something we need to track?
> 
> Also adding a timer means you'll need to update the table schema in the datalake in most cases.

If I understand correctly we currently wait for `getUserSyncSettings` to resolve, which means we can include the user IDs in the `userSync` prop (with User IDs) of the `pbjsConfig`, which is used when calling `window.pbjs.setConfig`.

The proposed enhancement is to not wait for `getUserSyncSettings` to resolve when creating `pbjsConfig`, and instead to add the `userSync` prop to the prebid config via `window.pbjs.mergeConfig` once all the IDs are available. 

The benefit being we remove the delay taken to retrieve the user IDs when we initialise Prebid which means we initialise Prebid quicker? I imagine Onye was thinking about this following the introduction of Liveramp in particular as it has a dependency on loading a 3rd party script before the ID is available - unlike the other ID providers.

This is extra complexity, so it would be good to have an idea what the time saving is, I'm not sure there's an existing timer we can use as a measurement here? If there is an existing timer in place we could possibly measure whether we've seen any difference in the current Liveramp test.

I also think there's a trade-off if we're potentially initialising Prebid without the `userSync` prop (with User IDs) in the initial config. We probably want to understand what the impact of this may be - perhaps an AB test is the best means of seeing whether there's any impact.

dskamiotis:
> If I understand correctly we currently wait for `getUserSyncSettings` to resolve, which means we can include the user IDs in the `userSync` prop (with User IDs) of the `pbjsConfig`, which is used when calling `window.pbjs.setConfig`.
> 
> The proposed enhancement is to not wait for `getUserSyncSettings` to resolve when creating `pbjsConfig`, and instead to add the `userSync` prop to the prebid config via `window.pbjs.mergeConfig` once all the IDs are available.
> 
> The benefit being we remove the delay taken to retrieve the user IDs when we initialise Prebid which means we initialise Prebid quicker? I imagine Onye was thinking about this following the introduction of Liveramp in particular as it has a dependency on loading a 3rd party script before the ID is available - unlike the other ID providers.
> 
> This is extra complexity, so it would be good to have an idea what the time saving is, I'm not sure there's an existing timer we can use as a measurement here? If there is an existing timer in place we could possibly measure whether we've seen any difference in the current Liveramp test.
> 
> I also think there's a trade-off if we're potentially initialising Prebid without the `userSync` prop (with User IDs) in the initial config. We probably want to understand what the impact of this may be - perhaps an AB test is the best means of seeing whether there's any impact.

@GHaberis yes I think your understanding is correct
Current Behaviour: We currently wait for getUserSyncSettings to resolve before initializing Prebid, ensuring userSync (with userIds) is included in the initial `pbjsConfig`.

The benefit of loading and initialising Prebid quicker, would especially be apparent for the first ad load.

I think the trade-off is that using mergeConfig to add the user IDs to the config in a non-blocking fashion (via Promise.all) means the updates will likely be chained in a .then() block. This forces the resolution of these IDs to wait in the queue and not resolve until the main thread is complete. As a result, the first ad request may not include user IDs in the userSync data, since the mergeConfig update happens asynchronously after the user IDs are resolved.

To summarise (for myself and for clarity)
Proposed feature: Initialise Prebid immediately without waiting for getUserSyncSettings.
Dynamically update the userSync property later using window.pbjs.mergeConfig once all userIds are resolved.

Benefits: Faster Prebid initialisation, reducing delays caused by user ID resolution (e.g., LiveRamp, which depends on a third-party script).

Trade-Offs: The first ad request may not include userIds in the userSync data, potentially impacting targeting and CPMs.

PR #2432, opened 2026-02-25
Move header bidding and load functions to Advert class

Jakeii:
## What does this change?
Moves `requestBids`, `refreshBids`, `load`, `display` and `refresh` to the `Advert` class.

## Why?
The main reason I've done this *now*, is this will enable only advert instances themselves to update their own status (suggested by @on-ye in https://github.com/guardian/commercial/pull/2406#discussion_r2792623095)

These function all revolve around the `Advert` class anyway and are all pretty small.

I believe we're hoping to move towards not spreading our ad loading logic across so many modules, this helps with that.

### DIFF
--- a/bundle/src/display/display-ads.ts
+++ b/bundle/src/display/display-ads.ts
@@ -28,7 +28,6 @@ const displayAds = (): void => {
 		: undefined;
 
 	if (firstAdvertToLoad) {
-		console.info(`Loading first advert with id ${firstAdvertToLoad.id}`);
 		firstAdvertToLoad.load();
 		dfpEnv.advertsToLoad = [];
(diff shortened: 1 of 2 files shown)


bundle/src/display/display-ads.ts
```diff
@@ -29,7 +28,8 @@ const displayAds = (): void => {
 		: undefined;
 
 	if (firstAdvertToLoad) {
-		loadAdvert(firstAdvertToLoad);
+		console.info(`Loading first advert with id ${firstAdvertToLoad.id}`);
```
deedeeh: It's great to have more logging info ✨

PR #2433, opened 2026-02-25
Update detection method for applying offset class to right rail ads on interactives

Jakeii:
## What does this change?
Just check if the width is `620px`, this is a better indicator if we need to apply the class that applies the negative margins to right rail ads.

## Why?
The current detection method doesn't work in some circumstances where the contentGrid is larger than the body. e.g. this article https://www.theguardian.com/politics/ng-interactive/2026/feb/25/how-rightwing-rhetoric-has-risen-sharply-in-the-uk-parliament-an-exclusive-visual-analysis

### DIFF
--- a/bundle/src/insert/spacefinder/interactive.ts
+++ b/bundle/src/insert/spacefinder/interactive.ts
@@ -20,6 +20,8 @@ const getInteractiveGridWidths = () =>
 
 /**
  * Determine whether the grid body is the standard desktop width, if it is then right rail ads don't need the offset right class, if it's not then we need to add the offset right class to move the ad over to the right.
+ *
+ * The 620px comes from the `ArticleContainer` component in DCR https://github.com/guardian/dotcom-rendering/blob/a321a1e296f54c73580151f3b4c7170ce373eb1f/dotcom-rendering/src/components/ArticleContainer.tsx
  */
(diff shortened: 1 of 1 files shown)


bundle/src/insert/spacefinder/interactive.ts
```diff
@@ -12,22 +12,17 @@ const getInteractiveGridWidths = () =>
...
- * Determine whether the grid body is the full width of the content grid. If the grid body is full width, then we need to insert the ad without the offset right class, otherwise the ad will be pushed too far into the right hand column and could end up outside of the viewport.
+ * Determine whether the grid body is the standard desktop width, if it is then right rail ads don't need the offset right class, if it's not then we need to add the offset right class to move the ad over to the right.
  */
-const isBodyFullWidthOfContentGrid = (
-	bodyWidth: number,
-	contentGridWidth: number,
-): boolean => {
-	return bodyWidth >= contentGridWidth;
+const isBodyStandardDesktopWidth = (bodyWidth: number): boolean => {
+	return bodyWidth === 620;
```
cemms1: I wonder if we could reference the [ArticleContainer](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/components/ArticleContainer.tsx) component for where 620 is the "standard" article width

PR #2501, opened 2026-04-02
Refactor AdEvent Public API with document-based ad event

dskamiotis:
## What does this change?
Refactors the way consumers use the `globalAdEvents` public API:
- Replaces `globalAdEvents` (EventTarget singleton) with `document` as the event bus.
- Exports a typed listener function from `@guardian/commercial-core` that filters by `status` and optional `slotName`.
- Updates the `Advert` class to dispatch namespaced events (`commercial:adStatusChange`) on `document`.
- Updates tests and documentation to reflect the new API.
- Removes redundant unit tests from Advert spec file.

### Example Usage:
Old usage
```
globalAdEvents.addEventListener('adStatusChange', (event) => {
    console.log(event.detail);
});
```

New usage
```
globalAdEvents('rendered', (event) => {
    console.log(event.detail);
});
```

## Why?
The current `globalAdEvents` API uses a module-level EventTarget singleton (const globalAdEvents = new EventTarget()). 
When external consumers like DCR or interactives import @guardian/commercial-core, their bundler (e.g., webpack) creates separate instances of the singleton in each bundle. 
This means events dispatched in one bundle are not visible to listeners in another.

Switching to document as the event bus ensures all listeners, across all bundles, receive the same events. The new API also simplifies usage by introducing a typed listener function that filters events by status and slotName.

### Notes:
- This change is not backward-compatible and requires consumers to update their usage of globalAdEvents.
- All existing tests have been updated, and additional tests have been added to cover edge cases.
- Using document as the event bus ensures consistent event propagation across bundles with negligible performance overhead.

---
  - https://app.asana.com/0/0/1213910125679479


deedeeh (commented):
I haven't been very involved in the queuing but your refactor makes total sense as we need to have the same behaviour in different projects bundles. 

Will this change require comms to different teams using the queue? Do we consider it breaking changes for anyone using the queue from `commercial-core`? 

These changes require a changeset @dskamiotis

dskamiotis:
> Will this change require comms to different teams using the queue


Yes, this change will require communication with teams like the Interactives team and others who rely on the globalAdEvents API. They will need to ensure they import the @guardian/commercial-core library to access the updated globalAdEvents module.

This implementation no longer uses the queue directly. Instead, it attaches listeners and handlers to the document object, ensuring consistent behaviour across different project bundles.

deedeeh:
> > Will this change require comms to different teams using the queue
> 
> Yes, this change will require communication with teams like the Interactives team and others who rely on the globalAdEvents API. They will need to ensure they import the @guardian/commercial-core library to access the updated globalAdEvents module.
> 
> This implementation no longer uses the queue directly. Instead, it attaches listeners and handlers to the document object, ensuring consistent behaviour across different project bundles.

Yup understood. We just need to make sure they are aware before probably you merge this PR and merge the new release and once you have the latest `commercial/core` version you send it to them.

dskamiotis:
> > > Will this change require comms to different teams using the queue
> > 
> > 
> > Yes, this change will require communication with teams like the Interactives team and others who rely on the globalAdEvents API. They will need to ensure they import the @guardian/commercial-core library to access the updated globalAdEvents module.
> > This implementation no longer uses the queue directly. Instead, it attaches listeners and handlers to the document object, ensuring consistent behaviour across different project bundles.
> 
> Yup understood. We just need to make sure they are aware before probably you merge this PR and merge the new release and once you have the latest `commercial/core` version you send it to them.

For sure good shout, thanks Dina, Im planning on updating @frederickobrien and then informing the rest of the team and direct them to readme before merging and publishing. Its not a breaking change at the moment but will be useful hopefully now consumers

docs/global-ad-events.md
```diff
@@ -29,42 +29,39 @@ Fired every time any ad slot changes status. All listeners receive every event 
...
+});
+
+// Later, remove the listener
+subscription.remove();
+```
+
 ## Notes
 
-- The event name is `'adStatusChange'` — it must match exactly or the listener will silently receive nothing.
+- The event name is `'commercial:adStatusChange'` — it must match exactly, or the listener will silently receive nothing.
```
cemms1: I'm not sure this comment makes sense any more with the updates as we use the ad status types instead of typed event listeners?

core/src/global-ad-events.ts
```diff
@@ -1,9 +1,62 @@
...
+	const event = e as AdEventCustomEvent;
+	eventHistory.push(event);
+});
+
+function globalAdEvents(
+	status: AdvertStatus | AdvertStatus[],
+	listenerHandler: (event: AdEventCustomEvent) => void,
+	slotName?: string,
+) {
+	const newStatus = Array.isArray(status) ? status : [status];
```
cemms1:
```suggestion
	const parsedStatus = Array.isArray(status) ? status : [status];
```

It's normally a good idea to avoid the word `new` when naming things as it's quite subjective and doesn't describe what it actually is. What we're doing here is making sure that the status is an array so maybe `parsedStatus` or `statusAsArray` or `statusArr` could work?

core/src/global-ad-events.ts
```diff
@@ -1,9 +1,62 @@
...
+	const matches = (event: AdEventCustomEvent) => {
+		const statusMatch = newStatus.includes(event.detail.name);
+		const slotMatch = !slotName || event.detail.slotName === slotName;
+		return statusMatch && slotMatch;
+	};
```
cemms1: _Optional / nitpick_

This could be easier to test if we add `newStatus` and `slotName` to the argument list to convert it into a pure function.

core/src/global-ad-events.ts
```diff
@@ -1,9 +1,62 @@
...
+	const listener = (e: Event) => {
+		const event = e as AdEventCustomEvent;
```
cemms1: Instead of `as`, we could use a type guard to check that it _is_ in fact an `AdEventCustomEvent`

Here's where we've done something similar previously:
https://github.com/guardian/commercial/blob/731c05bc3778c3ac42480a492a310175281c1189/bundle/src/init/consented/opinary.ts#L41-L56

core/src/global-ad-events.spec.ts
```diff
@@ -0,0 +1,316 @@
+import type { AdEventCustomEvent } from './global-ad-events';
+import {
+	_resetHistory,
+	eventHistory,
+	globalAdEvents,
+} from './global-ad-events';
+
+describe('globalAdEvents', () => {
+	let subscription: { remove: () => void } | undefined;
```
cemms1: Is there a reason for using `let` at the top rather than re-defining the subscription within each test?

core/src/global-ad-events.spec.ts
```diff
@@ -0,0 +1,316 @@
...
+					slotName: 'inline1',
+					name: 'rendering',
+					status: true,
+				},
+			}),
+		);
+
+		expect(handler).not.toHaveBeenCalled();
+	});
+	it('handler not called when slotName does not matching', () => {
```
cemms1:
```suggestion
	it('handler not called when slotName does not match', () => {
```

cemms1 (approved):
Looking really good! Thanks for adding the tests and documentation ✨ 

Have left a couple of comments to have a look at but happy to approve given that none of them are blockers

docs/global-ad-events.md
dskamiotis: great spot yes! the api logic deals with this under the hood

core/src/global-ad-events.ts
dskamiotis: nice

core/src/global-ad-events.spec.ts
dskamiotis: I've refactored the tests so the subscription variable is now defined inside each test, and cleanup is handled directly where needed. This gets rid of the let at the top of the file. The previous approach was just to avoid repeating the variable declaration and cleanup, but this way keeps things more self-contained

core/src/global-ad-events.ts
dskamiotis: Nice, Ive refactored to be a pure function 👍

PR #2514, opened 2026-04-08
Extract commentAdverts from commercialFeatures

deedeeh:
## What does this change?

This PR extracts `commentAdverts` from `commercialFeatures` to the closest location where it's being used once. 

## Why?

Refactoring to simplify the codebase. We're extracting all the logic from commercial features as it's hard to understand the Singleton.


dskamiotis (approved):
This looks good to me Dina! 

However, I think you're having the same issue I had earlier with the test coverage, so perhaps amending the `coverageThreshold` in the `jest.config.ts`?

cemms1 (approved):

PR #2515, opened 2026-04-08
Move third party tags logic out of commercial features

cemms1:
## What does this change?

- Removes `thirdPartyTags` from `commercialFeatures`, moving it to the third-party-tags file instead
- Removes `isUserPrefsAdsOff` logic entirely. I can't find any reference to it in code or documentation so assume it is no longer relevant today
- Removes the argument from `isSecureContactPage` since the pageId is stable and will not change for the duration of commercial scripts

## Why?

Simplifying and tidying up the codebase, making it easier to understand for current and future developers.


GHaberis (approved):
Looks good 👍

PR #2516, opened 2026-04-09
Specify types in tsconfig files

cemms1:
## What does this change?

- Adds `types: ["node", "jest"]` to the tsconfig files in `bundle` and in `core`
- Adds additional types "googletag" and "webpack-env" to the `bundle` `tsconfig.json` file
- Removes old `tsconfig.test.json` from the root as this doesn't seem to be used

## Why?

My VSCode keeps highlighting global `jest` methods as errors when they aren't explicitly imported. These do not error when actually running the tests.

This change allows the IDE to correctly pick up the global Jest (plus Googletag and Webpack) types to avoid littering the editor with false TS errors

| Before | After |
| ----- | ----- |
| ![before][] | ![after][] |

[before]:https://github.com/user-attachments/assets/88016216-f376-435c-bdd6-baa82714abec
[after]:https://github.com/user-attachments/assets/095b4672-81da-4324-9883-58c3e4ae9bc9


GHaberis (approved):
Looks good 👍

PR #2518, opened 2026-04-09
Remove `youtubeAdvertising` from commercial features

cemms1:
## What does this change?

Removes the `youtubeAdvertising` property from `CommercialFeatures`

## Why?

Tidying up and improving the readability of commercial code.

Turns out `commercialFeatures.youtubeAdvertising` isn't actually used anywhere 🤔 🔍
It seems to have been [inherited from frontend](https://github.com/guardian/frontend/blob/1d965e8c9b92a2d69374d94c69e16d0de253b32b/static/src/javascripts/projects/common/modules/commercial/commercial-features.ts#L25) but isn't actually used in `commercial`


GHaberis (approved):
Looks good 👍

PR #2519, opened 2026-04-09
Pin dependency versions

cemms1:
## What does this change?

Pins dependency versions in the `package.json` files

## Why?

As per the Guardian [NPM recommendations](https://github.com/guardian/recommendations/blob/6e2c41db1d942611d35b04407203eca06795f50c/dependencies.md?plain=1#L29-L54)


GHaberis (approved):
Looks good 👍

PR #2523, opened 2026-04-13
Move `articleBodyAdverts`  out of `commercialFeatures`

cemms1:
## What does this change?

- Takes the `articleBodyAdverts` code out of `commercialFeatures`, moving relevant tests over to the new location and updating any usages of this property in other places of the code base

## Why?

As part of some refactoring work to simplify the commercial bundle code, we're moving features away from `commercialFeatures` one by one, until we can remove it entirely.


dskamiotis (approved):
Great FWICS you have removed from commercialFeatures and created a new module to deal specifically with `article-body-adverts` including testing and updated imports 🎆

PR #2527, opened 2026-04-16
remove comscore from commercial features 

dskamiotis:
## What does this change?
This PR moves the `comscore` logic from commercialFeatures to the comscore module.

**Note:** removal of `commercial-features.spec.ts`
The only remaining test was for comscore, so the spec file has been deleted 
rather than left empty. `commercial-features.ts` itself still exists as adFree 
will be removed in a follow-up PR.

## Why?
As part of some refactoring work to simplify the commercial bundle code, we're 
moving features away from commercialFeatures one by one. A separate PR 
(ds/remove-adFree-from-commercialFeatures) will complete the removal.


cemms1:
Let's remove `commercialFeatures` entirely in a separate PR. I've just spotted that there's some usage of `adFree` in other places still! I should have removed those in the previous PR.

cemms1:
Can we still remove `this.comscore` from commercialFeatures in this PR?

dskamiotis:
> Let's remove `commercialFeatures` entirely in a separate PR. I've just spotted that there's some usage of `adFree` in other places still! I should have removed those in the previous PR.

Yep have removed that commit, this now just deals with comscore.
I have a separate PR ready to go once this one is merged which removes adFree completely and also the commercialFeatures 👍

cemms1:
> In addition the commercialFeatures is now no longer needed and has also been removed.

@dskamiotis The PR description needs tweaking a little still

cemms1 (approved):

PR #2529, opened 2026-04-16
Remove adFree from commercialFeatures and delete commercial-features.ts

dskamiotis:
## What does this change?
Removes `adFree` from commercialFeatures, inlining isAdFree() directly in the modules that need it. Since adFree was the last remaining property, commercial-features.ts is removed entirely. Tests are updated accordingly.
- Removes `commercial-features.ts` module.
- Modules now import `isAdFree()` directly from `lib/ad-free` rather than going through the commercialFeatures wrapper.
- Adds tests for `ad-free-slot-remove.ts`.

## Why?
As part of ongoing refactoring to simplify the commercial bundle, we're moving features out of commercialFeatures one by one. With adFree being the last property, we can now remove the module entirely.

---
  - https://app.asana.com/0/0/1214101224337661


bundle/src/init/consented/ad-free-slot-remove.ts
```diff
@@ -1,17 +1,24 @@
...
+const removeAdFreeSlots = () => {
+	if (!isAdFree()) {
 		return Promise.resolve();
 	}
 
 	return removeSlots();
-});
+};
+
+const adFreeSlotRemove = once(removeAdFreeSlots);
```
cemms1, line 18: Does it not work to test the exported function `adFreeSlotRemove`?
Wondering why we need to call the `once` function in a separate function

bundle/src/init/consented/ad-free-slot-remove.spec.ts
cemms1, line 1: 👏

bundle/src/init/consented/prepare-prebid.spec.ts
```diff
@@ -223,7 +218,7 @@ describe('init', () => {
 		window.guardian.config.switches = {
 			prebidHeaderBidding: true,
 		};
-		commercialFeatures.adFree = true;
+		(isAdFree as jest.Mock).mockReturnValue(true);
```
cemms1: _optional_ you can also do this to avoid using type assertions

```suggestion
		jest.mocked(isAdFree).mockReturnValue(true);;
```

cemms1 (approved):
🧹

bundle/src/init/consented/prepare-prebid.spec.ts
dskamiotis: Good shout — I'll update all the type assertions in this file to use jest.mocked() for consistency.

bundle/src/init/consented/ad-free-slot-remove.ts
dskamiotis: I hit issues testing adFreeSlotRemove directly because once() caches after the first call, and jest.resetAllMocks() doesn’t clear that internal cache. So only the first test was actually exercising the function.
That’s why I pulled the inner function out as removeAdFreeSlots and exposed it via export const _ = { removeAdFreeSlots }, so tests can call the unwrapped function directly, while keeping adFreeSlotRemove = once(removeAdFreeSlots) for production.
I noticed we use the same pattern in prepare-prebid.ts with _ = { setupPrebid } so I followed that approach.

PR #2530, opened 2026-04-17
Set prebid price floor V1

dskamiotis:
## What does this change?
Adds floor price configuration to Prebid via [pbjs.setConfig](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), gated behind an A/B test (commercial-prebid-price-floor).

We're applying a static $0.10 global floor — the minimum we'd ever floor at — to assess the impact on bid response rate.

This V1 approach:

Applies a single global floor (*|*: all media types, all sizes)
Ignores context (geo, device, slot, format)
Tests whether filtering low bids improves effective bidding
Tests added to verify floor config is included/excluded based on test group membership.

## Why?
We currently don't pass floor prices to our Prebid partners. Several partners have requested this, stating it helps filter unwinnable bids and improves effective bidding.

If V1 shows positive results, V2 would introduce more granular floors across dimensions like region, device, and slot.

### Testing:

https://theguardian.com/ab-tests/opt-in/commercial-prebid-price-floor:variant

Go to an article to can check in Chrome dev tools to confirm you're in the test:
window.guardian.modules.abTests.getParticipations()

Screenshot 2026-04-22 at 16 06 08

---
  - https://app.asana.com/0/0/1214026334703343

### DIFF
--- a/bundle/src/lib/header-bidding/prebid/index.ts
+++ b/bundle/src/lib/header-bidding/prebid/index.ts
@@ -49,6 +49,11 @@ const initialise = async (
 		'variant',
 	);
 
+	const isInPrebidFloorPriceTest = isUserInTestGroup(
+		'commercial-prebid-price-floor',
+		'variant',
+	);
+
 	// For control group users, await userSync before setConfig so it's included immediately.
 	// For test group users, skip the await — userSync will be merged into the config
 	// at the end of initialise via mergeConfig.
@@ -60,6 +65,20 @@ const initialise = async (
 		 */
 		bidderTimeout: PREBID_TIMEOUT,
 
+		...(isInPrebidFloorPriceTest
+			? {
+					floors: {
+						data: {
+							schema: {
+								fields: ['mediaType', 'size'],
+							},
+							values: {
+								'*|*': 0.1,
+							},
+						},
+					},
+				}
+			: {}),
 		/**
 		 * Prebid supports an additional timeout buffer to account for noisiness in
 		 * timing JavaScript on the page. This value is passed to the Prebid config
--- a/bundle/src/lib/header-bidding/prebid/index.ts
+++ b/bundle/src/lib/header-bidding/prebid/index.ts
@@ -64,17 +64,18 @@ const initialise = async (
 		 * The amount of time reserved for the auction
 		 */
 		bidderTimeout: PREBID_TIMEOUT,
-
+		/**
+		 * Applying one global floor price of £0.10 for all bids.
+		 * Initially gated behind an AB test.
+		 */
 		...(isInPrebidFloorPriceTest
 			? {
 					floors: {
+						enabled: true,
 						data: {
-							schema: {
-								fields: ['mediaType', 'size'],
-							},
-							values: {
-								'*|*': 0.1,
-							},
+							schema: { fields: ['mediaType'] },
+							values: { '*': 0.1 },
+							default: 0.1,
 						},
 					},
 				}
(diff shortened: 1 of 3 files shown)


bundle/src/lib/header-bidding/prebid/index.ts
```diff
@@ -60,6 +65,20 @@ const initialise = async (
...
 
+		...(isInPrebidFloorPriceTest
+			? {
+					floors: {
+						data: {
+							schema: {
+								fields: ['mediaType', 'size'],
+							},
+							values: {
+								'*|*': 0.1,
```
dskamiotis: This implemetation is different to the suggestion approach in the Ticket using:
```
pbjs.setConfig({
  floors: {
    enabled: true,
    data: {
      default: 0.10
    } ...
```

FWICS in the prebid docs/research `default` seems to be a "fallback if rule lookup fails".

This approach ensures: any media type and any size - to ensure we capture and apply to all ad bids

bundle/src/lib/header-bidding/prebid/index.ts
```diff
@@ -60,6 +65,20 @@ const initialise = async (
...
+							},
+						},
+					},
+				}
+			: {}),
```
cemms1, line 82: Maybe we could include a comment above here to explain what we're doing, as we've done for some of the other Prebid config items?

bundle/src/lib/header-bidding/prebid/index.ts
```diff
@@ -60,6 +65,20 @@ const initialise = async (
 		 */
 		bidderTimeout: PREBID_TIMEOUT,
 
+		...(isInPrebidFloorPriceTest
+			? {
+					floors: {
+						data: {
+							schema: {
+								fields: ['mediaType', 'size'],
```
cemms1: Why are we specifying `mediaType` and `size`?

bundle/src/lib/header-bidding/prebid/index.ts
cemms1: I think this original implementation task is to set a default to fall back to, rather than to state an explicit minimum across the board, though I realise that's basically the same thing..

The ticket says:
> For V1, we can add a simple bit of config (sourced [here](https://github.com/prebid/Prebid.js/blob/master/modules/priceFloors.md)) but set the floor as enabled and a default value of 0.10 (default is USD).

What it looks like you've done here is set a schema for your price floor lookup (`'${mediaType}|${size}'`) and then set that as `*|*`. I don't think we need this level of granularity until we do the second stage of this work

cemms1 (commented):
According to [this video and transcript](https://docs.prebid.org/dev-docs/floors-video-overview.html), I think we need to also install the `priceFloors` module. ([`modules/priceFloors`](https://github.com/prebid/Prebid.js/blob/b29f61a14d361e75bebc56f92334ff1f59586e20/modules/priceFloors.ts))

bundle/src/lib/header-bidding/prebid/index.ts
dskamiotis: FYI — we'd ideally just use data: { default: 0.1 }, but the @types/prebid.js FloorsConfig type requires `schema` and `values` as mandatory fields on data. So we've included schema: { fields: ['mediaType'] } with values: { '*': 0.1 } alongside default: 0.1 — functionally equivalent to a single catch-all floor.

cemms1 (approved):

PR #2535, opened 2026-04-23
update and fix `hangtime` conditions in US for inline2

dskamiotis:
## What does this change?
Adds Ozone hangtime placement ID (1500001025) for mobile inline2 slots in the US, by placing the check before the generic MPU fallback (1500001036) inside the isInUsa() block.

Also fixes the existing test for US inline2 hangtime, which was previously mocking isInUsOrCa instead of isInUsa — passing accidentally via the UK/ROW fallback path rather than the US code path.

## Why?
The hangtime placement ID (1500001025) was unreachable for US traffic. The generic mobile MPU fallback (1500001036) returned first, so inline2 in the US never got the hangtime-specific placement.

### Test

Tested in CODE:

| | Before - 1500001036 | After - 1500001025 |
|---|---|---|
| US mobile inline2 | Screenshot 2026-04-23 at 17 07 50 | Screenshot 2026-04-23 at 17 05 06 |

  | After
-- | --
UK mobile inline2 still hangtime: 1500001025 | Screenshot 2026-04-23 at 17 50 04

### DIFF
--- a/bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
+++ b/bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
@@ -755,7 +755,7 @@ describe('getOzonePlacementId', () => {
 		);
 	});
 
-	test('should return correct placementID for hangtime ads in inline2 in US or CAN', () => {
+	test('should return correct placementID for hangtime ads in inline2 in US', () => {
 		isInUsa.mockReturnValue(true);
 		getBreakpointKey.mockReturnValue('M');
 		containsMpu.mockReturnValue(true);
--- a/bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
+++ b/bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
@@ -730,13 +730,6 @@ describe('getOzonePlacementId', () => {
 		expect(getOzonePlacementId([[320, 50]])).toBe('3500014217');
 	});
 
-	test('should return correct placementID in US if none of the conditions are met', () => {
-		isInUsa.mockReturnValue(true);
-		getBreakpointKey.mockReturnValue('M');
-		containsMpu.mockReturnValue(true);
-		expect(getOzonePlacementId([[300, 250]])).toBe('1500001036');
-	});
-
 	test('should return correct placementID for mobile-sticky in ROW', () => {
 		isInRow.mockReturnValue(true);
 		getBreakpointKey.mockReturnValue('M');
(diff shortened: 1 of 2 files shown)


bundle/src/lib/header-bidding/prebid/bidders/config.ts
cemms1, line 310: Shall we remove these lines if we've put it further up?

bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
```diff
@@ -763,7 +756,7 @@ describe('getOzonePlacementId', () => {
 	});
 
 	test('should return correct placementID for hangtime ads in inline2 in US or CAN', () => {
```
cemms1:
```suggestion
	test('should return correct placementID for hangtime ads in inline2 in the US', () => {
```

bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
```diff
@@ -730,13 +730,6 @@ describe('getOzonePlacementId', () => {
 		expect(getOzonePlacementId([[320, 50]])).toBe('3500014217');
 	});
 
-	test('should return correct placementID in US if none of the conditions are met', () => {
-		isInUsa.mockReturnValue(true);
-		getBreakpointKey.mockReturnValue('M');
-		containsMpu.mockReturnValue(true);
-		expect(getOzonePlacementId([[300, 250]])).toBe('1500001036');
-	});
-
```
cemms1, line 739: I think this test is still relevant?

bundle/src/lib/header-bidding/prebid/bidders/config.ts
dskamiotis: Ah yes this handles inline2 hangtime for non-US regions (UK, ROW, AU/NZ). US traffic returns early from the isInUsa() block above and never reaches here, so we need both checks.

bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
dskamiotis: This was a duplicate — the same mocks and assertion already exist in this [test](https://github.com/guardian/commercial/blob/main/bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts#L719) with a more accurate description. So I removed the duplicate rather than keeping two identical tests with different names.

cemms1: gotcha!

cemms1 (approved):

PR #2542, opened 2026-04-28
Put spacefinder in interactives live

cemms1:
## What does this change?

Removes the code referencing the holdback AB test `commercial-spacefinder-in-interactives` to put the logic live for 100% of the audience

## Why?

We've had the results back from this AB test and allowing spacefinder to run in interactive articles results in a **3% decrease** in good CLS scores but a **43% increase** in ad ratio (and 44% increase in ad requests) so we've agreed to fully release the change and continue to adapt the spacefinder rules to reduce the negative impact on CLS going forward

### DIFF
--- a/bundle/jest.config.js
+++ b/bundle/jest.config.js
@@ -11,7 +11,7 @@ const jestConfig = {
 			branches: 45,
 			functions: 41,
 			lines: 60,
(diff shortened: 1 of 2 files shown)


dskamiotis (approved):
👍

PR #2543, opened 2026-04-28
Remove `commercial-loading-userids-async` test logic

cemms1:
## What does this change?

Removes the `commercial-loading-userids-async` AB test logic to fall back to the original logic implemented before #2429

## Why?

Removes code referencing this AB test and reverts it back to the original logic until we decide whether to move forward with the change or not.


dskamiotis (approved):
This looks good and probably makes the code a bit easier to reason about given the marginal improvement it might add to performance. Happy to approve if we want to go ahead with this.

PR #2549, opened 2026-05-05
Migrate cmp from guardian/libs to consent-manager

akinsola-guardian:
## What does this change?

In this [PR](https://github.com/guardian/csnx/pull/2358), we migrated the consent-manager-platform to it's own library within csnx.

This PR is migrating all the consent logic from guardian/libs to guardian/consent-manager. 

## Why?

The consent-management-platform was migrated from guardian/libs to guardian/consent-manager.


.changeset/full-dragons-end.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/commercial-core': patch
```
cemms1: We're updating the peer dependency list with a new peer dependency so I think this technically counts as a breaking change?

```suggestion
'@guardian/commercial-core': major
```

cemms1 (approved):
LGTM! 🚀

PR #2550, opened 2026-05-12
Enable IntentIQ analytics with GAM object reference

dskamiotis:
## What does this change?
Adds `gamObjectReference` IntentIQ analytics adapter configuration to Prebid.js initialisation.
- Updated index.ts to pass both analytics configs to pbjs.enableAnalytics(), independently — so if one is undefined, the other still gets enabled
- Adds consent and AB test gating before enabling the IntentIQ adapter

- The IntentIQ analytics module (prebid.js/modules/intentIqAnalyticsAdapter) was already imported but had no enableAnalytics config — this PR activates it.

## Why?
Allows the IntentIQ analytics adapter to access Google Publisher Tag (GPT) slot context in the browser, enabling enriched analytics reporting based on auction and impression lifecycle events.
We only want the feature to be available to users who have consented to `intentIQ` and initially part of the test variant group.


bundle/src/lib/header-bidding/prebid/index.ts
```diff
@@ -121,8 +122,16 @@ const initialise = async (
...
 	// configure analytics
-	const analytics = getAnalyticsConfig();
-	if (analytics) window.pbjs.enableAnalytics([analytics]);
+	const guAnalytics = getGUAnalyticsConfig();
+	const intentIQAnalyitics = getIntentIQAnalyticsConfig();
+	const filteredAnalytics: Array<AnalyticsConfig<string>> = [];
+	if (guAnalytics) {
+		filteredAnalytics.push(guAnalytics);
+	}
+	if (intentIQAnalyitics) {
```
GHaberis: We'll need a consent check for IntentIQ, also an AB test participation check before we include the `intentIQAnalyitics` config. You could put them in the `getIntentIQAnalyticsConfig` function.

bundle/src/lib/header-bidding/prebid/analytics.ts
```diff
@@ -51,3 +51,19 @@ export const getAnalyticsConfig = (): AnalyticsConfig<'gu'> | undefined => {
 		};
 	}
 };
+
+export const getIntentIQAnalyticsConfig = ():
+	| AnalyticsConfig<'iiqAnalytics'>
+	| undefined => {
+	if (shouldEnableAnalytics()) {
```
GHaberis: I don't think the `shouldEnableAnalytics()` check is relevant for `IntentIQ`, it seems related to the `gu` analytics module specifically.

bundle/src/lib/header-bidding/prebid/index.ts
```diff
@@ -121,8 +122,23 @@ const initialise = async (
...
+		'commercial-user-module-intentIq',
+		'variant',
+	);
+
+	const analytics: Array<AnalyticsConfig<string>> = [];
+	const guAnalytics = getGUAnalyticsConfig();
+	if (guAnalytics) {
+		analytics.push(guAnalytics);
+	}
+	const intentIQAnalytics = isUserInTestGroupIntentIQ
```
cemms1: _nitpick (optional)_
I think it might be slightly easier to read if we kept the `isUserInTestGroupIntentIQ` check alongside the `getConsent` check. Having it in two separate layers makes it slightly harder to follow. 
Not a biggie though!

cemms1 (approved):

bundle/src/lib/header-bidding/prebid/index.ts
dskamiotis: Great 👍 I have added the AB test gate on this top level so it wont attempt to run the function if not in the variant - and perhaps more obvious to remove later.

bundle/src/lib/header-bidding/prebid/index.ts
dskamiotis: This makes more sense 👍

PR #2557, opened 2026-05-15
Fix mobile-sticky e2e test with dismissable banners

cemms1:
## What does this change?

Updates the mobile-sticky Playwright test for the reader revenue dismissable banners to ensure it works for both two step and single step banners

## Why?

We don't reliably know which banner type we are going to get so we need to be able to handle both scenarios in this test

### DIFF
--- a/bundle/playwright/tests/mobile-sticky-slot.spec.ts
+++ b/bundle/playwright/tests/mobile-sticky-slot.spec.ts
@@ -18,7 +18,6 @@ test.describe('mobile-sticky', () => {
 		}) => {
 			await page.setViewportSize({ width, height });
 			const queryParams = {
-				'ab-commercial-mobile-sticky': 'variant',
 				adtest: 'mobileStickyTest',
 				'force-banner': '',
 			};
@@ -36,7 +35,6 @@ test.describe('mobile-sticky', () => {
 			}) => {
 				await page.setViewportSize({ width, height });
 				const queryParams = {
-					'ab-commercial-mobile-sticky': 'variant',
 					adtest: 'mobileStickyTest',
 					'force-banner': '',
 				};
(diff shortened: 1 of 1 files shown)


bundle/playwright/tests/mobile-sticky-slot.spec.ts
```diff
@@ -18,7 +18,6 @@ test.describe('mobile-sticky', () => {
 		}) => {
 			await page.setViewportSize({ width, height });
 			const queryParams = {
-				'ab-commercial-mobile-sticky': 'variant',
```
cemms1, line 21: Tidying up 🧹 since the AB test has been removed now

bundle/playwright/tests/mobile-sticky-slot.spec.ts
```diff
@@ -96,20 +91,21 @@ test.describe('mobile-sticky', () => {
...
-					.click();
+			const twoStepBanner = page
+				.getByRole('alert')
+				.getByRole('button', { name: 'Collapse banner' });
+
+			const oneStepBanner = page
+				.getByRole('alert')
+				.getByRole('button', { name: 'Close' });
+
+			expect(twoStepBanner.or(oneStepBanner)).toBeVisible;
```
cemms1, line 102: We now check that at least one of these two banner types is visible before continuing

bundle/playwright/tests/mobile-sticky-slot.spec.ts
```diff
@@ -96,20 +91,21 @@ test.describe('mobile-sticky', () => {
...
+
+			if (await twoStepBanner.isVisible()) {
```
cemms1, line 104: Instead of a try/catch block, we check if the two step banner is the one shown and if not, try to dismiss the single step one

deedeeh (commented):
Great to tidy up on the go ✨

deedeeh (approved):

PR #2565, opened 2026-05-21
Deduplicate geo locale function

DanielCliftonGuardian:
## What does this change?

Removes the `country-code.ts` module from `commercial-core` and consolidates all geo-detection in the codebase onto a single synchronous `getLocale` function.
>[!NOTE]
>The `getLocale` from `@guardian/libs` is async and can call the Guardian geolocation API as a last resort. The commercial- core version is synchronous. In practice this should make no difference since the cookie should always be set, and the codebase has intentionally avoided async geo detection in the other implementation.

## Why?

Improves the codebase by reducing duplication, making it easier to understand for future developers working on the commercial code.

Resolves https://app.asana.com/1/1210045093164357/project/1213489522854406/task/1214542443579281


core/src/geo/geo-utils.ts
```diff
@@ -1,22 +1,14 @@
-import type { CountryCode } from '@guardian/libs';
-import { getCountryCode } from './country-code';
-
-// cache the users location so we only have to look it up once
-let geo: CountryCode | undefined;
-const currentGeoLocation = (): CountryCode => {
-	geo = geo ?? getCountryCode();
-	return geo;
-};
+import { __resetCachedValue, getLocale } from './get-locale';
```
deedeeh, line 1: `_resetCachedValue` has a [comment ](https://github.com/guardian/commercial/blob/fe92d68743f30c6644defb8a9566d2a39e882350/core/src/geo/get-locale.ts#L25-L26) it's used just for tests so maybe we can change the comment and mention that we use it in here.

deedeeh (commented):
@DanielCliftonGuardian your first contribution as a CommDev developer and it looks good 🎉 

There are a couple of minor things I want to mention regarding the PR so lets have a catchup and talk through it.

deedeeh (approved):
🚀

dskamiotis (commented):
This is much easier to reason about, and keeping it synchronous removes a layer of indirection too. 👍

PR #2567, opened 2026-05-22
integrate teads with prebid

dskamiotis:
## What does this change?
Adds a `teadsBidder` to the Prebid configuration
Includes within it a `getTeadsPlacementId` function to return the correct page and placement ID for the various, sizes, regions and breakpoints.

## Why?
The proposal is to start including Teads as bidder for each Prebid Ad Unit alongside our existing bidders.

We want to use the targeting defined in this sheet Global Placement Mapping so we can map the appropriate pageId and placementId to an Ad Unit's slot sizes.

We are using additional dimensions for targeting in order to select the pageId and placementId:
- Geo (UK, ROW and US) 
- Platform (everything will be ‘web’)
- Device Category (‘Mobile’ or ‘Desktop’)

### Tests

Tested with: http://localhost:3030/Article/https://www.theguardian.com/sport/2023/nov/22/ronnie-osullivan-warns-he-will-quit-snooker-if-he-cannot-play-in-china?ab-commercial-teads-prebid=variant

Screenshot 2026-05-27 at 14 21 10

`Teads` request

Screenshot 2026-05-28 at 12 17 10

### DIFF
--- a/bundle/src/lib/header-bidding/prebid-types.ts
+++ b/bundle/src/lib/header-bidding/prebid-types.ts
@@ -118,7 +118,8 @@ type BidderCode =
 	| 'triplelift'
 	| 'trustx'
 	| 'xhb'
-	| 'ttd';
+	| 'ttd'
+	| 'teads';
 
 type PrebidParams =
 	| PrebidAppNexusParams
@@ -146,6 +147,11 @@ type PrebidMediaTypes = {
 	};
 };
 
+type PrebidTeadsParams = {
+	pageId: number;
+	placementId: number;
+};
+
 type SlotFlatMap = (slot: HeaderBiddingSlot) => HeaderBiddingSlot[];
 
 export type {
@@ -169,5 +175,6 @@ export type {
 	PrebidParams,
 	PrebidBidder,
 	PrebidMediaTypes,
+	PrebidTeadsParams,
 	SlotFlatMap,
 };
--- a/bundle/src/lib/header-bidding/prebid-types.ts
+++ b/bundle/src/lib/header-bidding/prebid-types.ts
@@ -133,6 +133,7 @@ type PrebidParams =
 	| PrebidTripleLiftParams
 	| PrebidTrustXParams
 	| PrebidXaxisParams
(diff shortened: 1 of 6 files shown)


bundle/src/lib/header-bidding/prebid-types.ts
cemms1, line 137: Should we include `PrebidTeadsParams` here too?

bundle/src/lib/header-bidding/prebid/bidders/config.ts
```diff
@@ -538,6 +623,7 @@ const currentBidders = (
 		[shouldInclude('ozone'), ozoneBidder(pageTargeting)],
 		[shouldInclude('oxd'), openxBidder(pageTargeting)],
 		[shouldInclude('kargo'), kargoBidder],
+		[isInTeadsTest && shouldInclude('teads'), teadsBidder],
```
cemms1: We could put the `isInTeadsTest` inside `shouldInclude` for `teads` to keep all the conditional logic in the same place?

bundle/src/lib/header-bidding/prebid/bidders/config.ts
```diff
@@ -262,6 +267,73 @@ const openxBidder: (pageTargeting: PageTargeting) => PrebidBidder = (
 const isCrosswordPage = (pageTargeting: PageTargeting = {}) =>
 	pageTargeting.s === 'crosswords';
 
+const getTeadsPlacementId = (sizes: Size[]): PrebidTeadsParams | undefined => {
```
cemms1: Since we return a `PrebidTeadsParams` type which is a bit more than just a `placementId` maybe this function can be named `getTeadsParams`?

bundle/src/lib/header-bidding/prebid-types.ts
dskamiotis: Good spot! Yea we need this since `PrebidParams` is the return type of bidParams on `PrebidBidder`.

cemms1 (approved):

PR #2571, opened 2026-05-27
adds price floors AB test gate for users outside of holdback test

dskamiotis:
## What does this change?
This PR adds an A/B test gate for the Prebid Price Floors feature, ensuring that the appropriate price floor configuration is only passed to window.pbjs.setConfig for users enrolled in the test variant.
## Why?
We want to incrementally roll out different iterations of Prebid Price Floors behind A/B tests. As part of this rollout, 5% of users will remain in the control group without the feature enabled, while the remaining users will receive the Price Floors configuration.

---
  - https://app.asana.com/0/0/1215066183685575


DanielCliftonGuardian (approved):
👌

DanielCliftonGuardian (commented):
I have prematurely approved this. The changes look good but looks like the tests need updating with this change.

cemms1 (approved):

PR #2572, opened 2026-05-27
Add Permutive consent check to initPermutive

DanielCliftonGuardian:
## What does this change?
Added a `getConsentFor('permutive', consentState)` check inside `initPermutive`, following the same pattern already [used by A9](https://github.com/guardian/commercial/blob/main/bundle/src/init/consented/prepare-a9.ts#L43). If a user specifically opts out of Permutive, all related setup is skipped.

## Why?
Ensure privacy/GDPR is comprehensively respected

---
  - https://app.asana.com/0/0/1213805716944216

### DIFF
--- a/bundle/src/init/consented/prepare-permutive.spec.ts
+++ b/bundle/src/init/consented/prepare-permutive.spec.ts
@@ -437,10 +437,6 @@ describe('Generating Permutive payload utils', () => {
 			window.guardian.config.switches.permutive = true;
 		});
 
-		afterEach(() => {
-			window.guardian.config.switches.permutive = false;
-		});
-
 		it('does not call getEmail when Permutive consent is not granted', async () => {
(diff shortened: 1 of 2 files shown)


bundle/src/init/consented/prepare-permutive.spec.ts
```diff
@@ -419,4 +426,31 @@ describe('Generating Permutive payload utils', () => {
...
+			(getConsentFor as jest.Mock).mockReturnValue(hasConsent);
+
+		beforeEach(() => {
+			window.guardian.config.switches.permutive = true;
+		});
+
+		afterEach(() => {
+			window.guardian.config.switches.permutive = false;
+		});
+
```
cemms1: I don't think we need this `afterEach`

cemms1 (approved):
LGTM!

PR #2575, opened 2026-06-04
Add US-specific IntentIQ gating for ID and analytics with tests

dskamiotis:
## What does this change?
Adds US region holdback gating for the IntentIQ user ID and analytics. 

## Why?
IntentIQ is currently enabled through the existing experiment setup across regions.
For the US, we now want a dedicated holdback model so that:

**Expected traffic split**
US:

Control: 5% -> 5%
Variant: 5% -> 95%
All other regions:

Control: unchanged at 5%
Variant: unchanged at 5%

This PR also updates tests to cover US holdback and non-holdback paths.

### DIFF
--- a/bundle/src/lib/header-bidding/prebid/id-handlers/intent-iq.spec.ts
+++ b/bundle/src/lib/header-bidding/prebid/id-handlers/intent-iq.spec.ts
@@ -101,3 +101,74 @@ describe('getUserIdForIntentIQ - when user is NOT in test group', () => {
 		expect(result).toEqual(undefined);
 	});
 });
+
+describe('getUserIdForIntentIQ - when user is in US region test group', () => {
+	beforeEach(() => {
+		jest.resetAllMocks();
+	});
+	test('when locale is US and US experiment is variant', async () => {
+		jest.mocked(isUserInTestGroup).mockReturnValueOnce(false); // isUserInTestGroupIntentIQ
+		jest.mocked(isUserInTestGroup).mockReturnValueOnce(true); // isUserInTestGroupIntentIQForUS
+		jest.mocked(getLocale).mockReturnValue('US');
+
+		const result = await getUserIdForIntentIQ();
+
+		expect(result).toEqual({
+			name: 'intentIqId',
+			params: {
+				partner: 377078111,
+				gamObjectReference: {},
+			},
+			storage: {
+				type: 'html5',
+				name: 'intentIqId',
+				expires: 0,
+				refreshInSeconds: 0,
+			},
+		});
+	});
+	test('when locale is US and both experiments are variant', async () => {
+		jest.mocked(isUserInTestGroup).mockReturnValueOnce(true); // isUserInTestGroupIntentIQ
+		jest.mocked(isUserInTestGroup).mockReturnValueOnce(true); // isUserInTestGroupIntentIQForUS
+		jest.mocked(getLocale).mockReturnValue('US');
+
+		const result = await getUserIdForIntentIQ();
+
+		expect(result).toEqual({
+			name: 'intentIqId',
+			params: {
+				partner: 377078111,
+				gamObjectReference: {},
+			},
+			storage: {
+				type: 'html5',
+				name: 'intentIqId',
+				expires: 0,
+				refreshInSeconds: 0,
+			},
+		});
+	});
+});
+describe('getUserIdForIntentIQ - when user is not in US region test group', () => {
+	beforeEach(() => {
+		jest.resetAllMocks();
+	});
+	test('when locale is US and US experiment is control', async () => {
+		jest.mocked(isUserInTestGroup).mockReturnValueOnce(false); // isUserInTestGroupIntentIQ
+		jest.mocked(isUserInTestGroup).mockReturnValueOnce(false); // isUserInTestGroupIntentIQForUS
+		jest.mocked(getLocale).mockReturnValueOnce('US');
+
+		const result = await getUserIdForIntentIQ();
+
+		expect(result).toEqual(undefined);
+	});
+	test('when locale is US and only global experiment is variant', async () => {
+		jest.mocked(isUserInTestGroup).mockReturnValueOnce(true); // isUserInTestGroupIntentIQ
+		jest.mocked(isUserInTestGroup).mockReturnValueOnce(false); // isUserInTestGroupIntentIQForUS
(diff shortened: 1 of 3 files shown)


bundle/src/lib/header-bidding/prebid/intent-iq-analytics.ts
```diff
@@ -14,15 +15,27 @@ const isUserInTestGroupIntentIQ = isUserInTestGroup(
 	'variant',
 );
 
+const isUserInTestGroupIntentIQForUS = isUserInTestGroup(
+	'commercial-user-module-intentIq-us-region',
+	'variant',
+);
```
cemms1, line 21: Suggest inverting this logic to check if _not_ in the hold back group instead

bundle/src/lib/header-bidding/prebid/intent-iq-analytics.ts
```diff
@@ -14,15 +15,27 @@ const isUserInTestGroupIntentIQ = isUserInTestGroup(
...
+
+const isUserInUSRegion = () => getLocale() === 'US';
```
cemms1: You can use the `isInUsa` function from `geo-utils`

bundle/src/lib/header-bidding/prebid/id-handlers/intent-iq.ts
```diff
@@ -36,8 +36,16 @@ const getUserIdForIntentIQ = async (): Promise<
 		'commercial-user-module-intentIq',
 		'variant',
 	);
+	const isUserInTestGroupIntentIQForUS = isUserInTestGroup(
+		'commercial-user-module-intentIq-us-region',
+		'variant',
+	);
 	const isEU = isUserInAllowedEURegion();
-	if (isUserInTestGroupIntentIQ && isUserInIntentIQRegion()) {
+	const isUS = isUserInUSRegion();
```
cemms1: You can use `isInUsa` from `geo-utils`

bundle/src/lib/header-bidding/prebid/id-handlers/intent-iq.ts
```diff
@@ -36,8 +36,16 @@ const getUserIdForIntentIQ = async (): Promise<
 		'commercial-user-module-intentIq',
 		'variant',
 	);
+	const isUserInTestGroupIntentIQForUS = isUserInTestGroup(
+		'commercial-user-module-intentIq-us-region',
+		'variant',
+	);
```
cemms1:
```suggestion
	const canRunIntentIqInUS = isUserInTestGroup(
		'commercial-user-module-intentIq-us-region',
		'variant',
	);
```

bundle/src/lib/header-bidding/prebid/id-handlers/intent-iq.ts
```diff
@@ -36,8 +36,16 @@ const getUserIdForIntentIQ = async (): Promise<
...
 	const isEU = isUserInAllowedEURegion();
-	if (isUserInTestGroupIntentIQ && isUserInIntentIQRegion()) {
+	const isUS = isUserInUSRegion();
+	if (
+		(isUserInTestGroupIntentIQ && isUserInIntentIQRegion()) ||
+		(isUS && isUserInTestGroupIntentIQForUS)
```
cemms1: Keeping a consistent order of logic on both lines would make this slightly easier to read

ie instead of
```
(isInTestGroup && isInRegion) || (isInRegion && isInTestGroup)
```

do something like
```
(isInTestGroup && isInRegion) || (isInTestGroup && isInRegion)
```

bundle/src/lib/header-bidding/prebid/intent-iq-analytics.ts
cemms1: If inverting, please add additional JSDoc comment above describing _why_

e.g.
```js
/**
 * Released to all audience apart from holdback group in the US
 */
const isUserInHoldbackGroupIntentIQForUS = isUserInTestGroup(
	'commercial-user-module-intentIq-us-region',
	'holdback',
);
```

bundle/src/lib/header-bidding/prebid/intent-iq-analytics.ts
```diff
@@ -14,15 +15,25 @@ const isUserInTestGroupIntentIQ = isUserInTestGroup(
...
-		isUserInTestGroupIntentIQ &&
-		getConsentFor('intentIQ', consentState) &&
-		isUserInIntentIQRegion()
-	) {
+	const hasConsent = getConsentFor('intentIQ', consentState);
+
+	const isNonUSEligible =
+		isUserInTestGroupIntentIQ && isUserInIntentIQRegion();
+
+	const isUSEligible = canRunIntentIqInUS && isInUsa();
```
cemms1, line 32: Ideally both const values would be named similarly - both named `canRun...` or both named `isInTest...`

bundle/src/lib/header-bidding/prebid/id-handlers/intent-iq.spec.ts
```diff
@@ -101,3 +105,74 @@ describe('getUserIdForIntentIQ - when user is NOT in test group', () => {
...
+	test('when locale is US and US experiment is in holdback', async () => {
+		jest.mocked(isUserInTestGroup).mockReturnValueOnce(false); // isUserInTestGroupIntentIQ
+		jest.mocked(isUserInTestGroup).mockReturnValueOnce(true); // canRunIntentIqInUS
+		jest.mocked(isInUsa).mockReturnValueOnce(true);
+
+		const result = await getUserIdForIntentIQ();
+
+		expect(result).toEqual(undefined);
+	});
+	test('when locale is US and only global experiment is variant and also in holdback', async () => {
```
cemms1, line 169: what does `and only global experiment is variant` mean?

cemms1 (approved):

bundle/src/lib/header-bidding/prebid/id-handlers/intent-iq.spec.ts
dskamiotis: made this a bit clearer:
`when locale is US, global experiment is variant, and US user is in holdback`

bundle/src/lib/header-bidding/prebid/intent-iq-analytics.ts
dskamiotis: changed the other allowed regions to `canRunIntentIqInAllowedRegions`

PR #2576, opened 2026-06-08
Update IntentIQ US test exposure and refine naming for consistency

dskamiotis:
## What does this change?
Updates naming of test group inclusion group: isUserInTestGroupIntentIQ > canRunIntentIqInAllowedRegions
## Why?
This updates the variable naming of the allowed groups in the intentIQ test groups for consistency with the naming of those not in the test groups, so its similar to `canRunIntentIqInUS`


deedeeh (approved):
That's easier to follow 👍🏼

PR #2588, opened 2026-06-15
enables mobile-sticky on US liveblog articles

dskamiotis:
## What does this change?
This change extends mobile-sticky ad eligibility to US LiveBlog pages.

- Adds a dedicated US LiveBlog eligibility check.
- Keeps existing mobile-sticky eligibility logic unchanged for current supported page types.
- Gates the US LiveBlog path behind the `commercial-mobile-sticky-liveblog-us` AB test.
- Uses a holdback setup so users in the holdback variant do not receive mobile-sticky on US LiveBlog pages.

## Why?
We want to safely roll out mobile-sticky ads on US LiveBlog pages while preserving existing behaviour elsewhere.
A 5% holdback lets us compare outcomes against a control group before wider rollout.

### DIFF
--- a/bundle/src/lib/header-bidding/utils.ts
+++ b/bundle/src/lib/header-bidding/utils.ts
@@ -248,6 +248,8 @@ export const shouldIncludeMobileSticky = once(
 		}) &&
 			!isInUk() &&
 			(isValidPageForMobileSticky() ||
+				// We intentionally use variant as the holdback arm for this test.
+				// Only users not in variant are eligible for this LiveBlog US mobile-sticky path.
 				(!isUserInTestGroup(
 					'commercial-mobile-sticky-liveblog-us',
 					'variant',
--- a/bundle/src/lib/header-bidding/utils.ts
+++ b/bundle/src/lib/header-bidding/utils.ts
@@ -50,15 +50,6 @@ const isValidPageForMobileSticky = (): boolean => {
 	);
 };
 
-const isLiveBlogPageInUs = (): boolean => {
-	const { contentType } = window.guardian.config.page;
-	return contentType === 'LiveBlog' && isInUsa();
-};
-
-// Use a holdback AB test: users in holdback do not get mobile-sticky on eligible US LiveBlog pages.
-const isMobileStickyLiveblogUsEnabled = (): boolean =>
-	!isUserInTestGroup('commercial-mobile-sticky-liveblog-us', 'holdback');
-
 /**
  * Cleans an object for targetting. Removes empty strings and other falsy values.
  * @param o object with falsy values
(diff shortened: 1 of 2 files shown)


bundle/src/lib/header-bidding/utils.ts
```diff
@@ -50,6 +50,15 @@ const isValidPageForMobileSticky = (): boolean => {
 	);
 };
 
+const isLiveBlogPageInUs = (): boolean => {
+	const { contentType } = window.guardian.config.page;
+	return contentType === 'LiveBlog' && isInUsa();
+};
+
+// Use a holdback AB test: users in holdback do not get mobile-sticky on eligible US LiveBlog pages.
```
cemms1: I think we agreed to run a 10% test with 5% control and 5% variant rather than a holdback test?

bundle/src/lib/header-bidding/utils.ts
```diff
@@ -247,7 +256,8 @@ export const shouldIncludeMobileSticky = once(
 			max: 'mobileLandscape',
 		}) &&
 			!isInUk() &&
-			isValidPageForMobileSticky() &&
+			(isValidPageForMobileSticky() ||
+				(isMobileStickyLiveblogUsEnabled() && isLiveBlogPageInUs())) &&
```
cemms1: I don't think there's any need to extract this logic out into separate functions. I think it might be easier to read if it was all here:


```suggestion
				(!isUserInTestGroup('commercial-mobile-sticky-liveblog-us', 'holdback') && window.guardian.config.page.contentType === "Liveblog" && isInUsa()) &&
```

cemms1: Or alternative would be to add that logic to `isValidPageForMobileSticky` or to name this logic something like `isValidForUSLiveblog`

bundle/src/lib/header-bidding/utils.ts
dskamiotis: The code behaviour was intended as everyone except holdback receives the feature, but agreed the wording is confusing - I will update the comment to explicitly describe the 10% split and expected behaviour for each cohort.

bundle/src/lib/header-bidding/utils.ts
dskamiotis: Ok sure I have also renamed the `holdback` to `variant` as suggested in the DCR PR and to keep the test config naming consistent

dskamiotis: Nearly got caught out here with `Liveblog` instead of `LiveBlog`

bundle/src/lib/header-bidding/utils.ts
```diff
@@ -247,7 +247,13 @@ export const shouldIncludeMobileSticky = once(
 			max: 'mobileLandscape',
 		}) &&
 			!isInUk() &&
-			isValidPageForMobileSticky() &&
+			(isValidPageForMobileSticky() ||
+				(!isUserInTestGroup(
```
cemms1, line 253: Please could we add a comment above this line to say which group does which? To make it easier for us to check the intended logic

cemms1 (approved):

PR #2594, opened 2026-06-25
Update Teads Prebid inline1 MPU mapping and add AU/NZ placement IDs

dskamiotis:
## What does this change?
- Updates `Teads` placement mapping for inline1 slots on desktop when the ad size is MPU [300, 250], across UK, RoW, US, and AU/NZ.
- Adds full `Teads` page and placement ID mappings for AU/NZ across desktop and mobile size groups.
- Removes AB test gate and exposes the feature to all users.

## Why?
This extends the Teads + Prebid integration introduced in PR #2567 and delivers the contractual mapping updates for inline1 and AU/NZ regions.

### Tested Slot sizes in CODE  for new AUS and NZ region:

| dfp-inline1 MPU | Billboard + Leaderboard |
|-----| -----|
| Screenshot 2026-06-26 at 12 30 30 |  Screenshot 2026-06-26 at 12 30 38 |  
| Screenshot 2026-06-26 at 12 28 22 |  Screenshot 2026-06-26 at 12 28 37 | 

### Desktop/Tablet inline1 slot with the new assigned Page ID and Placement ID
| Region / mapping | Evidence |
| --- | --- |
| UK | Screenshot 2026-06-26 at 15 59 15 |
| AU/NZ | Screenshot 2026-06-26 at 16 12 35 |
| RoW (FR, CY) | Screenshot 2026-06-26 at 16 06 59 |
| US | Screenshot 2026-06-26 at 16 15 31 |

### DIFF
--- a/bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
+++ b/bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
@@ -829,7 +829,7 @@ describe('getTeadsParams', () => {
 				});
 			},
 		);
-		test('should return correct pageD and placementID for MPU sized inline1 slots, in Uk when it is on desktop', () => {
+		test('should return correct pageId and placementID for MPU sized inline1 slots, in Uk when it is on desktop', () => {
 			isInUk.mockReturnValue(true);
 			getBreakpointKey.mockReturnValue('D');
 			containsMpu.mockReturnValue(true);
@@ -887,7 +887,7 @@ describe('getTeadsParams', () => {
 				});
 			},
 		);
-		test('should return correct pageD and placementID for MPU sized inline1 slots, in RoW when it is on desktop', () => {
+		test('should return correct pageId and placementID for MPU sized inline1 slots, in RoW when it is on desktop', () => {
 			isInRow.mockReturnValue(true);
 			getBreakpointKey.mockReturnValue('D');
 			containsMpu.mockReturnValue(true);
@@ -945,7 +945,7 @@ describe('getTeadsParams', () => {
 				});
 			},
 		);
-		test('should return correct pageD and placementID for MPU sized inline1 slots, in RoW when it is on desktop', () => {
+		test('should return correct pageId and placementID for MPU sized inline1 slots, in US when it is on desktop', () => {
 			isInUsa.mockReturnValue(true);
 			getBreakpointKey.mockReturnValue('D');
 			containsMpu.mockReturnValue(true);
@@ -1003,7 +1003,7 @@ describe('getTeadsParams', () => {
 				});
 			},
 		);
-		test('should return correct pageD and placementID for MPU sized inline1 slots, in AU/NZ when it is on desktop', () => {
+		test('should return correct pageId and placementId for MPU sized inline1 slots, in AU/NZ when it is on desktop', () => {
 			isInAuOrNz.mockReturnValue(true);
 			getBreakpointKey.mockReturnValue('D');
 			containsMpu.mockReturnValue(true);
--- a/bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
+++ b/bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
@@ -503,6 +503,49 @@ describe('bids', () => {
 			unit: '560429384',
 		});
 	});
+
+	test('should pass inline1 slotId through to Teads params for desktop MPU in UK', () => {
+		const mockShouldInclude = jest.fn().mockReturnValue(true);
+		jest.mocked(shouldIncludeBidder).mockReturnValue(mockShouldInclude);
+		isInUk.mockReturnValue(true);
+		getBreakpointKey.mockReturnValue('D');
+		containsMpu.mockReturnValue(true);
+
+		const teadsBid = bids(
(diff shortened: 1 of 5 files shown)


bundle/src/lib/header-bidding/prebid/bidders/config.ts
```diff
@@ -266,11 +266,19 @@ const openxBidder: (pageTargeting: PageTargeting) => PrebidBidder = (
...
 	const isDesktop = breakpoint === 'D' || breakpoint === 'T';
 
 	if (isInUk()) {
+		if (isDesktop) {
+			if (slotId === 'dfp-ad--inline1' && containsMpu(sizes)) {
+				return { pageId: 265029, placementId: 248133 };
+			}
+		}
 		if (isDesktop) {
 			if (
```
DanielCliftonGuardian, line 281: Could we merge these `if (isDesktop)` blocks

bundle/src/lib/header-bidding/prebid/bidders/config.ts
```diff
@@ -289,6 +297,11 @@ const getTeadsParams = (sizes: Size[]): PrebidTeadsParams | undefined => {
 		}
 	}
 	if (isInRow()) {
+		if (isDesktop) {
+			if (slotId === 'dfp-ad--inline1' && containsMpu(sizes)) {
+				return { pageId: 265030, placementId: 248134 };
+			}
+		}
 		if (isDesktop) {
```
DanielCliftonGuardian: We could merge the `if (isDesktop)` blocks in each region

bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
```diff
@@ -786,6 +829,17 @@ describe('getTeadsParams', () => {
 				});
 			},
 		);
+		test('should return correct pageD and placementID for MPU sized inline1 slots, in Uk when it is on desktop', () => {
```
DanielCliftonGuardian:
```suggestion
		test('should return correct pageId and placementId for MPU sized inline1 slots, in Uk when it is on desktop', () => {
```

Nitpick: Should it be pageId and placementId

bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
```diff
@@ -880,6 +945,75 @@ describe('getTeadsParams', () => {
 				});
 			},
 		);
+		test('should return correct pageD and placementID for MPU sized inline1 slots, in RoW when it is on desktop', () => {
+			isInUsa.mockReturnValue(true);
```
DanielCliftonGuardian:
```suggestion
		test('should return correct pageD and placementID for MPU sized inline1 slots, in US when it is on desktop', () => {
			isInUsa.mockReturnValue(true);
```
Should the description be US or the assertion ROW ?

DanielCliftonGuardian (commented):
Looks great. Just a couple of suggestions/questions

bundle/src/lib/header-bidding/prebid/bidders/config.ts
dskamiotis: Thanks, good point. I have merged the duplicated isDesktop checks in each region to remove repetition. I kept the overall shape as one branch per region and breakpoint because it is easier to read and stays consistent with the style used across this module.

bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
```diff
@@ -503,6 +503,49 @@ describe('bids', () => {
...
+			mockConsentState,
+		).find((bid) => bid.bidder === 'teads');
+
+		expect(teadsBid?.params).toEqual({
+			pageId: 265029,
+			placementId: 248133,
+		});
+	});
+
+	test('should pass inline2 slotId through to Ozone params for mobile MPU in US', () => {
```
DanielCliftonGuardian, line 528: Why do we need to test this here ?

bundle/src/lib/header-bidding/utils.ts
```diff
@@ -225,7 +221,6 @@ export const shouldIncludeBidder =
 				);
 			case 'teads':
 				return (
-					isInTeadsTest &&
 					isSwitchedOn('prebidTeads') &&
 					getConsentFor('teads', consentState) &&
 					(isInUk() || isInUsa() || isInRow())
```
DanielCliftonGuardian:
```suggestion
					(isInUk() || isInUsa() || isInRow() || isInAuOrNz())
```

DanielCliftonGuardian: Should we be adding Au or Nz here to the allowed bidders and updating the tests ?

bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
dskamiotis: Ah yeah, that was mainly to add a more concrete integration-path test in the bids suite, rather than only testing the helpers directly. Teads inline1 is the real change, and Ozone inline2 was added as a related slot-propagation example.

DanielCliftonGuardian (approved):
Looks great

domlander (approved):

PR #2595, opened 2026-06-26
Update value of exported constant `TOP_ABOVE_NAV_HEIGHT`

domlander:
## What does this change?

Updates the value of the constant `TOP_ABOVE_NAV_HEIGHT` from 90px to 250px

## Why?

The initial height of the top-above-nav ad slot is now 250px (plus ad label and padding), so this value was out of date.

This value is currently unused by DCAR, so this doesn't actually change anything. After this is merged, we can use this constant within DCAR to control the initial height of the ad space within the top-above-nav slot.

## Alternatives?

We remove this constant and delete the file, letting DCAR control the initial height of the top-above-nav ad slot. This would make it harder to find the useful comment about the AB testing, but we could add this to handbook/other documentation for visibility.

### DIFF
--- a/core/src/constants/top-above-nav-height.ts
+++ b/core/src/constants/top-above-nav-height.ts
@@ -1,17 +1,36 @@
 /**
  * Unit: pixels.
  *
- * The majority of ads in the top banner are 250px high. We ran an experiment
- * in October 2021 to set the minimum height to 250, and let smaller ads be
- * centred in the space. We did not process with this option, as it had a
- * negative impact on viewability and revenue.
+ * The initial height on page load of the top-above-nav ad slot. This excludes
+ * the height of the ad label and any padding around the slot.
  *
  */
-export const TOP_ABOVE_NAV_HEIGHT = 90;
+export const TOP_ABOVE_NAV_HEIGHT = 250;
 
 /*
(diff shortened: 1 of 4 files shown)


oliverabrahams (approved):
This seems like a good place for this information so think a good idea to keep it!

core/src/constants/top-above-nav-height.ts
```diff
@@ -1,18 +1,37 @@
...
+- Test: https://github.com/guardian/dotcom-rendering/pull/14198
+- Rollout: https://github.com/guardian/dotcom-rendering/pull/14510
+
+
+AB Test October 2021
+
+We ran an experiment in October 2021 to set the minimum height to 250, and let
+smaller ads be centred in the space. We did not process with this option, as it
+had a negative impact on viewability and revenue.
+
```
dskamiotis, line 34: This is really useful thanks @domlander 👍 We are definitely trying to add more JSDocs style commentary/annotation to our codebase

dskamiotis (approved):

cemms1 (approved):
Doesn't look like we actually use this value anywhere but perhaps we should. Thanks for updating the docs! ✨

PR #2605, opened 2026-06-30
Teads: Update Prebid size mapping

DanielCliftonGuardian:
## What does this change?

Adds `1x1` `outOfPage` as an eligible size in the header bidding size mapping for in-article placements:

- **Desktop:** `inline1`
- **Mobile:** `top-above-nav` and `inline1-x`

Tablet and non-article pages are unchanged.

## Why?

Teads requested that `1x1` be added as an eligible size to all in-article placements so they can serve outstream video and viewable display demand. Standard banner ads continue to return standard IAB sizes.

We can't restrict `1x1` to the Teads bidder within Prebid (bidders and sizes are configured independently, so any bidder can request any available size). To keep `1x1`/outstream demand contained to Teads, this will be isolated on the GAM side via a dedicated order/line item targeted to `hb_size=1x1` AND `hb_bidder=teads`.

> ⚠️ **Do not merge until the GAM line item is in place.** Releasing before then risks `1x1` demand leaking to other bidders. Pending Jasper's GAM-side change.

---
  - https://app.asana.com/0/0/1216043588481724


domlander:
I notice that we have a function that closes up 1x1 "out of page" ads. I'm wondering if this will get called for these slots and these adverts hidden? 

https://github.com/guardian/commercial/blob/63a570c52aeaa2c73baf80d388d6030ce9a6e7a7/bundle/src/events/render-advert.ts#L104

Do you know if the `1x1` size is exclusively used for Teads? 

(UPDATED: fixed link)

DanielCliftonGuardian:
> I notice that we have a function that closes up 1x1 "out of page" ads. I'm wondering if this will get called for these slots and these adverts hidden?
> 
> https://github.com/guardian/commercial/blob/doml/ozone-outstream/bundle/src/events/render-advert.ts#L105
> 
> Do you know if the `1x1` size is exclusively used for Teads?

It looks like this is mutually exclusive with the ozone AB test in our size mapping.

The `1x1` slots are not teads exclusives that is why we have isolated to Teads in GAM we've set up line items scoped with hb_bidder IS teads so only Teads win can render.

domlander:
I included a faulty link in my comment above, so I've updated it. Is it possible to test whether this `outOfPageCallback` gets called if a 1x1 ad gets served? It looks like the slot will be closed up and hidden if so. It is possible to test this? I'm not sure how I would do so.

We currently only serve exotic (non-mpu) ads in the first three inline slots on mobile. This PR will enable outstream video ads in every inline slot on mobile. I wonder if we should use an AB test to measure the effect that this will have?

bundle/src/lib/header-bidding/slot-config.ts
```diff
@@ -152,15 +152,18 @@ const getSlotSizeMapping = (): HeaderBiddingSizeMapping => {
 					]
 				: [getAdSize('mpu')],
 			tablet: [getAdSize('mpu')],
-			mobile: [getAdSize('mpu')],
```
domlander, line 155: Do we want Teads to be added to Fronts as well as articles? The `fronts-banner` ad slot is used exclusively on Fronts AFAIK

domlander:
UPDATE: We've had this confirmed that this should remain in the `banner` media type.

I'm wondering if the Teads 1x1 size needs to be in the `video` media type instead of `banner`. I think outstream video sizes should go under `video`, but I'm not sure on this. I also don't know whether 1x1 in this context would count as "outstream". I've taken a look at Teads documentation but I couldn't find anything concrete on this.

https://github.com/guardian/commercial/blob/63a570c52aeaa2c73baf80d388d6030ce9a6e7a7/bundle/src/lib/header-bidding/prebid/prebid-ad-unit.ts#L48-L62

bundle/src/lib/header-bidding/slot-config.ts
DanielCliftonGuardian: In the ticket it looks like we only want in articles per Jasper "Teads requested that we add 1x1 as an eligible size to all in-article placements (ad slots embedded within article paragraphs)"

domlander:
> I notice that we have a function that closes up 1x1 "out of page" ads. I'm wondering if this will get called for these slots and these adverts hidden?

This function looks like it doesn't actually close up the ad slots?! Perhaps a misleading comment describing past behaviour of this function. It does close up two ad slots, but neither will be affected by this PR, so I think we can ignore this function for the purposes of this PR. A future PR should address the ambiguity of this `outOfPage` callback function.

domlander (approved):

PR #2606, opened 2026-06-30
expose intent iq to all users

dskamiotis:
## What does this change?
- Removes IntentIQ experiment gating from both modules:
  - `commercial-user-module-intentIq`
  - `commercial-user-module-intentIq-us`
- Keeps IntentIQ enabled only for allowlisted regions (EU + non-EU, including US).
- Updates existing IntentIQ unit tests to match the new eligibility logic.
- Adds a new unit test suite for the IntentIQ analytics config module.

## Why?
The IntentIQ integration has been validated end-to-end with IntentIQ.
We now want to make IntentIQ available to all users in supported regions, without AB-test gating.
- Users outside the allowlist remain excluded.

### DIFF
--- a/bundle/src/lib/header-bidding/prebid/id-handlers/intent-iq.spec.ts
+++ b/bundle/src/lib/header-bidding/prebid/id-handlers/intent-iq.spec.ts
@@ -1,11 +1,6 @@
-import { isInUsa } from '@guardian/commercial-core/geo/geo-utils';
 import { getLocale } from '@guardian/commercial-core/geo/get-locale';
-import { isUserInTestGroup } from '../../../../ab-testing';
 import { getUserIdForIntentIQ } from './intent-iq';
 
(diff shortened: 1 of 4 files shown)


bundle/src/lib/header-bidding/prebid/id-handlers/intent-iq.spec.ts
DanielCliftonGuardian:
```suggestion
```
Looks like we can remove this now ?

DanielCliftonGuardian (approved):
Nice work, have left one minor comment

bundle/src/lib/header-bidding/prebid/id-handlers/intent-iq.ts
domlander: We don't need to export `intentIQEURegions` any more.

domlander (approved):
Thanks for adding tests

PR #2608, opened 2026-07-01
Use ad label constant instead of redefining

domlander:
## What does this change?

Use the constant AD_LABEL_HEIGHT for the height of the ad label, instead of redefining this value.

## Why?

If we were to update the height of the ad label in the future, we can be confident that every part of the code that uses the height of the ad label is updated.

### DIFF
--- a/.changeset/easy-toys-read.md
+++ b/.changeset/easy-toys-read.md
@@ -1,5 +1,0 @@
----
-'@guardian/commercial-core': patch
----
-
-Use ad label constant everywhere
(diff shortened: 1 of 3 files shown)


domlander:
I think a changeset is needed for this as it concerns a change in /bundle, but please let me know if I'm wrong.

deedeeh (approved):
Thanks for the refactor @domlander!

DanielCliftonGuardian (approved):
👏🏻

domlander:
Removed changeset. We only use changesets for core, not bundle.

PR #2609, opened 2026-07-01
Remove unused Prebid config: timeoutBuffer

domlander:
## What does this change?

Removes Prebid config: `timeoutBuffer`

## Why?

This is no longer used by Prebid:
- https://github.com/prebid/Prebid.js/issues/11959
- https://github.com/prebid/Prebid.js/pull/11960

I spent some time looking into how we might update this value when tweaking Prebid timeouts for optimisation purposes. It turns out Prebid no longer use this value, so I'd like to remove it to stop other devs going down this path!


deedeeh (approved):
It's much better to remove unused code @domlander 👍🏼

DanielCliftonGuardian (approved):

PR #2610, opened 2026-07-01
update global ad event adr doc

dskamiotis:
## What does this change?
Updates the `003-global-advert-event-listeners.md` highlighting the use of the `Mutation Observer` and the `onAdvert` window
## Why?

This offers a more immediate and robust solution for the "interactive" team to use, with a clearer pathway to a long term robust solution to deal with possible data attribute name changes.

### DIFF
--- a/docs/architecture/proposals/003-global-advert-event-listeners.md
+++ b/docs/architecture/proposals/003-global-advert-event-listeners.md
@@ -10,6 +10,8 @@ At the moment there is no supported API for determining the current lifecycle st
 
 Instead, some consumers inspect CSS classes or attributes that are added to the advert's DOM element. For example, the top-above-nav advert receives an `top-above-nav-ad-rendered` class once it has rendered.
 
+All ad slots when rendered have a class added `ad-slot--rendered` which can also be used currently to determine the rendered state.
+
 If the attribute is not yet present, the only option is to observe the DOM using a `MutationObserver` until it appears.
 
 This approach has a number of drawbacks:
--- a/docs/architecture/proposals/003-global-advert-event-listeners.md
+++ b/docs/architecture/proposals/003-global-advert-event-listeners.md
@@ -27,7 +27,7 @@ This approach has a number of drawbacks:
 
 Global advert events allow consumers to react to lifecycle changes, but they do not communicate the current status of an advert.
 
-If a listener is registered after an event has already been fired, there is no way to determine the advert's current lifecycle position without inspecting the DOM.
+If a listener is registered after an event has already been fired, there is no way to determine the advert's current lifecycle position.
 
 One possible approach would be to expose the current status of every advert alongside the lifecycle events.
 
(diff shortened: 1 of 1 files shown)


docs/architecture/proposals/003-global-advert-event-listeners.md
```diff
@@ -4,10 +4,57 @@
 
 ## Context
 
-There is currently no standard way for other code on the page (outside the commercial bundle) to listen for advert lifecycle events, for example to run code when an advert has rendered.
+### Current approach
+
+At the moment there is no supported API for determining the current lifecycle status of an advert.
+
+Instead, some consumers inspect CSS classes or attributes that are added to the advert's DOM element. For example, the top-above-nav advert receives an `ad-slot--rendered` class once it has rendered.
```
oliverabrahams: It looks like this class is added to all ads when they are rendered. Which is good. The specific top nav one is the attr `top-above-nav-ad-rendered`

dskamiotis: great spot!

docs/architecture/proposals/003-global-advert-event-listeners.md
```diff
@@ -4,11 +4,54 @@
...
+- It is not a documented or supported interface.
 
 ## Proposal
 
-Export a function that allows other code to register event listeners for Advert lifecycle events.
+### The gap
+
+Global advert events allow consumers to react to lifecycle changes, but they do not communicate the current status of an advert.
+
+If a listener is registered after an event has already been fired, there is no way to determine the advert's current lifecycle position without inspecting the DOM.
```
oliverabrahams:
```suggestion
If a listener is registered after an event has already been fired, there is no way to determine the advert's current lifecycle position.
```

Not sure if you could really do this by inspecting the DOM

docs/architecture/proposals/003-global-advert-event-listeners.md
```diff
@@ -4,11 +4,54 @@
 
 ## Context
 
-There is currently no standard way for other code on the page (outside the commercial bundle) to listen for advert lifecycle events, for example to run code when an advert has rendered.
+### Current solution:
+
+At the moment there is no supported API for determining the current lifecycle status of an advert.
+
+Instead, some consumers inspect CSS classes or attributes that are added to the advert's DOM element. For example, the top-above-nav advert receives an `top-above-nav-ad-rendered` class once it has rendered.
```
oliverabrahams, line 11:
```suggestion
Instead, some consumers inspect CSS classes or attributes that are added to the advert's DOM element. For example, the top-above-nav advert receives an `top-above-nav-ad-rendered` class once it has rendered.

All ad slots when rendered have a class added `ad-slot--rendered` which can also be used currently to determine the rendered state.
```

oliverabrahams (approved):
Looks good just a few suggestions

PR #2611, opened 2026-07-06
Implement new Ozone Outstream prebid size behind AB test

domlander:
AB test PR: https://github.com/guardian/dotcom-rendering/pull/16330

## What does this change?

- Introduces new Ozone Outstream video ad size (640x360)
- Adds `video` mediaType in Prebid requests for inline1 slot across all devices for outstream video sizes
- Uses outstream sizes as playerSize for this new media type

Encapsulates the new code in an AB test, so that this change can be tested before going live.

## Why?

We would like to improve our Ozone Outstream integration.

Tested locally and on CODE

## How to test?

- Go to an article with the AB test suffix appended to the URL: ?ab-commercial-ozone-outstream=variant
- Subscribe to the Commercial logger and check the PrebidAdUnit. There should be two ozone bidders: one of which should use the new inline1 placementId 1500001169. This is the video mediaType.
- Check that this is not true when not in the AB test.

### DIFF
--- a/.changeset/hip-ads-change.md
+++ b/.changeset/hip-ads-change.md
@@ -1,0 +1,5 @@
+---
+'@guardian/commercial-core': minor
+---
+
+Adds Outstream Ozone ad size
--- a/bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
+++ b/bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
@@ -1022,12 +1022,24 @@ describe('getOzonePlacementId', () => {
 		jest.resetAllMocks();
 	});
 
-	test('should return correct placementID for inline1 slot', () => {
+	test('should return correct placementID for inline1 slot when in AB test', () => {
 		isInUsa.mockReturnValue(true);
 		getBreakpointKey.mockReturnValue('M');
 		containsMpu.mockReturnValue(true);
 		containsMobileSticky.mockReturnValue(false);
+		jest.mocked(isUserInTestGroup).mockReturnValueOnce(true);
 		expect(getOzonePlacementId([[300, 250]], 'dfp-ad--inline1')).toBe(
+			'1500001169',
+		);
+	});
+
+	test('should NOT return inline1 placementID for inline1 slot when NOT in AB test', () => {
+		isInUsa.mockReturnValue(true);
+		getBreakpointKey.mockReturnValue('M');
+		containsMpu.mockReturnValue(true);
+		containsMobileSticky.mockReturnValue(false);
+		jest.mocked(isUserInTestGroup).mockReturnValueOnce(false);
+		expect(getOzonePlacementId([[300, 250]], 'dfp-ad--inline1')).not.toBe(
 			'1500001169',
 		);
 	});
--- a/bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
+++ b/bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
@@ -1023,7 +1023,6 @@ describe('getOzonePlacementId', () => {
 	});
 
 	test('should return correct placementID for inline1 slot when in AB test', () => {
-		isInUsa.mockReturnValue(true);
 		getBreakpointKey.mockReturnValue('M');
 		containsMpu.mockReturnValue(true);
 		containsMobileSticky.mockReturnValue(false);
@@ -1034,7 +1033,6 @@ describe('getOzonePlacementId', () => {
 	});
 
 	test('should NOT return inline1 placementID for inline1 slot when NOT in AB test', () => {
-		isInUsa.mockReturnValue(true);
 		getBreakpointKey.mockReturnValue('M');
 		containsMpu.mockReturnValue(true);
 		containsMobileSticky.mockReturnValue(false);
--- a/bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
+++ b/bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
@@ -1022,6 +1022,17 @@ describe('getOzonePlacementId', () => {
 		jest.resetAllMocks();
 	});
 
+
+	test('should return correct placementID for inline1 slot', () => {
+		isInUsa.mockReturnValue(true);
+		getBreakpointKey.mockReturnValue('M');
+		containsMpu.mockReturnValue(true);
+		containsMobileSticky.mockReturnValue(false);
+		expect(getOzonePlacementId([[300, 250]], 'dfp-ad--inline1')).toBe(
+			'1500001169',
+		);
+	});
+
 	test('should return correct placementID for billboard in US when it is in desktop', () => {
 		isInUsa.mockReturnValue(true);
 		getBreakpointKey.mockReturnValue('D');
@@ -1098,7 +1109,7 @@ describe('getOzonePlacementId', () => {
 		getBreakpointKey.mockReturnValue('M');
 		containsMpu.mockReturnValue(true);
 		containsMobileSticky.mockReturnValue(false);
-		expect(getOzonePlacementId([[300, 250]], 'dfp-ad--inline1')).toBe(
+		expect(getOzonePlacementId([[300, 250]], 'dfp-ad--inline3')).toBe(
 			'1500001036',
 		);
 	});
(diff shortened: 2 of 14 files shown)


domlander:
Is it worth creating a separate PR for the `/core` changes? This is where the new ad size is added.

bundle/src/lib/header-bidding/prebid/prebid-ad-unit.ts
```diff
@@ -30,8 +49,29 @@ export class PrebidAdUnit implements AdUnitDefinition {
...
+			...(isInOzoneAbTest && slot.key === 'inline1'
+				? {
+						video: {
+							playerSize: filterSizesForVideoMediaType(
+								slot.sizes,
+							),
+							context: 'outstream',
+							placement: 3, // in-article
+							plcmt: 4, // outstream
+						} as VideoMediaType,
```
cemms1, line 59: _non blocking/optional_
Is it possible to do this with a type guard rather than a type assertion?

bundle/src/lib/header-bidding/prebid/bidders/config.ts
```diff
@@ -374,13 +375,22 @@ const getOzonePlacementId = (
...
+	);
+
+	if (isInOzoneAbTest && slotId === 'dfp-ad--inline1') {
+		return '1500001169';
+	}
+
 	if (isInUsa()) {
 		if (getBreakpointKey() === 'D') {
-			if (containsBillboard(sizes)) {
+			if (containsLeaderboardOrBillboard(sizes)) {
```
domlander, line 387: AFAIK, this shouldn't actually change anything, since we currently always have a billboard size when there is a leaderboard size at desktop. However, I've updated this to more accurately reflect Ozone's request, which would be helpful for us if we didn't show a billboard for any reason.

Same for below with MPU & DMPU

.changeset/hip-ads-change.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/commercial-core': minor
+---
```
dskamiotis, line 3: Just checking if we need a changeset for this work? Since this is part of the bundle work?

bundle/src/lib/header-bidding/prebid/bidders/config.ts
dskamiotis: Agreed I think this is really helpful for understanding and debugging 👍

bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
```diff
@@ -1022,17 +1022,39 @@ describe('getOzonePlacementId', () => {
...
+		getBreakpointKey.mockReturnValue('M');
+		containsMpu.mockReturnValue(true);
+		containsMobileSticky.mockReturnValue(false);
+		jest.mocked(isUserInTestGroup).mockReturnValueOnce(true);
+		expect(getOzonePlacementId([[300, 250]], 'dfp-ad--inline1')).toBe(
+			'1500001169',
+		);
+	});
+
+	test('should NOT return inline1 placementID for inline1 slot when NOT in AB test', () => {
```
dskamiotis: Maybe a mention this test is for the US region might be helpful?

bundle/src/lib/header-bidding/prebid/prebid-ad-unit.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+		);
+
+		expect(adUnit.mediaTypes).toEqual({
+			banner: {
+				sizes: slot.sizes,
+			},
+		});
+		expect(adUnit.bids).toEqual([buildDummyBid()]);
+		expect(mockedBids).toHaveBeenCalledWith(
+			advert.id,
```
dskamiotis, line 77: I like this approach to testing—it's really clear what's happening at each step.

bundle/src/lib/header-bidding/prebid/bidders/config.ts
```diff
@@ -374,13 +375,22 @@ const getOzonePlacementId = (
...
 	pageTargeting?: PageTargeting,
 ) => {
+	const isInOzoneAbTest = isUserInTestGroup(
+		'commercial-ozone-outstream',
+		'variant',
+	);
+
+	if (isInOzoneAbTest && slotId === 'dfp-ad--inline1') {
+		return '1500001169';
+	}
```
dskamiotis, line 383: Assuming we want to return the same placement ID for all regions/breakpoints? Just checking as the existing logic below differentiates by region and breakpoint.

_nitpick_: maybe a short comment to differentiate that its an outstream placement ID?

dskamiotis (approved):

dskamiotis (approved):
Nice work 👍  a couple of clarifying questions, but otherwise looks good to me

.changeset/hip-ads-change.md
domlander: There's one change to core: adding a new size. I wonder if this should be split out into a new PR with a changeset, then the changeset attached to this PR removed?

cemms1: I think keeping as the same PR makes more sense

bundle/src/lib/header-bidding/prebid/bidders/config.spec.ts
domlander: This test checks that the inline1 placement ID is not used if the slot isn't inline1, irrespective of region. I'll remove the mock return value for the US to make it clearer and we don't need it.

bundle/src/lib/header-bidding/prebid/bidders/config.ts
domlander: Yes that's correct. We want to use this placement ID for all regions and breakpoints.

We'd like to use this placement ID for outstream only, but AFAICT there's no way to link a placement ID to a media type, as we send `mediaTypes` and `bids.params.placementId` separately within `adUnits`. Therefore, the placement ID has to be associated with the slot/sizes instead.

I'll add a comment to summarise this.

.changeset/hip-ads-change.md
dskamiotis: > There's one change to core: adding a new size. I wonder if this should be split out into a new PR with a changeset, then the changeset attached to this PR removed?

That makes perfect sense just spotted - thanks 👍

bundle/src/lib/header-bidding/prebid/prebid-ad-unit.ts
domlander: The `VideoMediaType` type requires every property listed [here](https://github.com/InteractiveAdvertisingBureau/openrtb2.x/blob/main/2.6.md#327---object-video-) to be defined, so I couldn't think of a good way to do this. I could create a default object with all of these properties defined, then extend it. If you have an idea, please let me know.

bundle/src/lib/header-bidding/prebid/bidders/config.ts
```diff
@@ -442,35 +436,59 @@ const teadsBidder: PrebidBidder = {
...
-						// Assigns a random integer between 0 and 99
-						...testgroup,
-						...buildAppNexusTargetingObject(pageTargeting),
+	const bidders: Array<'banner' | 'video'> = ['banner'];
+
+	const isVideoSlotSizes =
+		slotSizes.filter((size) => isOutstream(size)).length > 0;
+	if (isInOzoneAbTest && isVideoSlotSizes) {
+		bidders.push('video');
+	}
```
cemms1: I find initialising this as `['banner']` and then mutating it to add `'video'` under certain conditions a little hard to follow

Perhaps we could do something like
```suggestion
	const isVideoSlotSizes =
		slotSizes.some(isOutstream);
	
	const bidderTypes = (isInOzoneAbTest && isVideoSlotSizes) ? ['banner', 'video'] : ['banner']
```

cemms1: Alternatively we could have two distinct bidders for Ozone: `ozoneVideo` and `ozoneBanner` which might help with readability

domlander: I like both of these ideas, I think I prefer this latter one. I think it should be more readable now.

cemms1 (approved):
Looks great 👏

PR #2622, opened 2026-07-14
Add hangtime ad format definition

DanielCliftonGuardian:
## What does this change?
Documents "hangtime" ad format.
## Why?
We shouldn't have undocumented, unclear terms in the codebase.

---
  - https://app.asana.com/0/0/1216355421860747


deedeeh (approved):
Thank you for making it much clearer for everyone else checking the codebase @DanielCliftonGuardian ✨

PR #2627, opened 2026-07-16
Remove the second ozone bidder. Use one placementId for inline1 everywhere

domlander:
## What does this change?

- Use the same Placement ID for all inline1 slots.
- Only use 1 Ozone bidder.

## Why?

Request from Ozone. The initial understanding was that Ozone wanted a separate Placement ID for the video Media Type. Since bids cannot be linked to media types, we created a separate bidder for Ozone with a different Placement ID. However, Ozone now only want one Placement ID per slot, for performance reasons, we remove the second Ozone bidder.


dskamiotis (approved):
This looks good to me Dominic 👍

PR #2628, opened 2026-07-16
remove prebid-price-floor-holdback test gate

dskamiotis:
## What does this change?
Removes the A/B test gate for `commercial-prebid-price-floor-holdback`
## Why?
This test gate is now redundant since we have completed the holdback experiment and are applying the price floors are now enabled for 100% of users.

Related Test config PR in DCR: https://github.com/guardian/dotcom-rendering/pull/16399


cemms1:
@dskamiotis I thought we still wanted this for price floors v2?

PR #2641, opened 2026-07-27
Remove the outstream ozone AB test

domlander:
## What does this change?

Removes the Ozone outstream AB test. The new outstream size has been verified and will replace the two old outstream sizes. 

The Teads 1x1 size `outOfPage` is retained in slots in which it was used outside the variant of this AB test.

## Why?

The testing of this new size is complete. An AB test is not required to gradually increase the number of users exposed to this new size.

The Teads size was applied to all users before this PR and will be following its removal. There is no change here.

### DIFF
--- a/bundle/src/init/consented/static-ad-slots.ts
+++ b/bundle/src/init/consented/static-ad-slots.ts
@@ -2,7 +2,6 @@ import type { SizeMapping } from '@guardian/commercial-core/ad-sizes';
 import { adSizes, createAdSize } from '@guardian/commercial-core/ad-sizes';
 import { isInUsa } from '@guardian/commercial-core/geo/geo-utils';
 import { isNonNullable, log } from '@guardian/libs';
-import { isUserInTestGroup } from '../../ab-testing';
 import { createAdvert } from '../../define/create-advert';
 import { displayAds } from '../../display/display-ads';
 import { displayLazyAds } from '../../display/display-lazy-ads';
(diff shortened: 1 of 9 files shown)


bundle/src/init/consented/static-ad-slots.ts
```diff
@@ -23,27 +22,23 @@ const decideAdditionalSizes = (adSlot: HTMLElement): SizeMapping => {
...
-				adSizes.outstreamGoogleDesktop,
-			],
-			desktop: [
-				isInOzoneAbTest
-					? adSizes.outstreamOzone
-					: adSizes.outstreamDesktop,
-				adSizes.outstreamGoogleDesktop,
-			],
-		};
+		return name === 'inline1'
```
domlander, line 26: Restrict the new outstream size to inline1 only. We only want Ozone to bid on inline1 slots. This was missed in the 0% AB test implementation

bundle/src/insert/spacefinder/article.ts
```diff
@@ -222,27 +210,21 @@ const addDesktopRightRailAds = ({
...
+		case 2: // inline2
+			return {
+				mobile: [
+					adSizes.portraitInterstitial,
+					adSizes.pubmaticInterscroller,
+				],
+			};
+		default:
+			return undefined;
 	}
```
dskamiotis, line 227: nice clear approach 👍

dskamiotis (approved):
Great 🚢  Looks good, this sets up the next bit of work nicely.

GHaberis (commented):
It looks like by rolling out the changes in `isInOzoneAbTest` the PR is replacing references to the ad size `outstreamDesktop` with `outstreamOzone`. 

I'm wondering is the ad size `outstreamDesktop` still required anywhere? I can see references that remain here:
 https://github.com/guardian/commercial/blob/ffe79971e6a1fea55ce0ae7694f04b660f9341a1/bundle/src/events/render-advert.ts#L80 

And here: 

https://github.com/guardian/commercial/blob/ffe79971e6a1fea55ce0ae7694f04b660f9341a1/bundle/src/lib/header-bidding/prebid/price-config.ts#L72

Have they intentionally be left or do we need to change all references to `outstreamDesktop`?

domlander:
> Have they intentionally be left or do we need to change all references to `outstreamDesktop`?

These have been intentionally left for the cleanup task as I don't think leaving them in has any effect. There will be a PR that @dskamiotis is working on that will remove the sizes `outstreamDesktop` and `outstreamMobile` and will include the above two references.

GHaberis (commented):
Nice work 👍

PR #2645, opened 2026-07-29
Add mute button to interscroller video

oliverabrahams:
## What does this change?

Adds a mute button to the interscroller when is has the property 'hasAudio' set to 'true'

## Why?
Some interscroller videos have audio. We want to be able to unmute them.

Screenshot 2026-08-05 at 14 17 25

---
  - https://app.asana.com/0/0/1217123654565991


bundle/src/lib/messenger/background.ts
```diff
@@ -184,6 +185,77 @@ const setupParallax = (
 	observer.observe(backgroundParent);
 };
 
+// from https://github.com/guardian/csnx/blob/main/libs/%40guardian/source/src/react-components/__generated__/icons/SvgAudio.tsx
```
cemms1: Should we use a permalink here in case the path changes?

Assume the reason we can't import directly is because it's a React component?

cemms1 (approved):
LGTM

bundle/src/lib/messenger/background.ts
oliverabrahams: Good thinking. Will update this. 

Yea its a React component

bundle/src/lib/messenger/background.ts
```diff
@@ -184,6 +185,77 @@ const setupParallax = (
...
+
+// from https://github.com/guardian/csnx/blob/main/libs/%40guardian/source/src/react-components/__generated__/icons/SvgAudioMute.tsx
+const mutedIcon = `
+<svg
+		width="30"
+		height="30"
+		viewBox="-3 -3 30 30"
+		xmlns="http://www.w3.org/2000/svg"
+		focusable={false}
+		aria-hidden={true}
```
DanielCliftonGuardian:
```suggestion
		focusable="false"
		aria-hidden="true"
```

Should we change this to standard html ?

bundle/src/lib/messenger/background.ts
```diff
@@ -184,6 +185,77 @@ const setupParallax = (
...
 
+// from https://github.com/guardian/csnx/blob/main/libs/%40guardian/source/src/react-components/__generated__/icons/SvgAudio.tsx
+const unmutedIcon = `
+<svg
+		width="30"
+		height="30"
+		viewBox="-3 -3 30 30"
+		xmlns="http://www.w3.org/2000/svg"
+		focusable={false}
+		aria-hidden={true}
```
DanielCliftonGuardian:
```suggestion
		focusable="false"
		aria-hidden="true"
```

bundle/src/lib/messenger/background.ts
oliverabrahams: 🤦  yes oops

DanielCliftonGuardian (approved):

PR #2652, opened 2026-08-04
Reinstate old Ozone sizes

domlander:
## What does this change?

Reinstates the usage of the following outstream sizes, where they were removed in [this PR](https://github.com/guardian/commercial/pull/2641):
- 620x350
- 300x197

## Why?

It was thought that these sizes were no longer used. However, they are still used by some vendors.

### DIFF
--- a/bundle/src/lib/messenger/passback.ts
+++ b/bundle/src/lib/messenger/passback.ts
@@ -26,19 +26,33 @@ type PassbackMessagePayload = { source: string };
  */
 const mpu: [number, number] = [adSizes.mpu.width, adSizes.mpu.height];
 
+const outstreamDesktop: [number, number] = [
+	adSizes.outstreamDesktop.width,
+	adSizes.outstreamDesktop.height,
+];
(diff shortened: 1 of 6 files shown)


bundle/src/lib/messenger/passback.ts
```diff
@@ -26,19 +26,33 @@ type PassbackMessagePayload = { source: string };
...
+	adSizes.outstreamMobile.height,
+];
 const outstreamOzone: [number, number] = [
 	adSizes.outstreamOzone.width,
 	adSizes.outstreamOzone.height,
 ];
 
+const outstreamSizes = [outstreamDesktop, outstreamMobile, outstreamOzone];
+
+const outstreamHeightDesktop = Math.max(
```
domlander, line 44: Set the height of the passback slot to the height of the largest outstream size we have, to minimise CLS. This slot expands and contracts when a passback occurs before it sets this height, which is not great, but this quirk existed before this change so will be tackled in a future PR

DanielCliftonGuardian (approved):

PR #2653, opened 2026-08-04
Enable Prebid transaction IDs behind A/B test

i-hardy:
## What does this change?

This enables Prebid transaction IDs for users in the variant of a dedicated test group.

## Why?

This follows a request from The Trade Desk for their OpenPath integration, but we are enabling TIDs globally behind an A/B test.

---
  - https://app.asana.com/0/0/1216920173565507


domlander (approved):

dskamiotis (approved):
This looks good to 👍

PR #2657, opened 2026-08-06
allows all non-UK Liveblog articles on mobile to show moble-sticky

dskamiotis:
## What does this change?
Renders mobile-sticky ads on all LiveBlog articles in eligible non-UK regions.

## Why?
After initial rollout of mobile-sticky to LiveBlog articles in US with positive test results, we have approval to expand the feature to all other non-UK regions.

## Testing

Tested locally: http://localhost:3030/Article/https://www.theguardian.com/sport/live/2026/aug/07/tour-de-france-femmes-2026-stage-seven-mont-ventoux-finish-live
VPN : France

https://github.com/user-attachments/assets/87074077-20bf-4c60-8bba-19f54b281865

### DIFF
--- a/bundle/src/lib/header-bidding/utils.ts
+++ b/bundle/src/lib/header-bidding/utils.ts
@@ -250,8 +250,7 @@ export const shouldIncludeMobileSticky = once(
 		}) &&
 			!isInUk() &&
 			(isValidPageForMobileSticky() ||
-				(window.guardian.config.page.contentType === 'LiveBlog' &&
-					isInUsa())) &&
+				(window.guardian.config.page.contentType === 'LiveBlog')) &&
 			!window.guardian.config.page.isHosted),
 );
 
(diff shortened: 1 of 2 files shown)


bundle/src/lib/header-bidding/utils.ts
```diff
@@ -250,8 +250,7 @@ export const shouldIncludeMobileSticky = once(
 		}) &&
 			!isInUk() &&
 			(isValidPageForMobileSticky() ||
-				(window.guardian.config.page.contentType === 'LiveBlog' &&
-					isInUsa())) &&
+				(window.guardian.config.page.contentType === 'LiveBlog')) &&
```
cemms1: We could put this inside `isValidPageForMobileSticky`?

dskamiotis: Nice 👍

DanielCliftonGuardian (approved):
👌

PR #2639, opened 2026-07-24
Adds `devenv` to repo for devcontainer support

cemms1:
## What does this change?

Sets up [`devenv`](https://github.com/guardian/devenv) in this repo to allow us to work within `dev containers` when running on our local machines.

### What are dev containers?

See https://containers.dev

### What is `devenv`?

See https://github.com/guardian/devenv

### How do I test this branch?

- Check out this branch (`cemms1/add-devenv`)
- Install `devenv`. You can do this by running `mise install` in this repo on this branch or by following instructions in the [devenv repo](https://github.com/guardian/devenv)
- Optionally edit the `.devcontainer/devenv.yaml` file or add your own personal configuration at `~/.config/devenv/devenv.yaml`
- Run `devenv generate` to update the container configuration based on the project configuration + your personal configuration
- Open the repo in VSCode or IntelliJ, ensuring you open in a devcontainer when the option is available to you
- Try running commands you would normally run in this repo

### IDE set up instructions

- [VSCode](https://github.com/guardian/devenv/blob/4f0ff1bbb491f76e49ef4958e1ea5320abd8a04c/docs/containerised-development/containerised-development-in-vscode.md)

- [IntelliJ](https://github.com/guardian/devenv/blob/4f0ff1bbb491f76e49ef4958e1ea5320abd8a04c/docs/containerised-development/containerised-development-in-intellij.md)

## Why?

For safer development, we'd like to run scripts within a container on our local machine.

PR #2661, opened 2026-08-11
Add copilot-generated copilot instructions to repo

cemms1:
## What does this change?

Uses Github Copilot CLI to add `copilot-instructions.md` to the repo

## Why?

This could help Copilot to give more helpful suggestions and feedback in this repo, for PR reviews and in our IDEs.
It might also help Copilot agents to complete tasks autonomously now that we have dev container set up (see #2639)

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,88 @@
+# Copilot Instructions
+
+## Workspace and commands
+
+This is a pnpm 10 workspace (`core` and `bundle`); use the Node version in
+`.nvmrc`. Install dependencies with `pnpm install`.
+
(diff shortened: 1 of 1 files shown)


deedeeh (commented):
Is this required since we have copilot CLI enabled?

cemms1:
> Is this required since we have copilot CLI enabled?

Here are the docs for it https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/add-custom-instructions/add-repository-instructions

It's not required by any means, but I think it might help to give context to Copilot in this repo. We can always try it and remove if it's not helpful?

deedeeh (approved):

PR #2670, opened 2026-08-18
Bump Typescript to v6

cemms1:
## What does this change?

- Bumps version of `typescript` to `6.0.3` and `ts-jest` to `29.4.12` (we need at least `29.4.11` for these updates to work properly)
- Updates `tsconfig` in the root directory to add `types: ["node"]`
- Updates `tsconfig` in the `core` directory to use [`moduleResolution=bundler`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#combining---moduleresolution-bundler-with---module-commonjs) [instead of `node`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#deprecated---moduleresolution-node-aka---moduleresolution-node10)

## Why?

Keeping on top of dependencies & [readying ourselves for v7 of Typescript](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#preparing-for-typescript-70)


domlander:
Does this require a changeset?

cemms1:
> Does this require a changeset?

Oh good point! I'll add one

domlander (approved):
I had a look through the v6 release notes and I think you've got everything

PR #2671, opened 2026-08-18
Update Babel to v8

cemms1:
## What does this change?

Updates `bundle` package to use Babel v8:

- Bumps Node version to `22.18.0`
- Bumps `@babel*` and `babel*` packages to the latest available versions
- Converts the config file to a JSON instead of commonJS style module
- Removes unneeded plugin `@babel/plugin-syntax-dynamic-import` as this has been included in `@babel/preset-env` for a while now
- Updates `dependabot.yaml` config to group together Babel packages for updates

See upgrade docs: https://babeljs.io/docs/v8-migration

## Why?

Keeps on top of dependency updates.
We have a number of open Dependabot PRs to bump Babel to v8 but they should all be done together with some awareness of the major changes included in the version bump


domlander (approved):
The compressed-size-action bot reports that the compressed bundle size has gone down by 10%. Is this expected / is the bot reliable?

cemms1:
> The compressed-size-action bot reports that the compressed bundle size has gone down by 10%. Is this expected / is the bot reliable?

10% is more than I'd expect! According to [this blog post](https://babeljs.io/blog/2026/06/16/8.0.0), babel no longer compiles to ES5 and CommonJS by default - so that might have an impact?

domlander:
Ah that makes sense. I checked a few merged PRs and it looks like the bot produces expected numbers, so it looks like that 10% is real. Nice work!

PR #2684, opened 2026-08-18
Add a failsafe timeout to Prebid

domlander:
## What does this change?

Add a failsafe timeout to Prebid behind an AB test. If for some reason the `bidsBackHandler` is not called within three seconds, invoke this handler with the bids that have been returned.

## Why?

Prebid recommends having a failsafe timeout in case something goes wrong[^1]. This will ensure that Prebid doesn't wait indefinitely for bids and could lead to higher ad-ratio.

[^1]: https://docs.prebid.org/features/timeouts.html

### DIFF
--- a/.changeset/cold-shoes-play.md
+++ b/.changeset/cold-shoes-play.md
@@ -1,0 +1,5 @@
+---
+"@guardian/commercial-core": major
+---
+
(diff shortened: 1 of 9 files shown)


deedeeh (approved):
Looks good @domlander! I like that you noticed a small refactor like adding "consent" to the tests which would make it much clear.

PR #2691, opened 2026-08-25
Lazy load header-bidding refactor

domlander:
no-op

## What does this change?

Renames Prebid references in `lazy-load.ts` to refer to the more general "header bidder" or "header bidding"

## Why?

Both Prebid and APS auctions are run when lazy loading is enabled. The naming was a little confusing as it suggests only Prebid auctions are run eagerly, when in fact both header bidding auctions are run.

### DIFF
--- a/bundle/src/display/lazy-load.ts
+++ b/bundle/src/display/lazy-load.ts
@@ -39,7 +39,7 @@ const onIntersectDisplayAd = (
 	);
 };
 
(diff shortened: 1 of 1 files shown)


deedeeh (approved):
Much clearer @domlander!

DanielCliftonGuardian (approved):

PR #2693, opened 2026-08-26
Adds an extra test for Prebid timeouts

domlander:
no-op

## What does this change?

Adds a test to explain the relationship between the Prebid timeouts: auction and failsafe.

Updates the header bidding docs with timeout info.

### DIFF
--- a/core/src/constants/index.spec.ts
+++ b/core/src/constants/index.spec.ts
@@ -23,3 +23,16 @@ describe('Constant values are constant', () => {
 		expect(AD_LABEL_HEIGHT).toBe(24);
 	});
 });
+
+describe('Prebid timeout combination', () => {
+	/**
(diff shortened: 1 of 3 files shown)


core/src/constants/index.spec.ts
```diff
@@ -23,3 +23,16 @@ describe('Constant values are constant', () => {
 		expect(AD_LABEL_HEIGHT).toBe(24);
 	});
 });
+
+describe('Prebid timeout combination', () => {
+	/**
+	 * The failsafe timeout should always be much larger than the auction timeout. The failsafe timeout is a safety
+	 * net that invokes the callback in case something goes wrong with Prebid. If these two timeouts are too close
+	 * together, the failsafe timeout might often be invoked before the auction timeout has a chance to complete.
+	 */
```
cemms1, line 32: Nice comment!

cemms1 (approved):
👍

PR #2694, opened 2026-08-26
Add `bidtimeout` to Prebid analytics

domlander:
## What does this change?

Adds BID_TIMEOUT to the list of our Prebid event handlers.

## Why?

We would like to have reporting on the bidder timeout rates within Prebid. With this change, we no longer have to use a workaround to calculate bidder timeout rate.

### DIFF
--- a/bundle/src/lib/header-bidding/prebid/modules/analyticsAdapter/eventHandlers.ts
+++ b/bundle/src/lib/header-bidding/prebid/modules/analyticsAdapter/eventHandlers.ts
@@ -34,6 +34,7 @@ const GU_EVENTS = {
 	BID_RESPONSE: EVENTS.BID_RESPONSE,
 	BID_WON: EVENTS.BID_WON,
 	NO_BID: EVENTS.NO_BID,
+	BID_TIMEOUT: EVENTS.BID_TIMEOUT,
 } as const;
 
 interface GuEvents {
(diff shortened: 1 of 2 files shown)


bundle/src/lib/header-bidding/prebid/modules/analyticsAdapter/eventHandlers.ts
```diff
@@ -228,6 +238,21 @@ const eventHandlers: EventHandlers = {
...
+				n: bid.bidder,
+				bid: bid.bidId,
+				sid: bid.adUnitCode,
+				aid: bid.auctionId,
+				ttr: duration,
+			}),
+		);
+
+		return events;
+	},
```
cemms1, line 255: This seems like the time to bid rather than the _timeout_ since it registers when the bid happens after the auction start. Or have I got the wrong end of the stick here?

What is the definition of "bid timeout" here?

cemms1: Is it that "EVENTS.BID_TIMEOUT" is a specific event that will fire for each bid if no bid is made within the set time? In which case this code will only execute after that event has happened

cemms1: I think I've answered my own question here 🦆 

https://github.com/prebid/Prebid.js/blob/52be0c58d2eef9812bc1fa83221c633c7a132317/src/auction.ts#L100-L103

I'll approve ✅

cemms1 (approved):

bundle/src/lib/header-bidding/prebid/modules/analyticsAdapter/eventHandlers.ts
domlander: This is correct. We fire a bidtimeout event for each bidder that doesn't return a bid in time

image

PR #2695, opened 2026-08-26
Allow Playwright to run against DCR in a dev container

cemms1:
## What does this change?

Enables Playwright to be run within a dev container by using the Docker image already used by the end to end CI workflow when running Playwright locally

You can start DCR in a container by running `./scripts/start-dcr-container.sh` from the root level or `pnpm run start-dcr` from within the `bundle` directory.
It defaults to using the main branch but if you've recently pushed a commit to a branch in the dotcom-rendering repo, this commit should also be available to use. You can pass thwe commit hash into the command as the first argument, which should run the DCR build of that version. ie. `pnpm run start-dcr <commit-sha>`

> [!TIP]
> Make sure you regenerate (`devenv generate`) and rebuild your dev container when checking out this branch to ensure Playwright installs the chromium browser and dependencies in your environment

## Why?

When working within a dev container, Playwright wasn't working properly as it relies on DCR running separately on a different port. Since DCR was running _outside_ of the container, it wasn't accessible via a localhost port _within_ the container. 
The solution is to run DCR within the dev container as well. The simplest way to do this is to copy what the end-to-end CI workflow does and runs DCR inside a Docker container, within the dev container.

We need to be able to run Playwright locally so this PR addresses that and tries to make it as simple a process as possible


bundle/package.json
```diff
@@ -20,8 +20,9 @@
 		"build:dev": "webpack -c webpack.config.dev.mjs",
 		"build:prod": "webpack -c webpack.config.prod.mjs",
 		"clean": "rm -rf dist",
+		"start-dcr": "./scripts/start-dcr-container.sh",
 		"playwright:run": "playwright test",
-		"playwright:open": "playwright test --ui",
+		"playwright:open": "playwright test --ui --ui-host=0.0.0.0 --ui-port=4321",
```
cemms1, line 25: See https://playwright.dev/docs/test-ui-mode#docker--github-codespaces

domlander (approved):
🎉

PR #2702, opened 2026-08-27
Add missing switches to commercial code

cemms1:
## What does this change?

Adds missing switches, included in https://github.com/guardian/frontend/pull/29140, to commercial bundle to ensure we are able to quickly turn off third party services and integrations when needed

## Why?

A number of third party providers have been added to Commercial recently without having a feature switch behind them to enable easy turn on and off of features without code changes.
This PR rectifies that, adding some more logic to be able to turn off these integrations if needed

> [!WARNING]
> Do not merge this until the switches have been added by merging https://github.com/guardian/frontend/pull/29140 and waiting until this is ready in Production

### DIFF
--- a/bundle/src/init/consented/ipsos-mori.ts
+++ b/bundle/src/init/consented/ipsos-mori.ts
@@ -24,12 +24,16 @@ const loadIpsosScript = (locale: 'au' | 'uk') => {
  * @returns Promise
  */
 export const initIpsosMori = async (): Promise<void> => {
+	if (!window.guardian.config.switches.ipsosMori) {
+		log('commercial', 'Ipsos Mori switch is turned off');
+		return Promise.resolve();
+	}
 	const locale = getLocale();
 	const consentState = await onConsent();
 	const isAU = locale === 'AU' && !!consentState.aus;
 	const isUK = locale === 'GB' && !!consentState.tcfv2;
 
-	if (!isAU && !isUK && !window.guardian.config.switches.ipsosMori) {
+	if (!isAU && !isUK) {
 		log('commercial', 'Skipping ipsos process outside GB or AU');
 	}
 
--- a/bundle/src/init/consented/ipsos-mori.ts
+++ b/bundle/src/init/consented/ipsos-mori.ts
@@ -29,7 +29,7 @@ export const initIpsosMori = async (): Promise<void> => {
 	const isAU = locale === 'AU' && !!consentState.aus;
 	const isUK = locale === 'GB' && !!consentState.tcfv2;
 
-	if (!isAU && !isUK) {
+	if (!isAU && !isUK && !window.guardian.config.switches.ipsosMori) {
 		log('commercial', 'Skipping ipsos process outside GB or AU');
 	}
 
--- a/bundle/src/init/consented/third-party-tags.ts
+++ b/bundle/src/init/consented/third-party-tags.ts
@@ -101,7 +101,7 @@ const loadOther = (): Promise<void> => {
 			shouldRun: window.guardian.config.switches.permutive ?? false,
 		}),
 		ias({
-			shouldRun: window.guardian.config.switches.ias ?? false
+			shouldRun: window.guardian.config.switches.ias ?? false,
 		}),
 		inizio({ shouldRun: window.guardian.config.switches.inizio ?? false }),
 	].filter((_) => _.shouldRun);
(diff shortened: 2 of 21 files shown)


bundle/src/init/consented/third-party-tags.ts
```diff
@@ -100,7 +100,9 @@ const loadOther = (): Promise<void> => {
 		permutive({
 			shouldRun: window.guardian.config.switches.permutive ?? false,
 		}),
-		ias,
+		ias({
+			shouldRun: window.guardian.config.switches.ias ?? false,
```
DanielCliftonGuardian, line 104:
```suggestion
			shouldRun: window.guardian.config.switches.commercialIas ?? false,
```

Should this now be `commercialIas` after the frontend update. Matching others like `ipsosMori` etc

cemms1: GOOD spot!

cemms1: I'll actually swap the name of this in the frontend PR to simply `ias` for more consistent naming actually 👍

bundle/src/init/consented/ipsos-mori.ts
```diff
@@ -29,7 +29,7 @@ export const initIpsosMori = async (): Promise<void> => {
 	const isAU = locale === 'AU' && !!consentState.aus;
 	const isUK = locale === 'GB' && !!consentState.tcfv2;
 
-	if (!isAU && !isUK) {
+	if (!isAU && !isUK && !window.guardian.config.switches.ipsosMori) {
```
domlander: If the switch is off, shouldn't we be returning early? I would suggest using a different log message for when the switch is off.

domlander:
If a feature is behind a feature switch, I quite like the pattern of the first thing we do in the function to be checking the switch and then logging and returning early if the switch is off. I notice we don't currently do this very often.

bundle/src/lib/header-bidding/prebid/id-handlers/liveramp.ts
```diff
@@ -100,9 +99,7 @@ const getLiveRampParams = async (
 export const getUserIdForLiveRamp = async (
 	email: string | null,
 ): Promise<LiverampUserIdConfig | undefined> => {
-	const isLiverampEnabled = isSwitchedOn('prebidLiveramp');
```
cemms1, line 103: switch status checked in ./index.ts along with other ID providers

bundle/src/lib/header-bidding/prebid/id-handlers/index.spec.ts
```diff
@@ -1,39 +1,34 @@
...
->;
-const mockGetUserIdForTradeDesk = getUserIdForTradeDesk as jest.MockedFunction<
-	typeof getUserIdForTradeDesk
->;
+const mockGetEmail = jest.mocked(getEmail);
+const mockGetUserIdForId5 = jest.mocked(getUserIdForId5);
+const mockGetUserIdForLiveRamp = jest.mocked(getUserIdForLiveRamp);
+const mockGetUserIdForTradeDesk = jest.mocked(getUserIdForTradeDesk);
+const mockGetUserIdForIntentIq = jest.mocked(getUserIdForIntentIQ);
+const mockGetUserIdForOzone = jest.mocked(getUserIdForOzone);
```
cemms1, line 24: IntentIQ and Ozone were not previously covered by this test so test cases have been added here now to ensure the feature switch logic is tested alongside other logic

bundle/src/lib/header-bidding/prebid/id-handlers/index.spec.ts
```diff
@@ -191,17 +274,89 @@ describe('getUserSyncSettings', () => {
...
+				expect(mockGetUserIdForTradeDesk).toHaveBeenCalledWith(
+					'test@example.com',
+					consentStateWithTradeDesk,
+				);
+				expect(result.userIds).toContain(mockTradeDeskUserId);
+				expect(result.userIds).toContain(sharedId);
+			});
 
-			const result = await getUserSyncSettings(consentStateWithTradeDesk);
+			it('should include theTradeDesk userId when consent is given for theTradeDesk but the switch is off', async () => {
```
domlander: Is this test name correct? It looks like the trade desk user id should not be included?

domlander (approved):
Looks great, thanks for updating all those tests.

Remember to turn on the new switches "to cover integrations not currently protected by feature switches" before merging this PR.

bundle/src/lib/header-bidding/prebid/id-handlers/index.spec.ts
cemms1: d'oh! thank you for spotting!

PR #2703, opened 2026-08-28
Update changesets release action

cemms1:
## What does this change?

- Adds `continue-on-error` for changesets release action
  - This now fails when there is nothing to release (see https://github.com/changesets/changesets/commit/92b1c1b6b1733b1f1799a73828d378646b068798) 
  - We don't have any dependent steps after this so we don't need to rely on a fact that a new version was published or not
 

Additionally:
- removes GITHUB_TOKEN env variable from the other changesets/action usage, preferring to provide this via the `github-token` input instead (see https://github.com/changesets/action/commit/cb3f0110d7423cd340b1c5d63584c0ea6ee63959)
- Renames `app-id` to `client-id` in input for `actions/create-github-app-token` after warning in console about input deprecation
- Adds comments to explain why the two Github tokens are different for the beta changesets release and the full changesets release

## Why?

Fixing the NPM release workflow after upgrading to `changesets/action@v2` and `@changesets/cli@v3`

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -65,6 +65,9 @@ jobs:
           commit-message: Bump package versions
           # Changesets app token
           github-token: ${{ steps.app-token.outputs.token }}
+        # If there are no unreleased changes, this action fails
+        # We have no dependent steps so don't need to recognise this as a failure
+        continue-on-error: true
 
   publish-beta:
     name: '@guardian/commercial-core'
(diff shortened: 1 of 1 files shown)


DanielCliftonGuardian (approved):

cemms1:
We're green again 🎉 https://github.com/guardian/commercial/actions/runs/33156900003/job/98801730767

PR #2705, opened 2026-08-28
Ozone outstream tweaks

domlander:
## What does this change?

Limits the Prebid `video` media type to the Ozone Outstream size only. The `playerSize` value should of type: Array[Integer,Integer][^1].

Adds `mimes: ['video/mp4']` to the video media type. Some bidders explicitly look for this.

Updates the price floors to explicitly set the floor to 10c for both banner and video. When testing locally, the wildcard `'*'` was only being used for banner. I'm unsure to why this is, but being explicit fixes the issue.

Adds the size `outstreamOzone` to the list of proxy ad sizes. The height 360px is only fulfilled if there is room for it on the page, otherwise it will shrink down to maintain aspect ratio. We don't want to set a minimum height by default for this ad size.

## Why?

We are not seeing meaningful spending with the new Ozone outstream ad size and we hope these changes might fix that. 

## Screenshots

image

[^1]: https://docs.prebid.org/dev-docs/adunit-reference.html

### DIFF
--- a/bundle/src/lib/header-bidding/prebid/prebid-ad-unit.ts
+++ b/bundle/src/lib/header-bidding/prebid/prebid-ad-unit.ts
@@ -1,3 +1,4 @@
+import { outstreamSizes } from '@guardian/commercial-core/ad-sizes';
 import type { PageTargeting } from '@guardian/commercial-core/targeting/build-page-targeting';
 import type { ConsentState } from '@guardian/consent-manager';
 import { log } from '@guardian/libs';
@@ -9,7 +10,7 @@ import type { MediaTypes } from 'prebid.js/dist/src/mediaTypes';
 import type { VideoMediaType } from 'prebid.js/dist/src/video';
 import type { Advert } from '../../../define/Advert';
 import type { HeaderBiddingSlot } from '../prebid-types';
-import { isOutstream } from '../utils';
+import { isOutstreamOzone } from '../utils';
 import { bids } from './bidders/config';
 
 export class PrebidAdUnit implements AdUnitDefinition {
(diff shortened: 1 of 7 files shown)


bundle/src/lib/header-bidding/prebid/prebid-ad-unit.ts
```diff
@@ -33,10 +34,12 @@ export class PrebidAdUnit implements AdUnitDefinition {
...
 	) {
 		/**
-		 * Outstream ad sizes are only compatible with the mediaTypes.video property of PrebidAdUnit
+		 * Only the Outstream Ozone ad size is compatible with the mediaTypes.video property of PrebidAdUnit
 		 */
-		const bannerSizes = slot.sizes.filter((size) => !isOutstream(size));
-		const videoSizes = slot.sizes.filter((size) => isOutstream(size));
+		const bannerSizes = slot.sizes.filter(
+			(size) => !isOutstreamOzone(size),
+		);
```
cemms1, line 38: Do we want to keep the previous outstream definition for non video ads?

domlander: This is confusing at the moment. I'm confirming with Jasper that we do actually want to limit the video media type to only the Ozone outstream ad size, then I'll tidy the naming up.

domlander: Jasper has confirmed that we only want the Ozone outstream size to use the video media type, so I'll update the naming

domlander: I've had another pass. I've removed the comment as it's potentially misleading.

cemms1: It looks like we previously didn't allow any outstream size to be classed as banner but now we allow outstreams that aren't specifically Ozone ones. Is that what we want?

cemms1 (approved):
👍

bundle/src/lib/header-bidding/prebid/prebid-ad-unit.ts
domlander: Yes this is now what we want; sorry I haven't explained this at all. In summary:

- Initially, we had all outstream ad sizes using the `banner` media type.
- Then we introduced the `video` media type for prebid for the new outstream size: `outstreamOzone`. Now that we have the `video` media type, I thought it would be a good idea to move the old outstream sizes to the video type as well.
- The problem: when there are multiple sizes, `playerSize` is not populated in the network request (screenshot in description). This is not mandatory, but according to Jasper, creative matching is difficult for some video DSPs when this is omitted.
- The solution: since the old outstream sizes used to work fine using the `banner` media type, I've reverted to this behaviour (even though semantically it doesn't make sense). This means the only the Ozone outstream size will use the `video` media type so that `playerSize` is populated in the network request.

PR #2709, opened 2026-09-07
mobile above nav (Test for crosswords)

oliverabrahams:
## What does this change?

Adds a new mobile above nav slot with a new size `mobileLeaderboardXl: 100px 320px` I have kept the naming consistent with mobile sticky for now even though it should be `mobileLeaderboardXl`.

## Why?

We are experimenting with a above nav advert on the crossword pages and the crossword front

### DIFF
--- a/.changeset/sharp-lands-pull.md
+++ b/.changeset/sharp-lands-pull.md
@@ -2,5 +2,5 @@
 "@guardian/commercial-core": minor
 ---
 
-Adding a new slot mobile-above-nav and a new size mobilestickyXl 100x320
+Adding a new slot mobile-above-nav and a new size mobilestickyXl 320x100
 
--- a/.changeset/sharp-lands-pull.md
+++ b/.changeset/sharp-lands-pull.md
@@ -2,5 +2,5 @@
 "@guardian/commercial-core": minor
 ---
 
-Adding a new slot mobile-above-nav and a new size mobilestickyXl 320x100
+Adding a new slot mobile-above-nav and a new size mobileLeaderboardXl 320x100
 
--- a/.changeset/sharp-lands-pull.md
+++ b/.changeset/sharp-lands-pull.md
@@ -1,0 +1,6 @@
+---
+"@guardian/commercial-core": minor
+---
+
+Adding a new slot mobile-above-nav and a new size mobilestickyXl 100x320
+
--- a/bundle/src/events/render-advert-label.ts
+++ b/bundle/src/events/render-advert-label.ts
@@ -133,6 +133,17 @@ const renderAdvertLabel = (
 						'true',
 					);
 				}
+				if (
+					adSlotNode.parentElement?.classList.contains(
+						'ad-slot-container',
+					) &&
+					adSlotNode.id === 'dfp-ad--mobile-above-nav'
+				) {
+					adSlotNode.parentElement.setAttribute(
+						'mobile-above-nav-ad-rendered',
+						'true',
+					);
+				}
 				// \Remove this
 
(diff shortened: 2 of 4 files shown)


bundle/src/events/render-advert-label.ts
```diff
@@ -121,17 +126,15 @@ const renderAdvertLabel = (
...
 						'ad-slot-container',
 					) &&
-					adSlotNode.id === 'dfp-ad--top-above-nav'
+					renderedAttr
 				) {
-					adSlotNode.parentElement.setAttribute(
-						'top-above-nav-ad-rendered',
-						'true',
-					);
+					adSlotNode.parentElement.setAttribute(renderedAttr, 'true');
```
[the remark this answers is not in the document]
oliverabrahams: There isn't actually a test for this on top above nav at the moment.

core/src/ad-sizes.ts
```diff
@@ -136,6 +138,7 @@ const namedStandardAdSizes = {
 	halfPage: createAdSize(300, 600),
 	leaderboard: createAdSize(728, 90),
 	mobilesticky: createAdSize(320, 50),
+	mobilestickyXl: createAdSize(320, 100),
 	mpu: createAdSize(300, 250),
```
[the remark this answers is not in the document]
oliverabrahams: Done

bundle/src/events/render-advert-label.ts
```diff
@@ -80,6 +80,11 @@ const createAdTestLabel = (
 	return adTestLabel;
 };
 
+const renderedAttrSlots: Record<string, string> = {
+	'dfp-ad--top-above-nav': 'top-above-nav-ad-rendered',
+	'dfp-ad--mobile-above-nav': 'mobile-above-nav-ad-rendered',
+};
```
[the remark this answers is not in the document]
oliverabrahams: Disagree

core/src/ad-sizes.ts
```diff
@@ -90,6 +90,7 @@ type SizeKeys =
 	| 'merchandisingHigh'
 	| 'merchandisingHighAdFeature'
 	| 'mobilesticky'
+	| 'mobilestickyXl'
```
cemms1: Please can we not call it "mobile sticky" since we aren't intending to use it here and we shouldn't have really called it this in the first place - "mobile sticky" describes a slot rather than an advert size.
It would be nice to do a clean up exercise to rename this ad size across the repo. I don't believe we use this name anywhere else as I think we would usually use the dimensions when setting in the page and slot targeting

[According to Google](https://support.google.com/admanager/answer/1100453?hl=en), the size we commonly call "mobile sticky" is "mobile leaderboard" so we could call the new size "mobile leaderboard xl"?

Image

cemms1: I did it here actually as it was bothering me and hadn't noticed until you raised this PR! https://github.com/guardian/commercial/pull/2710

oliverabrahams: Amazing, done and also approved urs.

cemms1: Merged! Will resolve this thread

.changeset/sharp-lands-pull.md
```diff
@@ -0,0 +1,6 @@
+---
+"@guardian/commercial-core": minor
+---
+
+Adding a new slot mobile-above-nav and a new size mobilestickyXl 320x100
```
cemms1: Please can we update this description + the PR description with the new name?

oliverabrahams: Done

cemms1 (approved):

PR #2710, opened 2026-09-07
Rename `mobile-sticky` ad size to `mobileLeaderboard`

cemms1:
## What does this change?

Renames "mobile-sticky" ad size to "mobileLeaderboard"

This is a refactoring change only and should not affect slot targeting sent to GAM or elsewhere

## Why?

The term "mobile sticky" refers to the slot not the ad size. We aleady use the "mobile sticky" ad size in a slot that is not "mobile sticky": in the mobile crosswords banner slot

The fact that it is called "mobile sticky" is confusing as this describes the behaviour and not the slot size.
[According to Google](https://support.google.com/admanager/answer/1100453?hl=en), the name commonly used for the size 320x50 is "mobile leaderboard"

Image

### DIFF
--- a/.changeset/orange-hotels-retire.md
+++ b/.changeset/orange-hotels-retire.md
@@ -1,0 +1,6 @@
+---
+"@guardian/commercial-core": patch
+---
+
+Update name of mobile-sticky ad size to mobileLeaderboard
+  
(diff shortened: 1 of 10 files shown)


oliverabrahams (approved):
Looks good to me! will rename mine

PR #2713, opened 2026-09-08
Update Spacefinder rich links rules

domlander:
## What does this change?

When in the variant of the `commercial-rich-links` AB test:
- If the rich link component has not hydrated and the rich link image has not loaded by the time Spacefinder runs, calculate the eventual height of the image, combining this with the height of the rest of the rich link element to get its eventual height once it loads.
- From the `leftCol` breakpoint, when the left column content is above the potential advert placement, the top of the advert cannot be within 100px of the top of the content. Adverts can also be placed closer when the content is below the advert placement.
- At mobile/tablet breakpoints, reduces the amount of space required between the bottom of an advert and the top of a rich link from 200px to 100px. Currently, a paragraph needs 200px of space between its top and the top of a rich link, and 35px between its top and the bottom of an advert.

## Why?

- We do not wait for the rich link image to load before inserting inline ads. There is a race condition between the rich link image loading and Spacefinder running. Sometimes the rich link image loads before Spacefinder runs and sometimes after. The ad can appear adjacent to the rich link if the ad slots are inserted before the rich link image loads. At some breakpoints, this can create whitespace between a rich link, a paragraph and an ad.
- There is one caveat that sometimes rich links do not have an image and we will be assuming that it does. Given the images without images are seldom used, hopefully this does not have a large effect.

Given that the first change, despite being a fix, will have a negative impact on ad-ratio, we also tweak the Spacefinder rules at the leftCol breakpoint and mobile/tablet:

- Before this change, from the `leftCol` breakpoint when the left column content is above the potential advert placement, the top of the advert could not be within 50px from the _bottom_ of the content. From the left-col breakpoint, rich links appear entirely in the left-column. Content in the left column has less visual weight than inline content, so we don't need to avoid these elements as much when inserting adverts.
- At mobile/tablet breakpoints, rich links are only 140px wide so the visual weight of ads and rich links is not too strong. This change should have a milder effect on ad-ratio.

## Screenshots

When Spacefinder wins the race against the rich link image, we get the image on the right: Spacefinder does not take into account the height of the image and so thinks it can place an ad higher up than it should. The intention was to leave at least 50px between the bottom of the rich link and the top of the content, which it has clearly failed. When the rich link wins the race, we get the image on the left. This is the intended functionality, but does leave the advert quite far down the page.

image

Restrictions have been loosened on adverts appearing close to left-column content from the leftCol breakpoint.

Here is an example where the advert appears below the content.

ad-below

And one where the advert appears above the content. You can see the reason why the winner on the left was not a winner on the right in the Spacefinder Debug Mode message.

ad-above

An example of the change at mobile/tablet breakpoints. Look at the middle page and the second winning paragraph on this page. The one that begins: "My parents said they had tried...". Adverts are now able to be placed closer above rich links and thumbnails. Before this change, the bottom of the advert had to be at least 200px from the top of the content, as shown by Spacefinder Debug mode in the screenshot. This seems too high and has been reduced to 100px.

Screenshot 2026-09-07 at 18 38 38

### DIFF
--- a/bundle/playwright/tests/rich-links.spec.ts
+++ b/bundle/playwright/tests/rich-links.spec.ts
@@ -1,0 +1,55 @@
+import { expect, test } from '@playwright/test';
+import { breakpoints } from '../lib/breakpoints';
+import { cmpAcceptAll } from '../lib/cmp';
+import { loadPage } from '../lib/load-page';
+import { waitForIsland } from '../lib/util';
+
+const url =
+	'/Article/https://www.theguardian.com/football/2026/sep/06/exceptional-arteta-arsenal-fightback-win-chelsea';
+
+/**
+ * Within Spacefinder, we make an assumption on the height of the rich link images in
+ * order to place adverts appropriately. If the display of rich link images changes,
+ * we may need to revisit this assumption.
+ */
+test.describe('Rich links', () => {
+	const expectedImageHeightsByBreakpoint = {
+		mobile: 0,
+		tablet: 112,
+		desktop: 112,
+		wide: 176,
+	};
+
+	breakpoints.forEach(({ breakpoint, width, height }) => {
+		test(`Rich link has expected dimensions ${breakpoint}`, async ({
+			page,
+		}) => {
+			await page.setViewportSize({
+				width,
+				height,
+			});
+
+			await loadPage({ page, path: url });
+			await cmpAcceptAll(page);
+
+			await waitForIsland(page, 'RichLinkComponent');
+
+			// const richLinkImage = page.getByTestId('rich-link-image');
+			const richLinkImage = page.getByAltText(
+				'Martin Ødegaard (right) with Declan Rice',
+			);
+
+			const expectedImageHeight =
+				expectedImageHeightsByBreakpoint[breakpoint];
+
+			if (expectedImageHeight === 0) {
+				await expect(richLinkImage).not.toBeVisible();
+				return;
+			}
+
+			const boundingBox = await richLinkImage.boundingBox();
+
+			expect(boundingBox?.height).toBe(expectedImageHeight);
+		});
+	});
+});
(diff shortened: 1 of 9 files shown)


bundle/playwright/tests/rich-links.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+				width,
+				height,
+			});
+
+			await loadPage({ page, path: url });
+			await cmpAcceptAll(page);
+
+			await waitForIsland(page, 'RichLinkComponent');
+
+			// const richLinkImage = page.getByTestId('rich-link-image');
```
DanielCliftonGuardian:
```suggestion
```

core/src/geo/get-locale.ts
```diff
@@ -26,7 +26,7 @@ export const __resetCachedValue = (): void => (locale = undefined);
 
 /**
  * Fetches the user's current location as an ISO 3166-1 alpha-2 string e.g. 'GB', 'AU' etc
- * Note: This has been copied from guardian-libs and made syncronous by ommiting the call to
+ * Note: This has been copied from guardian-libs and made syncronous by omitting the call to
```
DanielCliftonGuardian:
```suggestion
 * Note: This has been copied from guardian-libs and made synchronous by omitting the call to
```

bundle/src/insert/spacefinder/spacefinder.ts
```diff
@@ -335,9 +371,11 @@ const testCandidate = (
...
 		} else {
 			// if the test fails, add debug information to the candidate metadata
-			const required = isOpponentBelow
-				? rule.marginTop
-				: rule.marginBottom;
+			const required = rule.isLeftColumnOpponent
+				? 0
+				: isOpponentBelow
+					? rule.marginTop
+					: rule.marginBottom;
```
DanielCliftonGuardian:
```suggestion
			const required = rule.isLeftColumnOpponent
				? rule.distanceBetweenTops;
				: isOpponentBelow
					? rule.marginTop
					: rule.marginBottom;
```

Should this be updated with the new logic ?

domlander: Ah yes, good spot! I think "actual" needs to be updated too.

DanielCliftonGuardian (approved):
Nice work. Looks good

PR #2716, opened 2026-09-08
update bundle config as defaults were overwriting browserslist

oliverabrahams:
## What does this change?

Attempt to fix issue with bundle in crosswords


bundle/babel.config.json
```diff
@@ -1,7 +1,6 @@
 {
 	"presets": ["@babel/preset-env", "@babel/preset-react"],
 	"plugins": ["@babel/plugin-transform-runtime"],
-	"targets": "defaults",
```
cemms1, line 4: We need this field in Babel v8. See PR description #2671 

I think we would need to update this to our browsers list?

oliverabrahams: I think that because we have the browserslist set in the package.json, Babel will pick it up from there. this overwrites the config in package json as far as i can tell

oliverabrahams: I am just testing some options to see if i can find out what it is. sorry this should have been a draft PR

PR #2719, opened 2026-09-09
Use port 3032 for DCR in playwright

cemms1:
## What does this change?

When running DCR in a docker container for use in Playwright, this now uses port 3032 in the dev container and forwards this port outside of the dev container too

## Why?

This ensures we avoid clashing with local running DCR on port 3030 when running DCR + commercial locally

### DIFF
--- a/scripts/start_dcr_container.sh
+++ b/scripts/start_dcr_container.sh
@@ -16,7 +16,7 @@ echo "Pulling image $IMAGE"
 # The container images are created via this workflow: https://github.com/guardian/dotcom-rendering/blob/18f4f9f6b861b04e2ad8deed82f442938e9f9197/.github/workflows/container.yml
 /usr/bin/docker run -d \
     --network host \
-    -p 3030:3030 \
+    -p 3030:3032 \
     -e "PORT=3030" \
     -e "COMMERCIAL_BUNDLE_URL=http://localhost:3031/graun.standalone.commercial.js" \
     $IMAGE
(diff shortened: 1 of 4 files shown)


scripts/start_dcr_container.sh
```diff
@@ -16,7 +16,7 @@ echo "Pulling image $IMAGE"
 # The container images are created via this workflow: https://github.com/guardian/dotcom-rendering/blob/18f4f9f6b861b04e2ad8deed82f442938e9f9197/.github/workflows/container.yml
 /usr/bin/docker run -d \
     --network host \
-    -p 3030:3030 \
+    -p 3030:3032 \
```
DanielCliftonGuardian, line 19: What is the first port doing here ? Should this be 3032 also ?
```suggestion
    -p 3032:3032 \
```

cemms1: We do actually want to keep the first one as 3030 as this is the port that DCR uses to run. What we're doing here is linking port 3030 running _inside_ the docker container to 3032 when accessing outside of the container (but within the dev container).

What we have is something like the following:

```
.______________________.
| dev container        |
| access DCR at 3032   |
|  ._______________.   |
|  | DCR in docker |   |
|  | port 3030     |   |
|  ._______________.   |
|                      |
.______________________.

```

We expose port 3032 from the docker container running _inside_ the dev container (also a docker container) and link this to port 3030 used inside the container so we can access the app running on port 3030 inside the container on port 3032 outside of the container.

Docs on this are here:
- https://docs.docker.com/reference/cli/docker/container/run/#publish
- https://docs.docker.com/engine/network/port-publishing/

DanielCliftonGuardian (approved):

scripts/start_dcr_container.sh
DanielCliftonGuardian: Thanks 👍

PR #2721, opened 2026-09-09
Disable spacefinder when data attribute detected in article body

cemms1:
## What does this change?

Prevents Spacefinder runnning if the `data-spacefinder-disabled` attribute is found on the tag with the class `article-body-commercial-selector`

## Why?

We need to be able to turn off Spacefinder for select articles ahead of them being published.
This is a proposed way to do this, as the interactive code can manipulate the DOM to add this data attribute


cemms1:
Closing in favour of #2722

PR #2722, opened 2026-09-10
Disable spacefinder for the US elections page

cemms1:
## What does this change?

Disables Spacefinder on the US midterm elections page when detecting presence of the interactive atom used to display the results

## Why?

We want to manually insert ads on these pages to have a more predictable outcome instead of allowing Spacefinder to run

We can't use a data attribute like `data-spacefinder-disabled` because we would need the interactive code to add this to the DOM after it loads which would be too late for Spacefinder to pick it up

If we decided to use the commercial queue via the `window.guardian` object, we also can't guarantee this information will be available by the time Spacefinder runs if it's being added by the interactive code. We would need to emit an Event (similar to in mobile-sticky handling) which would signal to stop running Spacefinder. This would likely happen after Spacefinder has already initialised which would result in removal or collapsing of slots which would result in non ideal layout shift.

The safest way of doing this to avoid race conditions is to ensure the information is available on the page itself and not relying on interactive code to send signals to the commercial bundle code to pick up

### DIFF
--- a/bundle/src/lib/article-body-adverts.ts
+++ b/bundle/src/lib/article-body-adverts.ts
@@ -36,11 +36,13 @@ const allowArticleBodyAdverts = (): boolean => {
 	const isHosted = window.guardian.config.page.isHosted;
 	const newRecipeDesign =
 		window.guardian.config.page.showNewRecipeDesign ?? false;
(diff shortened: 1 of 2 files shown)


oliverabrahams (approved):
LGTM

PR #2723, opened 2026-09-10
Refactor header bidding initialisation checks. Add APS constant

domlander:
## What does this change?

Creates a constant for the APS header bidding auction timeout. This is 1500ms, the same as Prebid.

When `requestBids` is called for Prebid/APS, run the same synchronous checks we do when we initialise Prebid and move the helper functions into the header bidding utils file.

## Why?

There's an upcoming AB test which will change both the Prebid and APS auction timeout values. This will be easier to achieve if there are constants for both. Also, a constant for APS will help to ensure APS is not forgotten when updating header bidding code. We sometimes overlook APS as we rarely make changes in this area.

When we initialise Prebid/APS, we run some checks to ensure that we are allowed to run it, including whether the feature switch is ON. Later, when we request bids, we check that Prebid/APS is initialised and we also check the feature switch again. With this change, instead of only checking the feature switch, we include the other synchronous checks too. This is not necessary, but is defensive and I think easier to follow. If we were to introduce a Prebid/APS check in the future that may change in the page lifecycle, then this change will protect against us running Prebid/APS when we shouldn't.

### DIFF
--- a/bundle/src/lib/header-bidding/utils.ts
+++ b/bundle/src/lib/header-bidding/utils.ts
@@ -11,6 +11,10 @@ import { type ConsentState, getConsentFor } from '@guardian/consent-manager';
 import { isString } from '@guardian/libs';
 import { once } from 'lodash-es';
 import type { Size } from 'prebid.js/dist/src/types/common';
+import { isAdFree } from '../../lib/ad-free';
+import { isGoogleProxy } from '../../lib/detect/detect-google-proxy';
+import { isSecureContactPage } from '../../lib/is-secure-contact';
+import { shouldLoadAds } from '../../lib/should-load-ads';
 import {
 	getCurrentTweakpoint,
 	matchesBreakpoints,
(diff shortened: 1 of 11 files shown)


bundle/src/lib/header-bidding/utils.ts
```diff
@@ -50,6 +54,13 @@ const isValidPageForMobileSticky = (): boolean => {
 	);
 };
 
+const canLoadHeaderBidders = (): boolean =>
+	!isGoogleProxy() &&
+	shouldLoadAds() &&
+	!isAdFree() &&
```
cemms1, line 60: Realise this was a lift and shift but in the nature of improving things, we already check for ad free in the `shouldLoadAds` function so don't strictly need to do it again here

cemms1 (approved):
🧹

bundle/src/lib/header-bidding/utils.ts
domlander: Nice spot, thank you

bundle/src/init/consented/prepare-a9.spec.ts
```diff
@@ -144,17 +137,6 @@ describe('prepareA9', () => {
 			a9HeaderBidding: true,
 		};
 		jest.mocked(shouldLoadAds).mockReturnValue(false);
-		jest.mocked(isAdFree).mockReturnValue(false);
-		await setupA9();
-		expect(a9.initialise).not.toHaveBeenCalled();
-	});
-
-	it('should not initialise a9 when ad-free is on', async () => {
```
domlander, line 152: Ad-free is now covered by the above test

